PR #2, opened 2025-02-13
[QHC-873] Implement Skeleton for Circuit and QiboBackend

fedonman:


src/qilisdk/digital/gates.py
```diff
@@ -0,0 +1,435 @@
...
+    The associated matrix is:
+        [[exp(-i*theta/2),              0],
+         [0,               exp(i*theta/2)]]
+    """
+
+    _NAME: ClassVar[str] = "RZ"
+    _NQUBITS: ClassVar[int] = 1
+    _PARAMETER_NAMES: ClassVar[list[str]] = ["theta"]
+
+    def __init__(self, qubit: int, *, theta: float) -> None:
```
GuillermoAbadLopez, line 417: @fedonman I think I would change here `theta` by `phi`, since that's the standard used in quantum physics, and what `qiskit` and `pennylane` use. 

`Qibo` uses `theta`, so I guess that is the origin of the name here, since `app_sdk` copied it from `qibo`, and I guess you from there, no? And actually, I came checking here because I just changed it in `app_sdk` in my PR and knew you were copying it here, so came to warn bout it 👌

src/qilisdk/digital/circuit.py
```diff
@@ -0,0 +1,103 @@
...
+    def __init__(self, nqubits: int) -> None:
+        """
+        Initialize a Circuit instance with a specified number of qubits.
+
+        Args:
+            nqubits (int): The number of qubits in the circuit.
+        """
+        self._nqubits: int = nqubits
+        self._gates: list[Gate] = []
+        self._init_state: np.ndarray = np.zeros(nqubits)
```
amirazzam2000, line 30: we should have a quantum state class and this should be an instance of that class for consistency.

src/qilisdk/digital/digital_result.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @property
+    def samples(self) -> np.ndarray:
+        return self._samples
+
+
+class SimulationDigitalResults(DigitalResults):
+    def __init__(
+        self,
+        state: np.ndarray,
```
amirazzam2000, line 53: we should discuss if this should also be added to the digital results if you can perform state tomography with the hardware.

src/qilisdk/digital/gates.py
```diff
@@ -0,0 +1,583 @@
...
+        Set the parameters for the gate using a dictionary mapping names to values.
+
+        Args:
+            parameters (dict[str, float]): A dictionary where keys are parameter names and values are the new parameter values.
+
+        Raises:
+            GateHasNoParameter: If any provided parameter name is not valid for this gate.
+        """
+        if any(name not in self._PARAMETER_NAMES for name in parameters):
+            raise GateHasNoParameter
```
amirazzam2000: maybe the exception name could be changed to "GateParameterNameInvalid" so that it is a bit more understandable. Because this reads as the gate has no parameters.

src/qilisdk/digital/gates.py
```diff
@@ -0,0 +1,583 @@
...
+    def parameters(self) -> dict[str, float]:
+        """
+        Retrieve a mapping of parameter names to their corresponding values.
+
+        Returns:
+            dict[str, float]: A dictionary mapping each parameter name to its numeric value.
+        """
+        return dict(zip(self._PARAMETER_NAMES, self._parameter_values))
+
+    def set_parameters(self, parameters: dict[str, float]) -> None:
```
amirazzam2000, line 140: do you think we should name this method `set_parameter_by_name` or `set_parameter_by_key` and make the list method the default with the name `set_parameters` to stay consistent with the other libraries?

src/qilisdk/digital/gates.py
```diff
@@ -0,0 +1,585 @@
...
+
+
+class M(Gate):
+    """
+    Measurement operation on a qubit. The measurement is performed in the computational basis and the
+    result is stored in a classical bit with the same label as the measured qubit.
+    """
+
+    _NAME: ClassVar[str] = "M"
+    _NQUBITS: ClassVar[int] = -1
```
amirazzam2000, line 180: if you plan to change this in future maybe it's worth adding a TODO tag?

src/qilisdk/digital/gates.py
```diff
@@ -0,0 +1,585 @@
...
+        """
+        super().__init__()
+        self._target_qubits = (qubit,)
+        self._parameter_values = [theta]
+        self._matrix = np.array([[1, 0], [0, np.exp(1j * theta)]], dtype=complex)
+
+
+class U2(Gate):
+    """
+    Represents the U2 gate defined by the angles `phi` and `lambda`.
```
amirazzam2000, line 463: should we be calling it lambda in the description if we follow the naming convention of lam in the code? it might be confusing for the user I think.

src/qilisdk/extras/qibo_backend.py
```diff
@@ -11,8 +11,138 @@
...
+        return SimulationDigitalResults(
+            state=qibo_results.state(),
+            probabilities=qibo_results.probabilities(),
+            samples=qibo_results.samples(),
+            frequencies=qibo_results.frequencies(),
+            nshots=nshots,
+        )
+
+    @staticmethod
+    def to_qibo(circuit: Circuit) -> QiboCircuit:
```
amirazzam2000, line 59: I think maybe we should do this similar to what Guille did in the AppSDK.

src/qilisdk/digital/circuit.py
fedonman: Yes, we can change it when the class is ready.

src/qilisdk/digital/digital_result.py
fedonman: Does the hardware send back the `state` though or do you retrieve it from postprocessing?

src/qilisdk/extras/qibo_backend.py
fedonman: This is similar to Guille's with added strong typing. What exactly would you change?

src/qilisdk/digital/digital_result.py
amirazzam2000: I mean there is a routine that we would have to perform, but maybe this could be a pre-programmed Qprogram and then qililab does all the processing and sends the recreated state?

src/qilisdk/digital/gates.py
fedonman: Every library uses its own convention. E.g. qiskit uses `assign_parameters` with a mapping as this one. Also `set_parameter_by_name` or `set_parameter_by_key` are both too long with no reason. I would leave the naming as it is.

src/qilisdk/digital/gates.py
fedonman: True. I will change it.

src/qilisdk/digital/gates.py
fedonman: Yes, I'll add a TODO. Maybe we don't need this property at all, we can get the nqubits by len(qubits).

src/qilisdk/digital/gates.py
fedonman: @amirazzam2000 What is your input on this? I have no opinion pro or against, except the fact that `theta` is used in all rotation gates.

src/qilisdk/digital/gates.py
fedonman: Yes I'll change it.

src/qilisdk/digital/digital_result.py
fedonman: This is definitely not implemented in qililab now.

src/qilisdk/digital/gates.py
GuillermoAbadLopez: Normally `theta`, is for when you rotate "vertically" (polar rotation) (`Rx` and `Ry`), and `phi` in the azimuthal rotations (`Rz`). Another option is to use "angle" or "rads", so you are not using any convention, but just stating it is the angle, like they do in cirq 🤔 

Well lets talk it in the next meeting tomorrow 👍

src/qilisdk/extras/qibo_backend.py
GuillermoAbadLopez: I like the strong typing definetly, but I get Ameer worry, this is much longuer, its like going back to the old way we had before [my recent PR](https://github.com/qilimanjaro-tech/applications-sdk/pull/82), but instead of defining new classes as `Gates`, they are just `static_methods`... 🤔 

Which I don;t totally dislike, the strong tiping definitely compensates, but...

GuillermoAbadLopez: Actually, on second thought It seems maybe we could do strong typing in the new shorted way, using some clever `TypeVar` (with the `_X_gate_map` `Gate` mapping dictionary), in the `X_gate_mapper` method 🤔

 Yeah, I think we can do it much shorter and still keep the strong typing. I can think about it on friday :)

GuillermoAbadLopez: But if we cannot, at least in this implementation, it's clear we don't need this full dict, no?:
```python
to_qibo_converters: dict[type[Gate], Callable[[Any], QiboGates.Gate]] = {
            X: QiboBackend._to_qibo_X,
            Y: QiboBackend._to_qibo_Y,
            Z: QiboBackend._to_qibo_Z,
            H: QiboBackend._to_qibo_H,
            S: QiboBackend._to_qibo_S,
            T: QiboBackend._to_qibo_T,
            RX: QiboBackend._to_qibo_RX,
            RY: QiboBackend._to_qibo_RY,
            RZ: QiboBackend._to_qibo_RZ,
            U1: QiboBackend._to_qibo_U1,
            U2: QiboBackend._to_qibo_U2,
            U3: QiboBackend._to_qibo_U3,
            CNOT: QiboBackend._to_qibo_CNOT,
            CZ: QiboBackend._to_qibo_CZ,
            M: QiboBackend._to_qibo_M,
        }
```
We can just use instead a one-liner mapping, like:  `whatever: QiboBackend._to_qibo_<whatever>` with a `get_method/has_attribute(._to_qibo_<whatever>)` or something like that, no?

src/qilisdk/digital/gates.py
```diff
@@ -0,0 +1,585 @@
...
+    The associated matrix is:
+        1/sqrt(2)*[[exp(-i*(phi+lambda)/2), -exp(-i*(phi-lambda)/2)],
+                   [exp(i*(phi-lambda)/2),    exp(i*(phi+lambda)/2)]]
+    """
+
+    _NAME: ClassVar[str] = "U2"
+    _NQUBITS: ClassVar[int] = 1
+    _PARAMETER_NAMES: ClassVar[list[str]] = ["phi", "lam"]
+
+    def __init__(self, qubit: int, *, phi: float, lam: float) -> None:
```
GuillermoAbadLopez, line 474: Nice, I prefer much more `lam`, so we don't need to change it from `lamb` for `qiskit`, etc... sin they all used `lam` already 🔥

src/qilisdk/digital/gates.py
amirazzam2000: I would vote for keeping it theta for consistency propose within our library.

amirazzam2000 (approved):

PR #3, opened 2025-02-18
[QHC-876] Implement Hamiltonian

fedonman:
Introduces the `Hamiltonian` class as a central component for Pauli-based operator arithmetic, with a flyweight pattern for single-qubit operators. Internally stores terms as a dictionary mapping tuples of `PauliOperator` objects to complex coefficients.

**Key Features**

1. **Flyweight Pauli Operators**
   - `Z`, `X`, `Y`, `I` constructors backed by a cache to avoid repeated instantiation of identical operators.
   - Concrete classes `PauliZ`, `PauliX`, `PauliY`, and `PauliI` handle core matrix definitions and qubit association.

2. **Dictionary-Based Storage**
   - Each Hamiltonian term is keyed by a tuple of `PauliOperator` objects (e.g. `(Z(0), Y(1))`) associated with a complex coefficient.
   - Uses `defaultdict(complex)` for coefficient accumulation.

3. **Arithmetic Operations**
   - Addition, subtraction, multiplication, and division with other Hamiltonians, individual Pauli operators, or scalars (`int`, `float`, `complex`).
   - Merges like terms automatically (e.g., `(Z(0) + Z(0))` becomes `2*Z(0)`).
   - Distributes terms for Hamiltonian–Hamiltonian multiplication, applying a Pauli product table to compute phases and resulting operators on a per-qubit basis.

4. **Simplification**
   - Removes terms below a configurable threshold (`_EPS`).
   - Aggregates any single-qubit identity operators `I(q)` into `I(0)` for a canonical form.

5. **Iteration Protocol**
   - Exposes `__iter__` returning `(coefficient, list_of_operators)` pairs, facilitating term-by-term processing without a separate parsing function.

6. **Equality Check**
   - Implements `__eq__` to compare Hamiltonians and detect pure scalar cases (`I(0) * c`).
   - Recognizes an effectively “zero” Hamiltonian and treats it consistently.

7. **String Representation**
   - Renders numeric coefficients in a concise format (omitting unnecessary decimals).
   - Correctly handles purely real or imaginary parts, negative signs, and identity terms.
   - Arranges single identity terms `I(0)` to appear first.

8. **Parse Method**
   - Implements a `@classmethod` named `parse` that reconstructs a Hamiltonian from its string representation.
   - Allows round-trip conversions (str(ham) → parse(...) → ham) for saving, logging, or user input.

### Code Example

```python
from qilisdk.analog import Hamiltonian, X, Y, Z, I

# Create a Hamiltonian using Pauli Operators
H = (Z(0) + X(0)) * (Z(0) - X(0))

# Access each term
for coeff, ops in product:
    print("Coefficient:", coeff, "Operators:", ops)

# Check equality
assert H == -2j * Y(0), "Both expressions yield the same Hamiltonian."

print(H)  # -2j Y(0)

# Parse a string to Hamiltonian
parsed_hamiltonian = Hamiltonian.parse("-2j Y(0)")

# Check equality
assert H == parsed_hamiltonian, "Hamiltonians are equal"
```

This release provides a robust framework for Pauli-operator arithmetic, scalar integration, and canonical simplification, forming a foundation for higher-level quantum analog functionality.


src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+            - qubit (int): the qubit that the operator will act on.
+        """
+        super().__init__(qubit=qubit)
+
+
+class Y(PauliOperator):
+    """The Pauli Y operator"""
+
+    _NAME: ClassVar[str] = "Y"
+    _MATRIX: ClassVar[np.ndarray] = np.array([[0, 1j], [1j, 0]], dtype=complex)
```
amirazzam2000: sorry but I spotted this earlier this week, the matrix for the pauli Y is wrong. it should be [[0, -1j],[1j, 0]].
image

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+            int: the number of qubits
+        """
+        n_qubits = -1
+        for key in self._elements:
+            for qid, _ in key:
+                n_qubits = max(n_qubits, qid)
+
+        if n_qubits == -1:
+            raise InvalidHamiltonianOperation("Can't compute the number of qubits if the hamiltonian has no terms")
+        return n_qubits
```
amirazzam2000: I think this should have a +1 because the qubit ids start at 0 so if we have 5 qubits the IDs will be between 0 and 4. Therefore, this method will return 4 instead of 5.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+    def elements(self) -> dict[tuple[tuple[int, str], ...], complex]:
+        """Returns the dictionary of the elements
+
+        Returns:
+            dict: a dictionary of the hamiltonian elements and their coefficient
+        """
+        return self._elements
+
+    # TODO(fedonman, #3): Do we need this? Why is it a Generator?
+    def variables(self) -> Generator[PauliOperator, None, None]:
```
amirazzam2000: This is useful in cases you wanted a sequence of the operators in the Hamiltonian. we use it somewhere in appSDK I don't recall where exactly.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+        Yields:
+            PauliOperator: A pauli operator object.
+        """
+        for key in self.elements:
+            for qid, pauli in key:
+                yield Hamiltonian._PAULI_MAP[pauli](qid)
+
+    # TODO(fedonman, #3): Do we need this? Why is it a Generator?
+    # We should rename. Method named `parse` is used for parsing string inputs.
+    def parse(self) -> Generator[tuple[complex, list[PauliOperator]], None, None]:
```
amirazzam2000: This method is quite useful to parse the terms of the Hamiltonian. It's a generator because you usually write: 
`for terms, coeff in ham.parse(): ...`

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+        if not self.elements:
+            return "0"
+
+        eps = 1e-14
+
+        def _format_coeff(c: complex) -> str:
+            re, im = c.real, c.imag
+
+            # 1) Purely real?
+            if abs(im) < eps:
```
amirazzam2000: we can replace this with `np.real_if_close` which is used other places in the code

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+        Returns:
+            int: the number of qubits
+        """
+        n_qubits = -1
+        for key in self._elements:
+            for qid, _ in key:
+                n_qubits = max(n_qubits, qid)
+
+        if n_qubits == -1:
+            raise InvalidHamiltonianOperation("Can't compute the number of qubits if the hamiltonian has no terms")
```
amirazzam2000: this should return 0 instead of raising this error.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+                    # e.g. base_str = '-3+2j' => we want " - 3+2j" or just "- 3+2j" ?
+                    coeff_str = f"- {base_str[1:]}" if len(base_str) > 1 else "-"
+                else:
+                    coeff_str = f"- {base_str}"
+            else:
+                # Positive
+                coeff_str = "+" if base_str == "1" else f"+ {base_str}"
+
+            # Build the operators string (e.g. "Z(0) Y(1)")
+            ops_str = " ".join(f"{op}({qid})" for qid, op in operators)
```
amirazzam2000: can we skip printing I(0)? and add the constant energy term first. It shouldn't be very complicated to do maybe we add one extra flag.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+                else:
+                    coeff_str = base_str
+            # Subsequent terms: show '+' or '-'
+            elif negative:
+                # Remove leading '-' if present
+                if base_str == "-1":
+                    coeff_str = "-"
+                elif base_str.startswith("-"):
+                    # e.g. base_str = '-3+2j' => we want " - 3+2j" or just "- 3+2j" ?
+                    coeff_str = f"- {base_str[1:]}" if len(base_str) > 1 else "-"
```
amirazzam2000: there is an error with this example: `h = 1 + Z(0) - 3 + Z(1) + Z(0) - 2j` it prints `H = 2 Z(0) - (-2-2j) I(0) + Z(1)` there is a double - sign.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+
+    def to_hamiltonian(self) -> Hamiltonian:
+        """Helper function to convert to Hamiltonian representation.
+
+        Returns:
+            Hamiltonian: a hamiltonian with the pauli operator stored in it.
+        """
+        return Hamiltonian({((self.qubit, self.name),): 1})
+
+    # TODO(fedonman, #3): Do we need this? Why is it a Generator?
```
amirazzam2000: this can be removed

src/qilisdk/analog/hamiltonian.py
amirazzam2000: can be removed

src/qilisdk/analog/hamiltonian.py
amirazzam2000: can be mapped to `__iter(self)__`

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,520 @@
...
+        if not isinstance(other, Complex):
+            raise InvalidHamiltonianOperation("Division of operators is not supported")
+
+        return (1 / other) * self
+
+    def __rfloordiv__(self, other: Complex | PauliOperator | Hamiltonian) -> Hamiltonian:
+        if not isinstance(other, Complex):
+            raise NotSupportedOperation("Division of operators is not supported")
+
+        return (1 / other) * self
```
amirazzam2000, line 562: add `__eq__(self, other)`

src/qilisdk/analog/hamiltonian2.py
```diff
@@ -0,0 +1,582 @@
...
+        if not isinstance(other, (int, float, complex)):
+            raise NotSupportedOperation("Division of operators is not supported")
+        return (1 / other) * self
+
+    def __itruediv__(self, other: Complex | PauliOperator | Hamiltonian) -> Self:
+        self._div_inplace(other)
+        self.simplify()
+        return self
+
+    def _add_inplace(self, other: Complex | PauliOperator | Hamiltonian) -> None:
```
amirazzam2000: I don't think we should allow for this. This will generate a lot of bugs. This was initially how addition worked and I quickly realized this was causing lots of issues. If your argument is that += should be in place then I would suggest not supporting the iadd ...etc. rather than making it affect the exact object.

fedonman: I will return it to previous implementation: we will support += but return a new object like +.

fedonman: __hash__ will be guaranteed to work too, since our hamiltonians will be essentially immutable.

tests/analog/test_hamiltionian.py
DZPM, line 1: Why the `hamiltonian2` module name?

amirazzam2000: We have currently two implementations of the Hamiltonian class and we are comparing them. But once this pull request is done there will be only one of the two 👍

src/qilisdk/analog/hamiltonian.py
```diff
@@ -273,12 +277,16 @@ def __hash__(self) -> int:
...
         return Hamiltonian(elements=self.elements.copy())
 
+    # ------- String representation --------
+
     def __repr__(self) -> str:
+        # Developer-friendly representation
+        # return f"Hamiltonian({dict(self._elements)})"
         return str(self)
 
     def __str__(self) -> str:
```
amirazzam2000, line 284: I am still getting an issue with the print: look at this example `h = 1 + Z(0) - 3 * Z(1) + Z(1) + Z(0) - 2j * Z(1)`

image

src/qilisdk/analog/hamiltonian.py
```diff
@@ -273,12 +277,16 @@ def __hash__(self) -> int:
...
         # Return "0" if there are no terms
-        if not self.elements:
+        if not self._elements:
             return "0"
 
         eps = 1e-14
```
amirazzam2000: this eps should be replaced by the global _EPS

src/qilisdk/analog/hamiltonian.py
```diff
@@ -288,227 +296,281 @@ def _format_coeff(c: complex) -> str:
...
             s = str(c)  # e.g. '(3+2j)'
             return s
 
-        parts = []
-        items = list(self.elements.items())
-
+        # We want to place the single identity term (I(0),) at the front if it exists
+        items = list(self._elements.items())
         try:
-            # Find index of first item whose key has exactly one tuple and the operator == TARGET_OP
```
amirazzam2000: if we always know that `I(0)` is the one storing the constants then can't we just check if `self.elements[I(0)] != 0` ? and then can we find it's index by doing something like `items.index(((I(0),), self.elements[(I(0), )]))` and pop it. I don't know if that would be more efficient or not.

src/qilisdk/analog/hamiltonian.py
fedonman: Check now, I think it is fixed. Added a test for this case too.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,663 @@
...
+        """Returns the internal dictionary of elements (read-only)."""
+        return self._elements
+
+    def simplify(self) -> Hamiltonian:
+        # 1) Remove near-zero
+        keys_to_remove = [key for key, value in self._elements.items() if abs(value) < Hamiltonian._EPS]
+        for key in keys_to_remove:
+            del self._elements[key]
+
+        # 2) Accumulate identity operators that do NOT act on qubit=0 => I(0)
```
amirazzam2000: I would change this sentence to just like "Accumulate identities". because technically the identity is applied to all qubits.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -0,0 +1,663 @@
...
+            # scale all coefficients
+            for k in self._elements:
+                self._elements[k] *= other
+            return None
+
+        if isinstance(other, PauliOperator):
+            # Convert single PauliOperator -> Hamiltonian with 1 key
+            # Then do the single-key Hamiltonian path below
+            tmp = Hamiltonian({(other,): 1})
+            other = tmp  # type: ignore
```
amirazzam2000: why not just directly say `other = Hamiltonian({(other,): 1})`?

tests/analog/test_hamiltionian.py
```diff
@@ -0,0 +1,140 @@
...
+        (Z(0) * 3, 3 * Z(0)),
+        ((Z(0) + Z(1)) * 2, 2 * Z(0) + 2 * Z(1)),
+        ((Z(0) + X(1)) * (Z(0) + X(1)), 2 + 2 * (Z(0) * X(1))),
+        ((Z(0) + X(1)) * (Z(1) + X(0)), (Z(0) * Z(1)) + 1j * Y(0) + (-1j) * Y(1) + (X(0) * X(1))),
+        ((Z(0) + X(0)) * (Z(0) - X(0)), -2j * Y(0)),
+        (
+            (Z(0) + 1) * (1j * X(0) * X(1) + 1),
+            # It's perfectly valid just to keep it as an expression, or expand it out.
+            # We'll keep it as an expression for brevity.
+            (Z(0) + 1) * (1j * X(0) * X(1) + 1),
```
amirazzam2000, line 72: my main concern is if the unfolding breaks we won't see it with this test.

fedonman: What do yo mean? PS. I can add any extra test needed.

amirazzam2000: maybe in this case we are comparing to equal things, but they are exactly equal. so it could be better to compare `(Z(0) + 1) * (1j * X(0) * X(1) + 1)` with `1 - Y(0) X(1) + Z(0) + 1j X(0) X(1)`. just to make sure when we unfold the parenthesis it's working properly and we are not messing that up
.

src/qilisdk/analog/hamiltonian.py
fedonman: Done

tests/analog/test_hamiltionian.py
fedonman: Added the test.

src/qilisdk/analog/hamiltonian.py
fedonman: Done

src/qilisdk/analog/hamiltonian.py
fedonman: I think index is slower. I changed the check to `key[0] == (I(0))` though which is clearer.

src/qilisdk/analog/hamiltonian.py
fedonman: done

amirazzam2000 (approved):

PR #4, opened 2025-02-24
[QHC-912] Implement Optimizer and SciPyOptimizer

amirazzam2000:
Added the `Optimizer` abstract base class and its concrete subclass `SciPyOptimizer`. The `SciPyOptimizer` class wraps `scipy.optimize.minimize` to optimize cost functions while supporting extra keyword arguments such as Jacobian, bounds, etc. This implementation provides a structured way to perform optimization and access optimal parameters via the `optimal_parameters` property.

### Code Example

```python
from optimizer import SciPyOptimizer

# Define a simple cost function, e.g., a quadratic function with minimum at [1, 1, 1]
def cost_function(params):
    return sum((p - 1) ** 2 for p in params)

# Create an instance of the SciPyOptimizer using the BFGS method.
optimizer = SciPyOptimizer(method="BFGS")

# Optimize the cost function starting from an initial guess.
initial_parameters = [0, 0, 0]
result = optimizer.optimize(cost_function, initial_parameters)

# Print the result and the optimal parameters stored in the instance.
print("Optimization result:", result)
print("Optimal Parameters:", optimizer.optimal_parameters)
```


src/qilisdk/common/optimizer.py
```diff
@@ -13,4 +13,112 @@
...
+            the number of parameters the optimizer will consider.
+
+        Returns:
+            list[float]: the optimal set of parameters that minimize the cost function.
+        """
+        res = sci_opt.minimize(
+            cost_function,
+            x0=init_parameters,
+            method=self.method,
+            jac=self.extra_arguments.pop("jac", None),
```
GuillermoAbadLopez: should you maybe do a `get` instead of a `pop`, in case the method `optimize` is called more than one from the same instance?
```suggestion
            jac=self.extra_arguments.get("jac", None),
```

( ❗ same applies in all the following lines)

GuillermoAbadLopez (changes requested):
Looks great, gj! 🚀 

Just one thing:

GuillermoAbadLopez:
❗The change of the `pop` -> `get` should be applied to more lines too :)

fedonman:
> ❗The change of the `pop` -> `get` should be applied to more lines too :)

OK, will check it. The PR is far from mergable, anyway, we have to write tests, split the classes etc.

GuillermoAbadLopez:
Ohh okey, I just saw two notificacions of merging my suggested commit and main, and thought you were going for it 😅

Cool 👍

src/qilisdk/common/optimizer.py
amirazzam2000: yeah you are right get would make more sense. I will change it 👍

fedonman (approved):

GuillermoAbadLopez (approved):
Gj!

PR #5, opened 2025-02-25
Implement QaaSBackend

fedonman:
This release introduces a **draft** version of the new `QaaSBackend` as part of the optional `qaas` module (install with `pip install qilisdk[qaas]`). The `QaaSBackend` enables users to interface with Qilimanjaro's cloud-based Quantum-as-a-Service (QaaS) platform via synchronous HTTP calls, providing a unified backend that supports both digital and analog workflows.

In this update, the backend allows for secure login and logout operations using credentials sourced through method parameters or environment variables. The returned authenticated token is stored securily with `keyring`.

`QaaSBackend` implements both `DigitalBackend` and `AnalogBackend` interfaces, paving the way for future extensions that support executing digital circuits and performing analog evolutions. It also offers functionalities such as list available devices.


src/qilisdk/extras/qaas/qaas_backend.py
```diff
@@ -0,0 +1,153 @@
...
+        Normally, you won't call __init__() directly.
+        Instead, use QaaSBackend.login(...) to create a logged-in instance.
+        """  # noqa: DOC501
+        credentials = load_credentials()
+        if credentials is None:
+            raise RuntimeError(
+                "No valid QaaS credentials found in keyring."
+                "Please call QaaSBackend.login(username, apikey) or ensure environment variables are set."
+            )
+        self._username, self._token = credentials
```
amirazzam2000, line 72: should we log in directly if we find the credentials stored?

src/qilisdk/extras/qaas/qaas_digital_result.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def samples(self) -> dict[str, int]:
+        return dict(self._samples)
+
+    def get_probabilities(self) -> dict[str, float]:
+        probabilities = {}
+
+        # Generate all possible bitstrings of the inferred length.
+        for bits in product("01", repeat=self.nqubits):
```
amirazzam2000: why don't we just loop around the samples instead? we don't need to show the probabilities of bitstrings that don't appear in the final sample. I guess this also applies to CudaBackend results as well.

src/qilisdk/extras/qaas/qaas_backend.py
fedonman: Yes, if you are already logged-in we retrieve the username and token from the keyring (the secure OS-specific password manager)

src/qilisdk/extras/qaas/qaas_digital_result.py
fedonman: I think I did this to match Qibo's behaviour, but I will change it.

src/qilisdk/extras/qaas/qaas_backend.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Manages communication with a hypothetical QaaS service via synchronous HTTP calls.
+
+    Credentials to log in can come from:
+      a) method parameters,
+      b) environment (via Pydantic),
+      c) keyring (fallback).
+    """
+
+    _api_url: str = "https://qilimanjaroqaas.ddns.net:8080/api/v1"
```
amirazzam2000, line 57: I don't know if the final decision was to move these to another location so that they are not just in the code.

amirazzam2000 (approved):
other than the URLs comment I think this looks very good! thanks for the changes!

PR #8, opened 2025-03-05
[QHC-895] Proposals for reducing `qibo_backend.py`

GuillermoAbadLopez:
## TODOs this PR:
**Reduced the source code of the `qibo_backend.py`, 2 options:** 
  - **First option:** Did a smaller change in the original file, only reducing duplication of info
  - **Second option:** Also added a 2nd option (in an extra file), much more different (similar to `app-sdk` repo)

(of course we will only keep 1 option before merging)

---

### Further Comment 
Both of my impIementations don't pass the specific `type` of `Gate`, for `mypy` to check, just the generic `Gate` `type`... 

but the current implementation (if I'm not mistaken) didn't do it either @fedonman, we were losing the precise `type` of `Gate` in this function call already, so no problem from this side 👍 :

Screenshot 2025-02-21 at 19 58 43


GuillermoAbadLopez:
@fedonman  @amirazzam2000 , after some thought, I think the 2nd option (`qibo_backend2.py`), becomes harder now that we might have different matrices/global_phases in different backends 🤔, it could also be done, but I think this approach wants to make it `too` short, and will end up complicating our lives...

Having the transformation explicitly for each case, is longer, but seems more robust, and easier to maintain 👌 
(Still the first reduction [`qibo_backend.py`] of the mapping could be done 🙌 )

fedonman (changes requested):
I don't like any of the two proposals tbh :face_in_clouds: 

`qibo_backend2.py` definitely not, because it uses too much custom logic with no reason (like if `if gate_type in {U2, U3}:` which is not scalable at all.

`qibo_backend.py` is better, but still I cannot find a reason to abandon the strong-typed mapping and try to retrieve the method's name on runtime using custom logic again (get the gate's name and assume the method has the same suffix).

In general, I want us to globally embrace the pattern `dict[type, Callable]`. We also use it in Qililab's QbloxCompiler and QuantumMachinesCompiler. It is super performant, has strong-typing and is very easy to read and navigate (e.g. ctrl+click works, renaming works and so on)

GuillermoAbadLopez:
> qibo_backend2.py definitely not, because it uses too much custom logic with no reason (like if if gate_type in {U2, U3}: which is not scalable at all.

After further checking it these days, I agree yess, I'm deleting this one now 👍 

> `qibo_backend.py` is better, but still I cannot find a reason to abandon the strong-typed mapping 

In this case though, I don't see how we are losing any strong `typing`, since it wasn't there to begin with, no? (check my "further comment" on this PR description). You always get as an output of `typing` the General `QiboGates.Gate` in both the previous implementation and this new one (not the specific `qibo` child `gate`, in any case).

> and try to retrieve the method's name on runtime using custom logic again (get the gate's name and assume the method has the same suffix).

I think this is a fair assumption, since it will be us, who name the method, so we can always fulfil this, in the previous implementation, you would also need to come here, and add the logic to the map. 

And respect the using custom logic, I believe it is there now anyway, I'm just changing it to automatically retrieving it from the gate name, instead of having to explicitly write it in the map 🤔 

It's a very silly reduction, that does:
- pros: removes lines of explicitly writing the `converter` names in the map (which is kind of info duplication)
- cons: limits the name of the `converter`s (which we were already fulfilling, and might even be a good practice having it standardized in a single way? 🤔)

So @fedonman, I don't wanna be annoying and push about merging this too much, since it's a really silly thing, so I'll be happy with what you think after this comment, let's do as you see 👍

fedonman:
@GuillermoAbadLopez Let's cancel this PR for now please, QiboBackend might get irrelevant pretty soon anyway (if we move with CudaBackend). In any case, I can explain more the `dict[type, Callable]` pattern in the code discussions if you are ok with that.

GuillermoAbadLopez:
Sure, let's do that 👍

And please tell me whenever you will talk about it, cause I won't be able to assist all Code Discussion now 😬

PR #9, opened 2025-03-05
[QHC-900] Change `U2` and `U3` matrix definitions and improve docs of them

GuillermoAbadLopez:
## This PR does:
- [x] Improve documentation expanding explanations of gates
- [x] Improve documentation adding comparison and example between gates
- [x] Change U2 and U3 to the standard matrix from `qiskit` and `pennylane` (specifying the difference of global phase with `QASM` [& therefore qibo])


src/qilisdk/digital/gates.py
```diff
@@ -477,16 +520,16 @@ def __init__(self, qubit: int, *, phi: float, lam: float) -> None:
...
         """
         super().__init__()
         self._target_qubits = (qubit,)
         self._parameter_values = [phi, lam]
         self._matrix = (1 / np.sqrt(2)) * np.array(
             [
-                [np.exp(-1j * (phi + lam) / 2), -np.exp(-1j * (phi - lam) / 2)],
-                [np.exp(1j * (phi - lam) / 2), np.exp(1j * (phi + lam) / 2)],
+                [1, -np.exp(1j * lam / 2)],
+                [np.exp(1j * phi / 2), np.exp(1j * (phi + lam))],
```
fedonman: This is from pennylane right? It produces different results from Qibo. I don't know which definition is the best one, but if we choose pennylane's we have to change the `QiboBackend` as well to create the equivalent gate.

image

fedonman (changes requested):
I super like the changes in documentation. :100: I have a critical comment on the U2 matrix definition. Also, let's add the change of `lam` -> `gamma` before merging.

GuillermoAbadLopez:
> I super like the changes in documentation. 💯 I have a critical comment on the U2 matrix definition. Also, let's add the change of `lam` -> `gamma` before merging.

@fedonman just merged the changes from `lam` to `gamma` 🙌 

And yes, the new `U2` definition is the same as `qiskit`, `pennylane` & `cuda`, which is different from `qibo`.
I saw that this change was already done for `U3` (but not its documentation), so I changed the docs there too.

src/qilisdk/digital/gates.py
GuillermoAbadLopez: Ouu, thats true, we need to add the global phase 😬, lemme check it 👌

GuillermoAbadLopez: @fedonman I've been checking, and since the difference is only a global phase, it shouldn't matter, both gates are equivalent, and there should be no physical difference in the results, even though their matrices are different :)

(I've also checked how to add a `GlobalPhase` if we wanted it anyway, and I found ways in `Qiskit`, but not in `Qibo`..., so this helps even more on deciding to not implementing it, an keeping the equivalent 🙃)

fedonman (approved):
:100: :100:

PR #11, opened 2025-03-07
[QHC-894] Implement CudaBackend

fedonman:
Introduces a new backend for quantum circuit simulation leveraging CUDA. The backend supports multiple simulation methods:
  - `STATE_VECTOR` (automatically selects GPU if available, otherwise falls back to CPU)
  - `TENSOR_NETWORK`
  - `MATRIX_PRODUCT_STATE`

  The `CudaBackend` handles basic gates (X, Y, Z, H, S, T, RX, RY, RZ, U1, U2, U3), adjoint gates, controlled gates with one control qubit (we found a bug for more qubits, see https://github.com/NVIDIA/cuda-quantum/issues/2731), and measurement operations.

**Example usage:**

```python
from qilisdk.extras.cuda_backend import CudaBackend
from qilisdk.digital import Circuit, H, M, DigitalSimulationMethod

# Create a quantum circuit with one qubit.
circuit = Circuit(nqubits=1)
circuit.add(H(0))  # Apply a Hadamard gate.
circuit.add(M(0))  # Measure the qubit.

# Initialize the CudaBackend with the TENSOR_NETWORK simulation method.
backend = CudaBackend(simulation_method=DigitalSimulationMethod.TENSOR_NETWORK)

# Execute the circuit with 1000 shots and print the results.
results = backend.execute(circuit, nshots=1000)
print("Results:", results)
```


src/qilisdk/digital/digital_result.py
```diff
@@ -13,69 +13,56 @@
...
 
-import numpy as np
-
 from qilisdk.common import Results
 
 
-class DigitalResults(Results):
-    def __init__(self, samples: np.ndarray, frequencies: dict[str, int], nshots: int) -> None:
+class DigitalResult(Results):
+    def __init__(self, nshots: int, samples: dict[str, int]) -> None:
```
amirazzam2000: I think it's better to add more options here than remove them. We can add more parameters with a default value, and then when the user tries to retrieve the information we can only return the ones that have values. I only say this to try and streamline the results object for different possibilities.

src/qilisdk/extras/cuda_backend.py
```diff
@@ -0,0 +1,178 @@
...
+            T: CudaBackend._handle_T,
+            RX: CudaBackend._handle_RX,
+            RY: CudaBackend._handle_RY,
+            RZ: CudaBackend._handle_RZ,
+            U1: CudaBackend._handle_U1,
+            U2: CudaBackend._handle_U2,
+            U3: CudaBackend._handle_U3,
+        }
+
+    def _apply_simulation_method(self) -> None:
```
amirazzam2000: missing method docstrings for all methods in this class

src/qilisdk/extras/qibo_backend.py
amirazzam2000, line 1: Before we remove the file completely we should create the extra library so that we don't lose the code we had here and have to look for it in old commits ...etc.

src/qilisdk/extras/cuda_backend.py
fedonman: Done.

src/qilisdk/digital/digital_result.py
fedonman: Done.

amirazzam2000 (approved):

PR #12, opened 2025-03-08
[QHC-903] Add support for OpenQASM 2.0

fedonman:
Implement utility functions `to_qasm2`, `to_qasm2_file`, `from_qasm2` and `from_qasm2_file` to serialize and deserialize a `Circuit` using OpenQASM 2.0 grammar.

### Code Example

```python
from qilisdk.digital.circuit import Circuit
from qilisdk.digital.gates import CNOT, RX, H, M
from qilisdk.utils.openqasm2 import from_qasm2, from_qasm2_file, to_qasm2, to_qasm2_file

# Create a sample circuit.
circuit = Circuit(3)
circuit.add(H(0))
circuit.add(CNOT(0, 1))
circuit.add(RX(2, theta=3.1415))
circuit.add(M(0, 1, 2))

# Convert to QASM.
qasm_code = to_qasm2(circuit)
print("Generated QASM:")
print(qasm_code)

# Reconstruct the circuit from the QASM.
reconstructed_circuit = from_qasm2(qasm_code)

# Convert to QASM and save to file
to_qasm2_file(circuit, "circuit.qasm")

# Reconstruct the circuit from the QASM file.
reconstructed_circuit = from_qasm2_file("circuit.qasm")
```


amirazzam2000:
I tried this [example ](https://docs.quantum.ibm.com/guides/interoperate-qiskit-qasm2) from Qiskit 
image

and in our case the measurement gates are not added:
image

I don't know if we should take into consideration this case when the measure command is given from a quantum register to a classical register without specifying the qubit/bit number specifically. i.e.:

```
measure q -> c;
```

instead of 

```
measure q[0] -> c[0];
measure q[1] -> c[1];
```

fedonman:
@amirazzam2000 I added support for this syntax when measuring all qubits. Also I added a shorthand `M()` for `M(*range(circuit.nqubits))` that both mean measure all qubits.

changes/12.feature.md
GuillermoAbadLopez, line 1: ### The only thing I found, is the difference in the `reconstructed circuit` between:

measuring all the qubits             |  only some of them
:-------------------------:|:-------------------------:
Screenshot 2025-03-12 at 16 08 13 | Screenshot 2025-03-12 at 16 07 31

### where an `M` gate with several qubits is changed -> for several `M` gates of 1 qubit

🤔🤔🤔

src/qilisdk/utils/openqasm2.py
```diff
@@ -0,0 +1,209 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import re
+from pathlib import Path
+
+from qilisdk.digital.circuit import Circuit
+from qilisdk.digital.exceptions import UnsupportedGateError
+from qilisdk.digital.gates import CNOT, CZ, RX, RY, RZ, U1, U2, U3, Gate, H, M, S, T, X, Y, Z
+
+OPENQASM2_MAP: dict[type[Gate], str] = {
```
GuillermoAbadLopez, line 21: I guess this will have conflicts with: https://github.com/qilimanjaro-tech/qilisdk/pull/14, since you will need to add the logic for the controlled ones, in this or the other PR.

Just to keep it in mind 👍

src/qilisdk/utils/openqasm2.py
```diff
@@ -0,0 +1,209 @@
...
+    U3: "u3",
+    CNOT: "cx",
+    CZ: "cz",
+}
+
+
+def to_qasm2(circuit: Circuit) -> str:
+    """
+    Convert the circuit to an OpenQASM 2.0 formatted string.
+
```
GuillermoAbadLopez, line 42: Missing argument docs

src/qilisdk/utils/openqasm2.py
```diff
@@ -0,0 +1,209 @@
...
+                parameters = ", ".join(str(p) for p in gate.parameter_values)
+                param_str = f"({parameters})"
+            # Format qubit operands.
+            qubit_str = ", ".join(f"q[{q}]" for q in gate.qubits)
+            qasm_lines.append(f"{qasm_name}{param_str} {qubit_str};")
+
+    return "\n".join(qasm_lines)
+
+
+def to_qasm2_file(circuit: Circuit, filename: str) -> None:
```
GuillermoAbadLopez, line 82: missing filename argument docs

src/qilisdk/utils/openqasm2.py
```diff
@@ -0,0 +1,209 @@
...
+    Save the QASM representation to a file.
+
+    Args:
+        filename (str): The path to the file where the QASM code will be saved.
+    """
+    qasm_code = to_qasm2(circuit)
+    Path(filename).write_text(qasm_code, encoding="utf-8")
+
+
+def from_qasm2(qasm_str: str) -> Circuit:
```
GuillermoAbadLopez, line 95: Maybe we can split such a long function into calls of smaller ones? For example the handling of `measurements` and the handling of pattern `m`, its a bit hard to follow quickly.

src/qilisdk/utils/openqasm2.py
```diff
@@ -0,0 +1,209 @@
...
+
+def to_qasm2_file(circuit: Circuit, filename: str) -> None:
+    """
+    Save the QASM representation to a file.
+
+    Args:
+        filename (str): The path to the file where the QASM code will be saved.
+    """
+    qasm_code = to_qasm2(circuit)
+    Path(filename).write_text(qasm_code, encoding="utf-8")
```
GuillermoAbadLopez, line 91: very clean implementation, nice!

tests/utils/test_openqasm2.py
```diff
@@ -0,0 +1,151 @@
...
+
+def test_from_qasm2_unsupported_gate():
+    """Test that an unsupported gate in QASM raises an UnsupportedGateError."""
+    # Create a QASM string with an unsupported gate "foo".
+    qasm_str = "\n".join(["OPENQASM 2.0;", 'include "qelib1.inc";', "qreg q[1];", "foo q[0];"])
+    with pytest.raises(UnsupportedGateError, match="Unknown gate: foo"):
+        from_qasm2(qasm_str)
+
+
+def test_to_qasm2_file_and_from_qasm2_file(tmp_path):
```
GuillermoAbadLopez, line 135: nice 🔥

tests/utils/test_openqasm2.py
```diff
@@ -0,0 +1,151 @@
...
+    return circuit
+
+
+# --- Test cases ---
+def test_empty_circuit_to_qasm():
+    """Test conversion of an empty circuit (with no gates)."""
+    circuit = create_empty_circuit(nqubits=3)
+    qasm_str = to_qasm2(circuit)
+    # Check header lines and qreg declaration.
+    assert "OPENQASM 2.0;" in qasm_str
```
GuillermoAbadLopez: Maybe check also the order of the lines?
(or check the full str, which would include ordering?)

--
(Not very important, depending on the energy you have 🙃)

changes/12.feature.md
GuillermoAbadLopez: ### Also tested it, for translating to `qiskit` and it worked nicely 🙌:

measuring all the qubits             |  only some of them
:-------------------------:|:-------------------------:
Screenshot 2025-03-12 at 16 28 37 | Screenshot 2025-03-12 at 16 32 33

✅✅✅

GuillermoAbadLopez (changes requested):
Great! 🚀 

Left a couple comments on docstrings, and a couple other non-very important comments, gj 👍

changes/12.feature.md
fedonman: When you measure whole circuit you can do in QASM2: `measure q -> c` but when you choose less qubits you have to add one line per each `measure q[0] -> c[0]`, `measure q[1] -> c[1]` etc. When loading from QASM2 we don't check if there are consecutive measurements to combine them into one, but we just add a measurement for each line. I can add it as a `TODO` for future improvement.

src/qilisdk/utils/openqasm2.py
fedonman: I agree with you, but would leave it for future improvement, especially because the converter is not yet complete (e.g. it doesn't support all QASM 2.0 gates and syntax, just the ones we currently have in circuit) so it will probably change pretty soon (when we have time). So we can refactor it better when we know the final implementation.

tests/utils/test_openqasm2.py
fedonman: done

changes/12.feature.md
GuillermoAbadLopez: From my side, it is not necessary, just wanted to raise it in case you hadn't noticed, and consider it important. 

So up to you! 🚀

fedonman: Added `TODO`.

src/qilisdk/utils/openqasm2.py
fedonman: done

src/qilisdk/utils/openqasm2.py
GuillermoAbadLopez: Right, its going to be much quicker to change now, that is true.

Maybe we can leave it as a `TODO:` here? Or not, as you wish, we will probably remember 👍

GuillermoAbadLopez (approved):
Nice! Thanks for the changes!

Just reopened one comment, with an answer for a possible `TODO:` to add, and that's it, gj!

src/qilisdk/utils/openqasm2.py
fedonman: Added a todo.

amirazzam2000 (approved):

PR #14, opened 2025-03-11
[QHC-907] Enhanced Gate Framework: Controlled, Adjoint, and Exponential modifiers for Unitary Gates

fedonman:
**Enhanced Gate Framework**

This PR refactors the quantum gate hierarchy to provide a more modern and flexible framework. The following key improvements were introduced:

1. **Adjoint Operation:**
    A new `Adjoint` class has been introduced to provide the Hermitian conjugate (adjoint) of any unitary gate. When you call the `adjoint()` method on a unitary gate, it returns an instance of `Adjoint` whose matrix is the conjugate transpose of the original gate's matrix. This operation is essential for many quantum algorithms that require reversing a gate's effect.

    *Example:*
    ```python
    x_adjoint = X(0).adjoint()
    print(x_adjoint.name)  # Output: X†
    print(x_adjoint.matrix)  # Displays the conjugate-transposed matrix of X
    ```

2. **Exponential Operation:**
    The new `Exponential` class computes the matrix exponential of a unitary gate's matrix. This feature is particularly useful for simulating continuous time evolution in quantum systems.

    *Example:*
    ```python
    x_exponential = X(0).exponential()
    print(x_exponential.name)  # Output: e^X
    print(x_exponential.matrix)  # Displays the matrix exponential of X's matrix
    ```

3. **Controlled Operation:**
    The refactored framework now includes a generic `Controlled` class that wraps around any unitary gate to add control qubits. When you invoke the `controlled()` method on a unitary gate, you get a new controlled gate instance that:
    - Checks for duplicate control qubits.
    - Ensures that the control qubits do not overlap with the target qubits of the underlying gate.
    - Preserves the type of the controlled gate using generics, providing better static type safety.

    *Example:*
    ```python
    # Creating an X gate and then its controlled version with control qubits 1 and 2.
    controlled_x = X(0).controlled(1, 2)
    print(controlled_x.name)  # Output: CCX
    print(controlled_x.matrix)  # Displays the controlled gate matrix computed from X's matrix
    ```

4. **Dynamic Matrix Updates:**
    With the new framework, the `matrix` property of all modified operations (Adjoint, Exponential, Controlled) is always up-to-date. Whenever you update the parameters of a gate (via `set_parameters` or `set_parameter_values`), the underlying matrix is re-generated automatically. This ensures that any dependent modified gate also reflects the updated parameters.

    *Example:*
    ```python
    # Create an RX gate with an initial theta value.
    rx_gate = RX(0, theta=3.14)
    print("Initial RX matrix:")
    print(rx_gate.matrix)

    # Update the rotation parameter.
    rx_gate.set_parameter_values([1.57])
    print("Updated RX matrix:")
    print(rx_gate.matrix)

    # If you create a controlled version, its matrix will update as well:
    controlled_rx = RX(0, theta=3.14).controlled(1)
    print("Controlled-RX matrix:")
    print(controlled_rx.matrix)

    # Update the rotation parameter.
    controlled_rx.set_parameter_values([1.57])
    print("Updated Controlled-RX matrix:")
    print(controlled_rx.matrix)
    ```

5. **Generic and Type-Safe Enhancements:**
    The new design leverages Python generics (with a type variable, e.g., `TUnitary`) to ensure that operations such as `adjoint`, `exponential`, and `controlled` correctly tie the modified gate to its underlying unitary type. This enhancement allows static type checkers to catch mismatches early and provides a cleaner, more robust API.

6. **Unified Method Signatures for Gate Modifications:**
    The `Unitary` class now provides convenience methods for creating these modified gates:
    - `controlled(*control_qubits: int) -> Controlled[Self]`
    - `adjoint() -> Adjoint[Self]`
    - `exponential() -> Exponential[Self]`

    These methods allow you to easily derive new gate instances from existing ones while keeping the underlying type information intact.

3. **Unified and Immutable Parameter Handling:**
    The new `Unitary` class now accepts a `parameters` dictionary during initialization and provides a read-only property that returns these values. The properties `parameter_names` and `parameter_values` are computed from the underlying parameters rather than being stored separately.

    This makes it simpler to update and validate parameter values:
    ```python
    rx_gate = RX(0, theta=3.14)
    print(rx_gate.parameters)  # Output: {'theta': 3.14}
    rx_gate.set_parameter_values([1.57])
    ```

**Impact:**
These features collectively enhance the modularity and expressiveness of our quantum gate framework. Users can now effortlessly generate new gate variants (controlled, adjoint, and exponential) with robust type safety and clear, intuitive API calls, laying a solid foundation for advanced quantum circuit design.


src/qilisdk/digital/__init__.py
GuillermoAbadLopez, line 1: Like I said in https://github.com/qilimanjaro-tech/qilisdk/pull/12, depending on which one gets merged before, we might need to add the logic for the controlled ones, in the QASM mapping, maybe?

Just to keep it in mind 👍

tests/digital/test_gates.py
GuillermoAbadLopez, line 1: Maybe we should add tests to cover the `_generate_matrix()` of `Exponential`, `Adjoint`, etc.. At least to the `Controlled` on, which is the only complicated one I would argue? 🤔 

And in general, add tests to all this new logic? Or will that be done later on?

src/qilisdk/digital/gates.py
```diff
@@ -11,41 +11,56 @@
...
+    @abstractmethod
     def name(self) -> str:
         """
         Retrieve the name of the gate.
 
         Returns:
             str: The name of the gate.
         """
-        return self.NAME
+        ...
```
GuillermoAbadLopez: I think with `@abstract` you don't need the there dots `...`

src/qilisdk/digital/gates.py
```diff
@@ -161,13 +177,244 @@ def set_parameter_values(self, values: list[float]) -> None:
...
+        self._matrix = self._generate_matrix()
+
+    @abstractmethod
+    def _generate_matrix(self) -> np.ndarray: ...
+
+    def is_modified_from(self, gate_type: type[TBasicGate]) -> bool:
+        return isinstance(self.unitary_gate, gate_type)
+
+
+class Controlled(Modified[TBasicGate]):
```
GuillermoAbadLopez, line 328: Amazing added logic, love it! 🔥

src/qilisdk/digital/gates.py
```diff
@@ -161,13 +177,244 @@ def set_parameter_values(self, values: list[float]) -> None:
...
+
+    def set_parameter_values(self, values: list[float]) -> None:
+        super().set_parameter_values(values=values)
+
+        for key, value in zip(self.parameters, values):
+            self._parameters[key] = value
+
+        self._matrix = self._generate_matrix()
+
+    def controlled(self: Self, *control_qubits: int) -> Controlled[Self]:
```
GuillermoAbadLopez, line 231: This is amazing, gj

changes/14.feature.md
```diff
@@ -0,0 +1,87 @@
...
+    *Example:*
+    ```python
+    # Creating an X gate and then its controlled version with control qubits 1 and 2.
+    controlled_x = X(0).controlled(1, 2)
+    print(controlled_x.name)  # Output: CCX
+    print(controlled_x.matrix)  # Displays the controlled gate matrix computed from X's matrix
+    ```
+
+4. **Dynamic Matrix Updates:**
+    With the new framework, the `matrix` property of all modified operations (Adjoint, Exponential, Controlled) is always up-to-date. Whenever you update the parameters of a gate (via `set_parameters` or `set_parameter_values`), the underlying matrix is re-generated automatically. This ensures that any dependent modified gate also reflects the updated parameters.
```
GuillermoAbadLopez, line 40: cool! 🔥

GuillermoAbadLopez (commented):
Such a great work, amazing PR Vyron!

--

The only thing I would say, is the lack of testing for all this new logic 😬...

1) I would argue that there is some complex logic in the PR, that would be much better if it were tested, for protecting ourselves from future fucked ups more than anything. At least in the `Controlled._generate_matrix` would be nice.

2) Maybe even a small integration test would be great in this case, checking a Modified gate works nicely, and it relates correctly to the previous etc..

tests/digital/test_gates.py
fedonman: Added lots of tests!

src/qilisdk/digital/gates.py
fedonman: You are right. Removed them.

src/qilisdk/digital/__init__.py
fedonman: Yes, I added a TODO in #12 to continue working on the QASM2.

GuillermoAbadLopez (approved):
That's great, thanks for the tests, amazing PR 🚀

PR #16, opened 2025-03-12
Analog algorithms

amirazzam2000:
This release introduces the new `TimeEvolution` analog algorithm along with supporting classes that empower users to simulate time-dependent quantum systems with greater flexibility and precision. The update brings a schedule-based approach that allows for dynamic interpolation between different Hamiltonians over time, making it easier to model complex evolution scenarios. New classes such as `Schedule` and `TimeEvolution` simplify the process of defining the evolution parameters, specifying initial states, and monitoring observables. For instance, the following code example demonstrates how to set up a one-qubit system with two Hamiltonians—gradually transitioning from one to the other using a linear schedule—and execute the simulation on a CUDA backend for enhanced performance:

```python
T = 10
dt = 0.1
steps = np.linspace(0, T, int(T / dt))

nqubits = 1

H1 = sum(X(i) for i in range(nqubits))
H2 = sum(Z(i) for i in range(nqubits))

schedule = Schedule(
    T,
    dt,
    hamiltonians={
        "h1": H1,
        "h2": H2
    },
    schedule={
        t: {
            "h1": 1 - steps[t] / T,
            "h2": steps[t] / T,
        }
        for t in range(len(steps))
    },
)

state = tensor([(ket(0) + ket(1)).unit() for _ in range(nqubits)]).unit()
dm = (state @ state.dag()).unit()

time_evolution = TimeEvolution(
    backend=CudaBackend(),
    schedule=schedule,
    initial_state=state,
    observables=[Z(0), X(0), Y(0), Z(nqubits - 1), X(nqubits - 1), Y(nqubits - 1)],
)

results = time_evolution.evolve(store_intermediate_results=True)
```


src/qilisdk/analog/analog_result.py
```diff
@@ -0,0 +1,62 @@
...
+# limitations under the License.
+from pprint import pformat
+
+import numpy as np
+
+from qilisdk.analog.quantum_objects import QuantumObject
+from qilisdk.common import Results
+
+
+class AnalogResults(Results):
```
fedonman: Is this the results we will get from every analog algorithm or just TimeEvolution? If it's the second, we should make `AnalogResults` one level higher and rename this to `TimeEvolutionResults`.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -219,9 +220,10 @@ def __init__(self, elements: dict[tuple[PauliOperator, ...], complex] | None = N
 
     @property
     def nqubits(self) -> int:
-        """Number of distinct qubits across all terms."""
+        """Number of qubits acting on the hamiltonian."""
         qubits = {op.qubit for key in self._elements for op in key}
-        return len(qubits)
+
+        return max(qubits) + 1
```
fedonman: If `H = Z(0) + X(2)` shouldn't we return 2? These are the number of qubits the hamiltonian is acting upon. The full system, yes, it might be 3 qubits but it could be even more.

fedonman (commented):
I am just leaving some comments as I read the code.

src/qilisdk/analog/hamiltonian.py
amirazzam2000: in this case this hamiltonian is acting on 3 qubits, 0, 1, 2. Even tho there are no present terms that act on qubit 1 it's assumed that it's the identity.

src/qilisdk/analog/algorithms.py
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(
+        self,
+        backend: AnalogBackend,
+        schedule: Schedule,
+        observables: list[PauliOperator | Hamiltonian],
+        initial_state: QuantumObject,
+        n_shots: int = 1000,
+    ) -> None:
+        """
+        Args:
```
GuillermoAbadLopez, line 82: missing backend argument in docstring

src/qilisdk/analog/algorithms.py
```diff
@@ -0,0 +1,59 @@
...
+        """
+        super().__init__(backend)
+        self.initial_state = initial_state
+        if not isinstance(self.initial_state, QuantumObject):
+            raise NotImplementedError("currently only QuantumObjects are accepted as initial states.")
+        self.schedule = schedule
+        self.observables = observables
+        self.n_shots = n_shots
+
+    def evolve(self, store_intermediate_results: bool = False, **kwargs: dict) -> AnalogResult:
```
GuillermoAbadLopez: Maybe add docstrings? You can add them with Copilot or some AI :)

src/qilisdk/analog/algorithms.py
```diff
@@ -0,0 +1,59 @@
...
+class TimeEvolution(AnalogAlgorithm):
+    def __init__(
+        self,
+        backend: AnalogBackend,
+        schedule: Schedule,
+        observables: list[PauliOperator | Hamiltonian],
+        initial_state: QuantumObject,
+        n_shots: int = 1000,
+    ) -> None:
+        """
```
GuillermoAbadLopez, line 79: Maybe add a small description of the class

src/qilisdk/analog/analog_backend.py
```diff
@@ -0,0 +1,44 @@
...
+
+class AnalogBackend(Backend):
+    @abstractmethod
+    def evolve(
+        self,
+        schedule: Schedule,
+        initial_state: QuantumObject,
+        observables: list[PauliOperator | Hamiltonian],
+        store_intermediate_results: bool = False,
+        **Kwargs: dict,
```
GuillermoAbadLopez:
```suggestion
        **kwargs: dict,
```

src/qilisdk/analog/algorithms.py
GuillermoAbadLopez: Well maybe that is better, under the class definition line, not here

src/qilisdk/analog/analog_result.py
```diff
@@ -0,0 +1,61 @@
...
+# limitations under the License.
+from pprint import pformat
+
+import numpy as np
+
+from qilisdk.analog.quantum_objects import QuantumObject
+from qilisdk.common import Result
+
+
+class AnalogResult(Result):
```
GuillermoAbadLopez, line 22: A quick explanation of the class, might be useful

src/qilisdk/analog/hamiltonian.py
GuillermoAbadLopez: I guess thats kind of like with `QuantumCircuits`, you assume all qubits exist till the maximum... Dunno if that's bad, I let you guys discuss that :)

src/qilisdk/analog/schedule.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def add_hamiltonian(
+        self, label: str, hamiltonian: Hamiltonian, schedule: Callable | None = None, **kwargs: dict
+    ) -> None:
+        """Adds a hamiltonian object to the list of hamiltonians known by the schedule
+
+        Args:
+            label (str): the label used by the schedule to identify the hamiltonian.
+            hamiltonian (Hamiltonian): the hamiltonian object.
+            schedule (Callable):
```
GuillermoAbadLopez: Missing an optional, and I guess, the jump of line is not intended?

src/qilisdk/analog/quantum_objects.py
```diff
@@ -0,0 +1,336 @@
...
+import numpy as np
+from scipy.sparse import csc_array, csr_matrix, issparse, kron, sparray, spmatrix
+from scipy.sparse.linalg import expm
+from scipy.sparse.linalg import norm as scipy_norm
+
+Complex = int | float | complex
+TWO = 2
+
+
+class QuantumObject:
```
GuillermoAbadLopez, line 28: Some docstring for the Class at least wouldn't hurt, and would help a lot for reviewing :)

GuillermoAbadLopez (changes requested):
Sorry @amirazzam2000,  I thought I was adding all the comments to a single review, instead, I created 17.000 separated comments, my bad 😅 😂 

I just did a quick first review, to see if I could understand the code quickly, but it's too long, so I ended up mostly reviewing documentation, I'll do a more in-depth of the `quantum_objects.py`, as you asked, later next week 👍

--
### Summarizing all the above comments:
If you could add some more documentation, to the main classes, since you have the context, it would help me much for the review, and posterior maintenance :)

fedonman (approved):

PR #17, opened 2025-03-12
Add U1 to U3 mapping in docstring

GuillermoAbadLopez:
### This PR does:
- Adds documentation about how to map `U3` to `U1` gates
- Changes the single \` for two `` for docstring codes in the `gates.py` file
- Adds a jump of line, for the representations of gates, so all gates have similar docstring structure and it's easier to read

Update:
- Also make matrices code blocks


fedonman (approved):
💯 💯

PR #21, opened 2025-04-07
[QHC-917] Implement VQE

fedonman:
This release introduces a new Variational Quantum Eigensolver (VQE) algorithm designed to approximate the ground state energy of quantum systems via a hybrid quantum–classical approach. The VQE algorithm now provides users with a structured and detailed output that includes the estimated ground state energy (optimal cost) and the corresponding optimal ansatz parameters. In addition, users can optionally enable the recording of intermediate optimization results—each stored as an instance of `OptimizerResult`—to facilitate detailed analysis and debugging of the optimization process. Notable enhancements in this release include:
- A new `VQE` class that integrates seamlessly with user-defined ansatzes, backends, and cost functions, and produces comprehensive results encapsulated in a `VQEResult`.
- A detailed `VQEResult` output that contains both the final optimal values and, if requested, a list of intermediate results (each as an `OptimizerResult` instance) showcasing the progression of the optimization.
- Enhanced optimizer functionality via an updated `SciPyOptimizer`, which now supports a `store_intermediate_results` flag to record intermediate steps as part of its return structure.

Below is an example demonstrating how to use the new VQE implementation:

```python
import numpy as np
from qilisdk.common import algorithm
from qilisdk.common.optimizer import SciPyOptimizer
from qilisdk.digital.vqe import VQE
from qilisdk.digital.digital_result import DigitalResult
from qilisdk.extras.cuda_backend.cuda_backend import CudaBackend

# Define the problem parameters.
n_items = 4  # Number of items in the optimization problem.
max_weight_perc = 0.6  # Maximum allowed weight percentage.
# Randomly generate weights for each item.
weights = [np.random.randint(1, 5) for _ in range(n_items)]
# Randomly generate values for each item.
values = [np.random.randint(1, 10) for _ in range(n_items)]
# Compute the maximum allowed weight as a fraction of the total weight.
max_w = int(max_weight_perc * sum(weights))

# Initialize a hardware-efficient ansatz.
# 'n_qubits' is set to the number of items for this example.
nqubits = n_items
    n_qubits=nqubits,
    connectivity="Full",
    layers=1,
    one_qubit_gate="U3",
    two_qubit_gate="CNOT"
)

# Set up the backend, for instance using a CUDA backend for hardware acceleration.
backend = CudaBackend()

# LM is defined as the total sum of values and used within the cost function.
LM = sum(values)

def cost_function(x: DigitalResult) -> float:
    """
    Compute the cost based on the most probable measurement outcomes.
    Each outcome is evaluated by:
      - Calculating a penalty (H_a) if the total weight exceeds the maximum allowed weight.
      - Calculating a reward (H_b) for the selected items, given as a negative cost.
    The final cost is computed as the weighted sum (by the probability of each outcome) of these contributions.
    """
    # Get the most probable measurement outcomes.
    most_probable = x.get_probabilities()

    final_cost = 0
    for n in range(len(most_probable)):
        # Convert the measurement bitstring into a list of integers.
        x_n = [int(i) for i in list(most_probable[n][0])]
        # Calculate the total weight for this outcome.
        t = sum(weights[i] * x_n[i] for i in range(n_items))
        # Penalty if the total weight exceeds the allowed maximum.
        H_a = (t - max_w) if t > max_w else 0
        # Reward (negative cost) based on the values for selected items.
        H_b = -sum(values[i] * x_n[i] for i in range(n_items))
        # Final cost is accumulated using the probability of the outcome.
        final_cost = three_most_probable[n][1] * (LM * H_a + H_b)

    return final_cost

# Set up the SciPy optimizer using the "Powell" method and define parameter bounds.
optimizer = SciPyOptimizer(
    method="Powell",
    bounds=[(0, np.pi) for _ in range(ansatz.nparameters)]
)

# Create the VQE algorithm instance with an initial guess for the parameters.
algorithm = VQE(
    ansatz,
    [0.5 for _ in range(ansatz.nparameters)],
    cost_function
)

# Execute the VQE algorithm using the specified backend and optimizer.
# The flag store_intermediate_results is set to True to capture each optimization step.
results = algorithm.execute(backend, optimizer, store_intermediate_results=True)

# Display the optimization results.
print("Optimal Cost:", results.optimal_cost)
print("Optimal Parameters:", results.optimal_parameters)
print("Intermediate Optimization Results:")
for intermediate in results.intermediate_results:
    print("Cost:", intermediate.cost)
```


src/qilisdk/digital/digital_result.py
```diff
@@ -95,6 +97,20 @@ def get_probability(self, bitstring: str) -> float:
             float: The probability of the specified bitstring occurring.
         """
 
+    def most_probable(self, n: int | None = None) -> list[tuple[str, float]]:
```
amirazzam2000: I would suggest merging this method with `get_probabilities`

amirazzam2000 (commented):
I like this! good job 🙌

src/qilisdk/digital/digital_result.py
fedonman: I have removed the `get_probabilities` in the QaaSBackend PR and made it a property `probabilities`

amirazzam2000 (approved):

PR #27, opened 2025-04-14
[QHC-961] Implement models and methods for QaaSBackend's Execute requests

fedonman:
Improved `QaaSBacked` functionality to include methods for executing digital and analog algorithms.

### DIFF
--- a/src/qilisdk/yaml.py
+++ b/src/qilisdk/yaml.py
@@ -65,10 +65,7 @@ def lambda_constructor(constructor, node):
 
 def pydantic_model_representer(representer, data):
     """Representer for Pydantic Models."""
-    value = {
-        "type": f"{data.__class__.__module__}.{data.__class__.__name__}",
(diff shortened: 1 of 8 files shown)


src/qilisdk/yaml.py
```diff
@@ -19,6 +19,7 @@
 
 import numpy as np
 from dill import dumps, loads
+from pydantic import BaseModel
```
GuillermoAbadLopez, line 22: @fedonman this makes `pydantic` pass to be a mandatory requirement, and not only for `qaas` as before, right?
Screenshot 2025-05-06 at 12 34 41

amirazzam2000: ohhh yeah that's true. We can move it to a main dependency I think in this case it should be fine.

GuillermoAbadLopez: Cool, I'm doing a PR moving it 👍

PR #29, opened 2025-05-06
[QHC-963] Making `pydantic` a mandatory requirement

GuillermoAbadLopez:
Made `pydantic` pass to be a mandatory requirement, and not only for qaas as before. Solving a problem with installation overseen in previous PRs.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,6 +26,8 @@ dependencies = [
     "numpy>=2.2.4",
     "ruamel-yaml>=0.18.10",
     "scipy>=1.15.1",
+    "pydantic>=2.10.6",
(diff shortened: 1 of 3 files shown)


amirazzam2000 (approved):

GuillermoAbadLopez:
## PR following the discussion on:
https://github.com/qilimanjaro-tech/qilisdk/pull/27/files#r2075188697

## DOES:
Where a change made `pydantic` pass to be a mandatory requirement, and not only for `qaas` as before. So using only `uv sync` gave errors when importing the following:


With the change sin this PR now `uv sync`, installs it correctly👍

PR #31, opened 2025-05-06
fixing cudaq version

amirazzam2000:
- Updated ``Cudaq`` to version 0.10.0 to fix issues encountered in version 0.9.1
- Migrated ``CudaBackend`` to use the new version of ``Cudaq``


.github/workflows/code_quality.yml
```diff
@@ -43,6 +43,9 @@ jobs:
       - name: Install the project
         run: uv sync --all-groups --all-extras
 
+      - name: Explicit installation of cudaq
```
GuillermoAbadLopez, line 46: I guess we will need to do this manually also, when we wanna install the library locally now, no?

src/qilisdk/extras/cuda/cuda_backend.py
```diff
@@ -277,7 +277,7 @@ def get_schedule(key: str) -> Callable:
 
         return CudaAnalogResult(
             final_expected_values=np.array(
-                [exp_val.expectation() for exp_val in evolution_result.final_expectation_values()]
+                [exp_val.expectation() for exp_val in evolution_result.final_expectation_values()[0]]
```
GuillermoAbadLopez, line 280: Check this is intended

GuillermoAbadLopez (approved):
Once you remove the duplicated `cuda` dependency in `uv.lock` (appear both as mandatory and extra), the PR seems good to go 👍 

The changes in the file `uv.lock` are huge, and I couldn't check it all ofcourse, I leave it up to your criterion if you think we should merge that

.github/workflows/code_quality.yml
amirazzam2000: yes but I am looking for alternatives in the meanwhile.

src/qilisdk/extras/cuda/cuda_backend.py
amirazzam2000: yes, they changed the output format in version 0.10.0

PR #36, opened 2025-05-07
Release/0.1.3

amirazzam2000:


README.md
```diff
@@ -193,7 +193,7 @@ For analog simulations, the new `TimeEvolution` and `Schedule` classes allow you
 
 ```python
 import numpy as np
-from qilisdk.analog import TimeEvolution, Schedule, tensor, ket, X, Z, Y
+from qilisdk.analog import Schedule, TimeEvolution, ket, X, Z, Y, tensor_prod
```
GuillermoAbadLopez, line 196: I think we can delete this, since its not used later, no?
```suggestion
from qilisdk.analog import Schedule, TimeEvolution, ket, X, Z, Y
```

GuillermoAbadLopez (approved):
Approved with one small comment

README.md
amirazzam2000: it is used but I forgot to change the line it was used on should be fixed now

PR #37, opened 2025-05-07
[QHC-972] Improving `ptrace` function

GuillermoAbadLopez:
## PENDING TODOs: 
- [x] Contrast code vs Deepseek, Copilot and ChatGPT-mini-high ✅
- [x] Performance comparison against old implementation (by @amirazzam2000), more info on: https://github.com/qilimanjaro-tech/qilisdk/pull/37#issuecomment-2886670640

## Problems the original code had, which now work:

1. Running out of einsum labels
string.ascii_letters gives you 52 distinct single-character labels. But you need 2·n labels for an n-partite system. If n > 26, you’ll exhaust the iterator (or violate numpy’s one-char-per-axis requirement) and get a cryptic error.

2. No validation of keep
You should check that every index in keep lies in [0, len(dims)) and that there aren’t duplicates. Otherwise you could index out of range or double-count.

3. Minor type/coercion nit
np.prod(dims) can return a numpy scalar. It’s safer to do
`total_dim = int(np.prod(dims))`
so that your shape‐check and reshape calls can’t trip over dtype mismatches.

4. Didn't work for Bra's and Ket's

5. (GPT msg): Mismatched ordering when rebuilding the output shape. You collect the “kept” letters in ascending subsystem-index order, but then do
`dims_keep = [dims[i] for i in keep]`
If the user passes keep = [2,0], the einsum output indices will be in the order (0,2) but you’ll build dims_keep = [dims[2], dims[0]]. That will mis‐reshape your tensor.

### DIFF
--- a/src/qilisdk/analog/quantum_objects.py
+++ b/src/qilisdk/analog/quantum_objects.py
@@ -158,7 +158,7 @@ class QuantumObject:
         Raises:
             ValueError: If the product of the dimensions in dims does not match the
                 shape of the QuantumObject's dense representation or if any dimension is non-positive.
-            ValueError: If the indices in 'keep' are not unique, or are out of range, or duplicates.
+            ValueError: If the indices in 'keep' are not unique or are out of range.
 
         Returns:
             QuantumObject: A new QuantumObject representing the reduced density matrix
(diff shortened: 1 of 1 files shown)


GuillermoAbadLopez:
## SUMMARY STATUS PR:
We need to compare the performance of these 2 branches:
- `checking_performance` (old)
- `improving_ptrace` (new)

#### Cons of old:
- Has "Running out of einsum labels" problem, which can be solved with a not huge tweak of course
- Less readable (but can also be solved, by defining a intermediate private function _tracing() ...)

And we are missing to see where the best performance is, and depending on that, we chose one or the other. And if the older is chosen, then we apply both changes from above :+1:

GuillermoAbadLopez:
After a performance comparison, since the old implementation seems faster, this PR might get closed, by a new one: https://github.com/qilimanjaro-tech/qilisdk/pull/38, which keeps the old `einsum` implementation, but adds all the checks and readabilty improvements

GuillermoAbadLopez:
Closed in favor of: https://github.com/qilimanjaro-tech/qilisdk/pull/38

PR #38, opened 2025-05-16
[QHC-972] Improving `ptrace` function V2

GuillermoAbadLopez:
Follows from the changes suggested in the discussion/review of this other PR: https://github.com/qilimanjaro-tech/qilisdk/pull/37

### DIFF
--- a/src/qilisdk/analog/quantum_objects.py
+++ b/src/qilisdk/analog/quantum_objects.py
@@ -209,6 +209,9 @@ class QuantumObject:
 
         Returns:
             tuple[np.ndarray]: The resulting tensor after tracing out the specified subsystems.
+
+        Raises:
+            ValueError: If the number of subsystems exceeds the available ASCII letters.
         """
         # Check that the number of subsystems is not too large, that we run out of ascii letters.
         needed, MAX_LABELS = len(dims) + len(keep), len(string.ascii_letters)
--- a/src/qilisdk/analog/quantum_objects.py
+++ b/src/qilisdk/analog/quantum_objects.py
@@ -187,7 +187,6 @@ class QuantumObject:
             raise ValueError("duplicate indices in keep")
 
         # 3) Trace out the subsystems not in `keep`.
-        # Check if the input is a valid for a partial trace
         rho_t = self._tracing_indices(rho, dims, keep)
 
         # 4) The resulting tensor has separate indices for each subsystem kept.
--- a/src/qilisdk/analog/quantum_objects.py
+++ b/src/qilisdk/analog/quantum_objects.py
@@ -198,7 +198,7 @@ class QuantumObject:
             raise ValueError("duplicate indices in keep")
 
         # 4) Trace out the subsystems not in `keep`.
-        rho_t = self._compute_traced_tensor_via_einstein_summation(rho, keep, dims)
+        rho_t = self._compute_traced_tensor_via_einstein_summation(rho, keep_set, dims)
 
         # 5) The resulting tensor has separate indices for each subsystem kept.
         # Reshape it into a matrix (i.e. combine the row indices and column indices).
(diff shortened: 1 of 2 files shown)


src/qilisdk/analog/quantum_objects.py
```diff
@@ -150,22 +150,75 @@ def ptrace(self, dims: list[int], keep: list[int]) -> "QuantumObject":
...
-        total_dim = np.prod(dims)
+        # 1) Get the density matrix representation:
+        rho = self.to_density_matrix().dense if self.is_ket() or self.is_bra() else self.dense
+
+        # 2) Basic checks for dims
+        total_dim = int(np.prod(dims))
         if rho.shape != (total_dim, total_dim):
-            raise ValueError("Dimension mismatch between provided dims and QuantumObject shape")
+            raise ValueError(
+                f"Dimension mismatch: QuantumObject shape {rho.shape} does not match the expected shape ({total_dim}, {total_dim}) from prod(dims)."
```
amirazzam2000: > from prod(dims)

I would remove this cuz for the user there is no context to what that is.

src/qilisdk/analog/quantum_objects.py
```diff
@@ -150,22 +150,75 @@ def ptrace(self, dims: list[int], keep: list[int]) -> "QuantumObject":
...
+            ValueError: If the indices in 'keep' are not unique or are out of range.
 
         Returns:
             QuantumObject: A new QuantumObject representing the reduced density matrix
                 for the subsystems specified in 'keep'.
         """
-        rho = self.dense
-        total_dim = np.prod(dims)
+        # 1) Get the density matrix representation:
+        rho = self.to_density_matrix().dense if self.is_ket() or self.is_bra() else self.dense
```
amirazzam2000: the method `to_density_matrix` should return self if self is a density matrix, so no need to add the check here.

src/qilisdk/analog/quantum_objects.py
```diff
@@ -150,22 +150,75 @@ def ptrace(self, dims: list[int], keep: list[int]) -> "QuantumObject":
         The input 'dims' represents the dimensions of each subsystem,
         and 'keep' indicates the indices of the subsystems to be retained.
 
+        If the QuantumObject is a ket or bra, it will first be converted to a density matrix.
+
         Args:
             dims (list[int]): A list specifying the dimensions of each subsystem.
```
amirazzam2000: I was thinking if possible to allow this to be None and build it automatically assuming it's a system of N qubits [dim=2] (which you can extract from the data of self)

src/qilisdk/analog/quantum_objects.py
GuillermoAbadLopez: I think it's actually pretty useful. Without it I think its hard for the user to correct their code to match the dimensions.

But lemme try and write it with more context, so the user can understand it better

GuillermoAbadLopez: @amirazzam2000 read it again, and see if its better now 👍

src/qilisdk/analog/quantum_objects.py
GuillermoAbadLopez: Just implemented it, check if you like it :) 🙌

src/qilisdk/analog/quantum_objects.py
```diff
@@ -141,7 +142,7 @@ def adjoint(self) -> QuantumObject:
         out = QuantumObject(self._data.conj().T)
         return out
 
-    def ptrace(self, dims: list[int], keep: list[int]) -> "QuantumObject":
+    def ptrace(self, keep: list[int], dims: list[int] | None = None) -> "QuantumObject":
```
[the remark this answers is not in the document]
GuillermoAbadLopez: Not needed for this stage of devs

src/qilisdk/analog/quantum_objects.py
GuillermoAbadLopez: In the end we kind of needed this, but just for the case, where they pass a square operator, but which is not a density matrix, so I updated to just that check, and left the rest inside the function as you said 👌

amirazzam2000 (approved):

PR #39, opened 2025-05-20
[QHC-923] Add digital configuration (routing & optimization), to qaas

GuillermoAbadLopez:
For the routing/optimizations, etc.., in the future we would need something like this PR 🤔


GuillermoAbadLopez:
Outdated PR

PR #40, opened 2025-06-17
Removing manual installation of CUDAQ from Github Actions

amirazzam2000:
removing
```
- name: Explicit installation of cudaq
        run: uv pip install cudaq==0.10.0
```

from `code_quality.yml` and `test.yml`.


fedonman (approved):
Nice catch!

fedonman:
Add a small changelog entry though please :)

PR #42, opened 2025-06-26
[ALGP-275] Use the Abstract Model as the cost function in the Optimizer module.

amirazzam2000:
Replacing the callable cost function with the abstract model. The Abstract Model allows the user to write an objective function that is subjected to a set of constraints. Then this is used to evaluate the cost of a given sample. 

Example: 
```python
from qilisdk.common import SciPyOptimizer
from qilisdk.common.model import QUBO, Model
from qilisdk.common.variables import LEQ, BinaryVariable
from qilisdk.digital.digital_result import DigitalResult
from qilisdk.digital.vqe import VQE
from qilisdk.extras import CudaBackend


## Define the model

model = Model("Knapsack")

values = [2, 3, 7]
weights = [1, 3, 3]
max_weight = 4
b = [BinaryVariable(f"b{i}") for i in range(len(values))]

obj = sum(b[i] * values[i] for i in range(len(values)))
model.set_objective(obj, "obj")

con = LEQ(sum(b[i] * weights[i] for i in range(len(weights))), max_weight)

model.add_constraint("max_weight", con)

## Define the Ansatz: 
n_qubits = 3
    n_qubits=n_qubits, layers=2, connectivity="Linear", structure="grouped", one_qubit_gate="U2", two_qubit_gate="CNOT"
)

## Build the VQE object
vqe = VQE(
    ansatz=ansatz,
    initial_params=[0 for _ in range(ansatz.nparameters)],
    model=model,
)

## Define the Backend and the Optimizer
backend = CudaBackend()
optimizer = SciPyOptimizer(method="Powell")

## Execute the VQE to find the optimal parameters
result = vqe.execute(backend, optimizer, nshots=1000)

## Sample the circuit using the optimal parameters
circuit = ansatz.get_circuit(result.optimal_parameters)
results = backend.execute(circuit)

## Print the probabilities
print(results.get_probabilities())
```


src/qilisdk/common/variables.py
```diff
@@ -870,6 +870,8 @@ def num_binary_equivalent(self) -> int:  # noqa: PLR6301
 
     def evaluate(self, value: list[int] | Number) -> float:
         if isinstance(value, Number):
+            if value in {1.0, 0.0, 1, 0}:
```
fedonman: either `{0, 1}` or `{0.0, 1.0}` will work the same.

src/qilisdk/common/model.py
```diff
@@ -803,6 +824,19 @@ def _build_qubo_objective(
                 sense=ObjectiveSense.MINIMIZE,
             )
 
+    def evaluate(
```
fedonman: write a docstring for this method since it's public. `sample` and results `dict` will need clear explanations.

src/qilisdk/common/model.py
```diff
@@ -803,6 +824,19 @@ def _build_qubo_objective(
...
+        self, sample: Mapping[BaseVariable, Number | list[int]]
+    ) -> dict[str, float]:  # TODO (ameer): replace this with a model result object?
```
fedonman: If we explain what `dict[str, float]` is it's ok to leave it as it is.

If the dictionary is complicated to understand and/or we need heavy post-processing of the results, then we might need to switch to a results class.

fedonman: For now, I would leave it a `dict` and iterate in another sprint.

fedonman (changes requested):
Left some minor comments to address. Please add a changelog entry too.

fedonman (approved):
:100: :100: :100:

PR #44, opened 2025-07-01
[ALGP-286] Improve coverage of Hamiltonian, Variables, and Model classes.

amirazzam2000:
Increasing the coverage of the unit tests for the Hamiltonian, Model, and Variables classes. Now the coverage for each of them is: 
- Hamiltonian class (98.59%)
- Variable class (96.85%)
-  Model class (95.56%)


fedonman (approved):
Awesome work! 💯 🚀 

rosycheeks-approved

PR #45, opened 2025-07-01
[ALGP-287] Fix Hamiltonian.parse() method

fedonman:
The `parse` method has been made far more forgiving of extra or missing spaces around coefficients and operators. Internally, all whitespace inside any parentheses—whether around a complex coefficient like `(2.5 + 3j)` or an operator index like `Y( 0 )`—is now stripped out, and runs of spaces elsewhere are collapsed to single spaces. A post‐processing regex also ensures there’s always exactly one space between a closing parenthesis and the next Pauli operator token. As a result, variants such as `(2.5 + 3j)Y(0)`, `(2.5+3j)   Y(0)`, and `(2.5 + 3j) Y(0)` will all parse correctly without altering any downstream logic.


amirazzam2000 (approved):
amazing thanks!

PR #48, opened 2025-07-16
[QHC-1044] Implement QaaSBackend Methods for Interacting with Jobs

fedonman:
- You can now pass a custom callable to both `list_devices` and `list_jobs`, enabling flexible client‑side filtering without touching the server API. Two new convenience helpers extend the job‑handling workflow: `get_job_details` fetches a complete record (payload, results, logs, and error information) for a given job ID, while `wait_for_job` blocks the caller until that job reaches a terminal state—`COMPLETED`, `ERROR`, or `CANCELLED`. In addition, the execution helpers (`execute`, `evolve`, `run_vqe`, and `run_time_evolution`) have been streamlined to return only the integer job identifier, making immediate responses lighter and encouraging explicit follow‑up calls when you actually need detailed data.

- Fixed versions of cuQuantum libraries to avoid incompatibility issues.


src/qilisdk/extras/qaas/models.py
amirazzam2000, line 1: I would still change the name of this file to at least `qaas_models` just to make sure no confusion happens

amirazzam2000 (approved):

PR #50, opened 2025-07-21
[ALGP-289] Refactor to Functionals, unified Backend and QaaS

fedonman:
Introducing **Functionals**, a unified abstraction for quantum work units. The library now includes two functional families: **`Sampling`**, which executes gate‑based circuits and returns shot counts, and **`TimeEvolution`**, which drives Hamiltonian schedules and yields evolved states or expectation values. All back‑ends have been consolidated into the `qilisdk.backends` module and are now Functional‑aware—each advertises the families it supports and executes them via `backend.execute(functional)`. Meanwhile, **QaaS** has been refactored into a standalone API client located in the `qilisdk.qaas` module, providing endpoints for submitting Functionals to remote devices, monitoring job status, and listing available jobs and hardware.


src/qilisdk/functionals/sampling.py
```diff
@@ -12,9 +12,16 @@
...
-from qilisdk.analog.analog_result import AnalogResult
+from qilisdk.digital.circuit import Circuit
+from qilisdk.functionals.functional import Functional
+from qilisdk.functionals.sampling_result import SamplingResult
 from qilisdk.yaml import yaml
 
 
 @yaml.register_class
-class QaaSTimeEvolutionResult(AnalogResult): ...
+class Sampling(Functional[SamplingResult]):
```
amirazzam2000, line 22: why are we calling it Sampling, maybe Circuit_Execution? because we still sample the state at the end of the time evolution. Like all of our results are samples.

src/qilisdk/qaas/qaas.py
```diff
@@ -298,75 +303,51 @@ def _ensure_device_selected(self) -> int:
             raise ValueError("Device not selected.")
         return self._selected_device
 
-    def execute(self, circuit: Circuit, nshots: int = 1000) -> int:  # type: ignore[override]
-        """Submit a digital circuit to the selected device.
+    def submit(self, functional: Functional) -> int:
```
amirazzam2000, line 306: If we’re keeping it the same as the backends in terms of how it’s used, then I’d keep the name of this method as execute. Otherwise, if we want to make it different (like adding extra parameters such as device type or something) then maybe a new name makes sense. But as it is right now, it just feels like an inconvenience for the user, as they would have to change how it’s called without getting any real benefit out of it. Does that make sense?

src/qilisdk/functionals/sampling.py
fedonman: Functionals represent the fundamentals actions performed to the quantum structures to get classical outputs. These actions are executed by the backend. So `backend.execute(Sampling(circuit, 1000))` can be read as "**Execute** a Sampling of **circuit** over 1000 **nshots**" and `backend.execute(TimeEvolution(schedule, initial_state, observables))` can be read as "**Execute** a **TimeEvolution** of **schedule** given an **initial state** and get the expectation values of the **observables**".

`backend.execute(CircuitExecution(circuit, 1000))` is a bit weird. It reads "**Execute** a **circuit execution** of **circuit** over 1000 **nshots**".

I prefer the names of the actual actions as they are. `Sampling` and `TimeEvolution`. Even `TimeEvolution` could be shortened to just `Evolution` (since evolution implies there is a time progress anyway)

src/qilisdk/qaas/qaas.py
fedonman: `QaaS` now is not a `Backend`. So the process changes as well.

In backends, users **execute** a functional and get back the results.

In QaaS, users **submit** a functional to the QaaS and get back the submitted job id. Then they have to poll the job status, and when the job is finished, retrieve the results.

src/qilisdk/functionals/sampling.py
amirazzam2000: okay yes makes sense, but I would also like to point out that when we were doing the demo at the ACM school the most asked for feature was to be able to get the state vector at the end of the execution. So maybe we can call it digital evolution and analog evolution?

src/qilisdk/qaas/qaas.py
amirazzam2000: yeah okay that's fair I am convinced!

amirazzam2000 (approved):
I think we can approve this for now and if we decide in the future to do a name change we can do it in a different PR 👍 thanks a lot Vyron!!

src/qilisdk/functionals/sampling.py
fedonman: I was also thinking about getting the state. But there are two things to consider:
1) We cannot get the state of the circuit through hardware execution. It is a simulation-only thing.
2) Even in simulators, you need to execute the circuit again to get the state. For example, in cudaq the `get_state()` method is different than the `sample()` method. You also need to remove the measurement gates to get the quantum state and not the classical one after measurement. https://nvidia.github.io/cuda-quantum/latest/using/examples/executing_kernels.html#get-state If we add it in our current Functional, we will do additional unnecessary extra executions since users don't always want the state but rather the measurement samples.

Taking that into mind, if we want to provide users the functionality of getting the state, I would make it a separate Functional called `State()` or `StateEvolution()` and let them explicitly ask for it.

amirazzam2000: or I would just add it as a flag in kwargs

PR #52, opened 2025-07-23
[Algp-295] Bug encountered when executing time evolution with Qutip backend with more than one qubit

amirazzam2000:


amirazzam2000:
branched from the wrong branch

PR #53, opened 2025-07-23
[ALGP-295] Bug encountered when executing time evolution with Qutip backend with more than one qubit

amirazzam2000:
Qutip backend dimension mismatch bug fix. This happened when executing time evolution with more than one qubit as the dimensions of the state and the observables did not match. for instance, for three qubits, the state had the dimension [1, 8] while the observables had [[2, 2, 2],[2, ,2, 2]]. the bug was solved by explicitly specifying the dimensions of all the Qobj objects when constructed.


fedonman (changes requested):
Since it is a bug, write a test case covering the bug, so it is not introduced again.

fedonman (approved):

PR #62, opened 2025-08-11
[ALGP-301] Implement VariationalProgram

amirazzam2000:
Updating the structure of how variational algorithms are executed. We added a new `VariationalProgram` class that takes in a parameterized Functional, an Optimizer, and a model representing the cost function, then uses these elements to optimize the parameters of the functional. 

Moreover, due to these changes the VQE class was removed and now all variational programs will be constructed and executed using VariationalProgram. Here is an example of optimizing a VQE to solve the knapsack problem: 

```python
import numpy as np

from qilisdk.backends import QutipBackend
from qilisdk.common import BinaryVariable, LessThanOrEqual, Model, ObjectiveSense
from qilisdk.digital import CNOT, U2, HardwareEfficientAnsatz
from qilisdk.functionals import Sampling, VariationalProgram
from qilisdk.optimizers import SciPyOptimizer

values = [2, 3, 7]
weights = [1, 3, 3]
max_weight = 4
binary_var = [BinaryVariable(f"b{i}") for i in range(len(values))]

model = Model("Knapsack")

model.set_objective(sum(binary_var[i] * values[i] for i in range(len(values))), sense=ObjectiveSense.MAXIMIZE)

model.add_constraint("max_weights", LessThanOrEqual(sum(binary_var[i] * weights[i] for i in range(len(weights))), max_weight))


n_qubits = 3
    n_qubits=n_qubits, layers=3, connectivity="Linear", structure="grouped", one_qubit_gate=U2, two_qubit_gate=CNOT
)
circuit = ansatz.get_circuit([np.random.uniform(0, np.pi) for _ in range(ansatz.nparameters)])

optimizer = SciPyOptimizer(method="Powell")

backend = QutipBackend()
result = backend.execute(VariationalProgram(functional=Sampling(circuit), optimizer=optimizer, cost_model=model))

print(result)
```


src/qilisdk/functionals/functional.py
```diff
@@ -13,13 +13,62 @@
...
+    def compute_cost(self, results: TResult, cost_model: Model) -> float:
+        """Compute the cost of the functional given a cost model.
+
+        Args:
+            results (Result): The functional results
+            cost_model (Model): The Model object used to represent the cost of different states.
+
+        Returns:
+            float: the cost of the results.
+        """
```
fedonman: We could add this method to the `Result` class instead as `compute_cost(self, cost_model: Model)`. There is no reason to belong in `Functional`.

src/qilisdk/functionals/sampling.py
```diff
@@ -25,3 +27,26 @@ class Sampling(Functional[SamplingResult]):
...
+    def get_parameters(self) -> dict[str, Number]:
+        return self.circuit.get_parameters()
+
+    def get_parameter_names(self) -> list[str]:
+        return list(self.circuit.get_parameters().keys())
+
+    def get_parameter_values(self) -> list[Number]:
+        return list(self.circuit.get_parameters().values())
+
+    def compute_cost(self, results: SamplingResult, cost_model: Model) -> float:  # noqa: PLR6301
```
fedonman: The `# noqa: PLR6301` is also an indicator for this, we don't actually use anything from the Functional.

src/qilisdk/functionals/parameterized_program.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+from qilisdk.common.model import Model
+from qilisdk.functionals.functional import Functional
+from qilisdk.optimizers.optimizer import Optimizer
+from qilisdk.yaml import yaml
+
+
+@yaml.register_class
+class ParameterizedProgram:
```
fedonman: Since `ParameterizedProgram` is not a Functional, maybe it belongs to the `Common` module? Or even better a new one?

fedonman: Alternatively, we can explore if it makes sense to implement it as `Functional` or a generic over `Functional`. I'll think about it too.

src/qilisdk/functionals/functional.py
```diff
@@ -13,13 +13,62 @@
...
+    @abstractmethod
+    def get_parameters(self) -> dict[str, Number]:
+        """Gets the values of the parameters of the functional.
+
+        Returns:
+            dict[str, Number]: a dictionary with the parameter label and its current value.
+        """
+
+    @abstractmethod
+    def get_parameter_names(self) -> List[str]:
```
fedonman: Don't use `List` or `Dict` annotations from `typing` module`. The native `list` and `dict` are preferred.

src/qilisdk/backends/backend.py
```diff
@@ -59,3 +61,39 @@ def _execute_sampling(self, functional: Sampling) -> SamplingResult:
 
     def _execute_time_evolution(self, functional: TimeEvolution) -> TimeEvolutionResult:
         raise NotImplementedError(f"{type(self).__qualname__} has no TimeEvolution implementation")
+
+    def optimize(
+        self, parameterized_program: ParameterizedProgram, store_intermediate_results: bool = False
```
fedonman: Shouldn't `store_intermediate_results` be an attribute of `ParameterizedProgram`? In the sense, that it is also an attribute of `TimeEvolution` and we do `execute(time_evo)` directly and not `execute(time_evo, store_intermediate_results=True)`.

src/qilisdk/speqtrum/speqtrum.py
```diff
@@ -385,13 +384,15 @@ def _submit_time_evolution(self, time_evolution: TimeEvolution) -> int:
         logger.info("Time evolution job submitted: {}", job.id)
         return job.id
 
-    def submit_vqe(
-        self, vqe: VQE, optimizer: Optimizer, nshots: int = 1000, store_intermediate_results: bool = False
+    def submit_parameterized_program(
```
fedonman: Why not include this in the same `submit` method?

fedonman (commented):
Super nice work! Made some comments.

src/qilisdk/speqtrum/speqtrum.py
amirazzam2000: because ParameterizedProgram is not a Functional

fedonman:
After some thought, I strongly believe that ParameterizedProgram should be a Functional. 

Also since Functionals are the representation of the "action" a renaming to `Optimization` would be ideal. Or just `Parameterization`.

Behind the scenes `Optimization` is a Functional and also a generic over a Functional (that is also a generic over the result type) and so when the user creates an `Optimization` the results type is automatically inferred. Also, there is no need for seperate methods, everything is handled the same way.

I tried the change in my local, and works perfectly.

Draft:
```

R = TypeVar("R", bound=Result)

@yaml.register_class
class ParameterizedProgram(Functional[ParameterizedProgramResults[R]], Generic[R]):
    def __init__(self, functional: Functional[R], optimizer: Optimizer, cost_model: Model) -> None:
        self._functional = functional
        self._optimizer = optimizer
        self._cost_model = cost_model

    # ---- Functional[...] interface (delegated) ----
    @property
    def result_type(self) -> Type[ParameterizedProgramResults[R]]:
        # Generics are erased at runtime; cast is for static type checkers
        return cast(Type[ParameterizedProgramResults[R]], ParameterizedProgramResults)

    def set_parameters(self, parameters: dict[str, Number]) -> None:
        self._functional.set_parameters(parameters)

    def get_parameters(self) -> dict[str, Number]:
        return self._functional.get_parameters()

    def get_parameter_names(self) -> list[str]:
        return self._functional.get_parameter_names()

    def get_parameter_values(self) -> list[Number]:
        return self._functional.get_parameter_values()

    # ---- Accessors (unchanged) ----
    @property
    def functional(self) -> Functional[R]:
        return self._functional

    @property
    def optimizer(self) -> Optimizer:
        return self._optimizer

    @property
    def cost_model(self) -> Model:
        return self._cost_model
```

fedonman:
And if we don't like the get/set parameters in the Optimization, we can make them methods of `ParameterizedFunctional` (which TimeEvolution and Sampling) inherit, while Optimization inherits from the base Functionll. 

Functional -> ParameterizedFunctional -> Sampling, TimeEvo
Functional -> Optimization

Then it is also very intuitive: Optimization is an optimization of a parameterized functional given an optimizer and a cost model.

amirazzam2000:
> And if we don't like the get/set parameters in the Optimization, we can make them methods of `BasicFunctional` (which TimeEvolution and Sampling) inherit, while Optimization inherits from the base Functional.
> 
> Functional -> BasicFunctional -> Sampling, TimeEvo Functional -> Optimization

okay regarding this, I don't fully agree because if we make Parameterized Program a functional it might lead to some people passing parameterized program to a parameterized program
I think the better course of action is to remove parameterized program and move it to its own name space, something like:
variatoinals:
 |_ parameterized_program (or we can rename it to variational_program)
Funciontals: # this is the primitive executions and I don't think that variationals are primitive.
 |_ ...
plus I don't think the parameterized program should get the same treatment as the functionals in the backend.

fedonman:
Look at the second comment, this is solved

> > And if we don't like the get/set parameters in the Optimization, we can make them methods of `BasicFunctional` (which TimeEvolution and Sampling) inherit, while Optimization inherits from the base Functional.
> > 
> > Functional -> BasicFunctional -> Sampling, TimeEvo Functional -> Optimization
> 
> okay regarding this, I don't fully agree because if we make Parameterized Program a functional it might lead to some people passing parameterized program to a parameterized program
> I think the better course of action is to remove parameterized program and move it to its own name space, something like:
> variatoinals:
>  |_ parameterized_program (or we can rename it to variational_program)
> Funciontals: # this is the primitive executions and I don't think that variationals are primitive.
>  |_ ...
> plus I don't think the parameterized program should get the same treatment as the functionals in the backend.

src/qilisdk/functionals/functional.py
```diff
@@ -13,13 +13,53 @@
...
+TGenericResult_co = TypeVar("TGenericResult_co", bound=Result, covariant=True)
 
-class Functional(Protocol[TResult_co]):
+
+class Functional(Protocol[TGenericResult_co]): ...
+
+
+class PrimitiveFunctional(Functional, Protocol[TResult_co]):
     @property
     def result_type(self) -> Type[TResult_co]: ...
```
fedonman: There are some tweaks needed here.
1. The property `result_type` need to be in parent `Functional`, since all functionals have some kind of results and we need to know about their type.
2. There is no need to have a TypeVar of the `Result` class, since we have created the new `FunctionalResult` class. Functionals return some kind of FunctionalResult.


```suggestion
TResult_co = TypeVar("TResult_co", bound=FunctionalResult, covariant=True)


class Functional(Protocol[TResult_co]):
    @property
    def result_type(self) -> Type[TResult_co]: ...


class PrimitiveFunctional(Functional[TResult_co]):
```

src/qilisdk/functionals/variational_program.py
```diff
@@ -0,0 +1,59 @@
...
+
+@yaml.register_class
+class VariationalProgram(Functional[VariationalProgramResults]):
+    def __init__(
+        self,
+        functional: PrimitiveFunctional,
+        optimizer: Optimizer,
+        cost_model: Model,
+        store_intermediate_results: bool = False,
+    ) -> None:
```
fedonman: If we leave it like this, then we don't know what kind of results the underlying functional returns. It's easy to make static typers aware, but is a bit tricky: We need to say that instead of a `PrimitiveFunctional` we have a `PrimitiveFunctiona[R]` (a functional that returns R results)


```suggestion

R = TypeVar("R", bound=FunctionalResult)


@yaml.register_class
class VariationalProgram(Functional[VariationalProgramResults[R]], Generic[R]):
    def __init__(
        self,
        functional: PrimitiveFunctional[R],
        optimizer: Optimizer,
        cost_model: Model,
        store_intermediate_results: bool = False,
    ) -> None:
```

src/qilisdk/functionals/variational_program.py
```diff
@@ -0,0 +1,59 @@
...
+                                is the cost function used by the optimizer.
+        """
+        self._functional = functional
+        self._optimizer = optimizer
+        self._cost_model = cost_model
+        self._store_intermediate_results = store_intermediate_results
+
+    @property
+    def functional(self) -> PrimitiveFunctional:
+        return self._functional
```
fedonman: We need to also use the `R` here, and also implement the property `result_type` of `Functional`.

```suggestion
    @property
    def result_type(self) -> Type[VariationalProgramResults[R]]:
        # Generics are erased at runtime; cast helps static analyzers.
        return cast("Type[VariationalProgramResults[R]]", VariationalProgramResults)

    @property
    def functional(self) -> PrimitiveFunctional[R]:
        return self._functional
```

src/qilisdk/common/result.py
```diff
@@ -11,7 +11,22 @@
...
+    @abstractmethod
+    def compute_cost(self, cost_model: Model) -> float:
+        """Compute the cost of the functional given a cost model.
+
+        Args:
+            cost_model (Model): The Model object used to represent the cost of different states.
+
+        Returns:
+            float: the cost of the results.
+        """
```
fedonman: This should be moved to `functionals` module in `functional_result.py` file.

fedonman (commented):
Super nice! I left some comments regarding generics. The idea is to be able for type checkers to understand that if we have a VariationalProgram of Sampling, the underlying execution_results of VariationalProgramResult is a SamplingResult.

Image

src/qilisdk/functionals/variational_program_results.py
```diff
@@ -0,0 +1,94 @@
...
+
+from qilisdk.common.result import FunctionalResult, Result
+from qilisdk.optimizers.optimizer_result import OptimizerIntermediateResult, OptimizerResult
+from qilisdk.yaml import yaml
+
+TResult_co = TypeVar("TResult_co", bound=FunctionalResult, covariant=True)
+
+
+@yaml.register_class
+class VariationalProgramResults(Result, Generic[TResult_co]):
```
fedonman: Since `SamplingResult` and `TimeEvolutionResult` is singular, let's rename this to `VariationalProgramResult` too. (and the file)

Also this should inherit from `FunctionalResult` too. (it's perfectly fine to inherit from class A and also be a generic over class A.

```suggestion
TResult_co = TypeVar("TResult_co", bound=FunctionalResult, covariant=True)


@yaml.register_class
class VariationalProgramResult(FunctionalResult, Generic[TResult_co]):
```

src/qilisdk/backends/backend.py
```diff
@@ -42,7 +45,7 @@ def execute(self, functional: Sampling) -> SamplingResult: ...
     def execute(self, functional: TimeEvolution) -> TimeEvolutionResult: ...
```
fedonman, line 46: We need an overload for the specific case of `VariationalProgram`.
```suggestion
    @overload
    def execute(self, functional: VariationalProgram[TResult]) -> VariationalProgramResults[TResult]: ...
```

fedonman (approved):

PR #63, opened 2025-08-11
[QHC-905] Implement Circuit Drawing using Matplotlib

fedonman:
Added a high-level `Circuit.draw(style: CircuitStyle = CircuitStyle(), filepath: str | None = None)` that renders the circuit via `MatplotlibCircuitRenderer.plot()` and optionally saves it (format inferred from the file extension). Crucially, passing the `style` parameter **bypasses all library defaults** for visuals—theme, colors, fonts (including the default PlusJakartaSans), DPI, spacing, and math/label formatting — without mutating global Matplotlib `rcParams`. This makes figures reproducible and sandboxed: the style you pass is the style you get, regardless of notebook or app settings. For example, `circuit.draw(style=CircuitStyle(theme=dark, fontsize=12, fontfname=None, fontfamily="DejaVu Sans"), filepath="circuit.svg")` forces a dark theme and DejaVu Sans font locally to this call. If `style` is omitted, the library’s default style is used.


src/qilisdk/utils/visualization/PlusJakartaSans-SemiBold.ttf
amirazzam2000, line 1: what is this file?

amirazzam2000 (commented):
I am only missing the addition of this to the docs. If you can add that in one of the examples or in the circuit section, that would be amazing. Just to mention what things can we change etc.

src/qilisdk/utils/visualization/PlusJakartaSans-SemiBold.ttf
fedonman: The font we are using.

amirazzam2000 (approved):
Perfect! thanks a lot 🙏

PR #65, opened 2025-08-20
[Hotfix] Make matplotlib a core dependency / Update dependencies / Fix typos

fedonman:
Made `matplotlib` a core dependency, as it is required for the `Circuit.draw()` method.  
Updated various dependencies to their latest versions.  
Fixed typos in SpeQtrum’s models.


amirazzam2000 (approved):
perfect! thanks

PR #66, opened 2025-08-25
[ALGP-307] Refactor Ansatz to inherit from Circuit

fedonman:
Ansatz now **inherits** from Circuit.

We refactored the hierarchy so that `Ansatz` is an abstract subclass of `Circuit` rather than a separate builder/factory. Concrete templates like `HardwareEfficientAnsatz` are therefore circuits themselves. This simplifies downstream code: anything that accepted a `Circuit` now accepts an ansatz instance directly, without an extra “build” or “get\_circuit” step. As part of this change, the circuit is constructed during initialization, configurable attributes were made private with read-only properties (`layers`, `connectivity`, `structure`, `one_qubit_gate`, `two_qubit_gate`), and parameter handling became deterministic: all single-qubit parameters default to `0.0`; The `structure` flag is now meaningful at build time: `grouped` schedules `U(all) → E(all)` per layer, while `interposed` applies `U(q) → E(all)` for each qubit within a layer. Measurement gates are no longer appended automatically; add them explicitly if needed. The `nparameters` computation remains available via the parent `Circuit`.

**Example**

```python
from qilisdk.digital import HardwareEfficientAnsatz, U3, CNOT

    nqubits=4,
    layers=2,
    connectivity="linear",        # or "circular" / "full"   
    structure="grouped",          # or "interposed"
    one_qubit_gate=U3,            # or U1 / U2
    two_qubit_gate=CNOT,          # or CZ
)
ansatz.draw()                     # ansatz is a Circuit

# add measurements explicitly if your workflow requires them
```


src/qilisdk/digital/ansatz.py
```diff
@@ -11,75 +11,85 @@
...
-        n_qubits: int,
+        nqubits: int,
         layers: int = 1,
-        connectivity: Literal["Circular", "Linear", "Full"] | list[tuple[int, int]] = "Linear",
-        structure: Literal["grouped", "interposed"] = "grouped",
+        connectivity: Connectivity = "linear",
+        structure: Structure = "grouped",
         one_qubit_gate: Type[U1 | U2 | U3] = U1,
         two_qubit_gate: Type[CZ | CNOT] = CZ,
+        initial_parameters: Sequence[float] | None = None,
```
amirazzam2000: should we make this a sequence of Parameter objects? because it feels inconsistent to make it floats.

amirazzam2000 (changes requested):

src/qilisdk/digital/ansatz.py
fedonman: This is the initial values of the parameters as it was in the old `get_circuit(parameters)` method.

fedonman: Maybe I should rename it to `initial_parameter_values`?

fedonman: Or we could remove it altogether and let users use one of the `set_parameters` methods.

amirazzam2000: yeah maybe actually that would be more clear. Actually I would also make it as well a mandatory field. 

On another note, it might be a bit tricky for the user to know how many parameters there are in the circuit. so we should have a way of providing this information. because before with the get_circuit you can know the number of parameters before you provide them.

fedonman: Definitely not mandatory field, since all parameters take a default value (0.0) and it is **very** bothersome to have to give different values.

For me the best solution is to delete the field, and let the users set the parameters the normal way with the `set_*` methods.

amirazzam2000: okay yes i agree, let's delete it all together and just mention it in the documentation that all parameters are initialized to 0

amirazzam2000 (approved):
Perfect, thanks a lot!

PR #68, opened 2025-08-26
[ALGP-309] Add Observable Cost Function

amirazzam2000:
Adding Observable cost function that now could be used with VariationalPrograms. 

You can construct an Observable cost function by using ``Hamiltonian``, ``PauliOperators``, or ``QuantumObjects``.

The cost is the expected value of the observable given the final state after the simulation. 

```python
from qilisdk.analog import Z, PauliZ
from qilisdk.common import QuantumObject
from qilisdk.cost_functions import ObservableCostFunction
import numpy as np


h = Z(0) + 2 * Z(1)
cost_function = ObservableCostFunction(h)

## or

cost_function = ObservableCostFunction(PauliZ(0))

## or

cost_function = ObservableCostFunction(QuantumObject(np.array([[1, 0], [0, -1]])))
```

Usage: 
```python
from qilisdk.analog import Z
from qilisdk.common import ket, tensor_prod
from qilisdk.cost_functions import ObservableCostFunction
from qilisdk.functionals import TimeEvolutionResult

n = 2

H = sum(Z(i) for i in range(n))

ocf = ObservableCostFunction(H)

te_results = TimeEvolutionResult(
    final_expected_values=np.array([[-0.9, 0]]),
    expected_values=None,
    final_state=tensor_prod([ket(1), ket(1)]),
    intermediate_states=None,
)
cost = ocf.compute_cost(te_results)
# Output: -2
```

### DIFF
--- a/changes/68.feature.md
+++ b/changes/68.feature.md
@@ -2,6 +2,9 @@ Adding Observable cost function that now could be used with VariationalPrograms.
 
 You can construct an Observable cost function by using ``Hamiltonian``, ``PauliOperators``, or ``QuantumObjects``.
 
+The cost is the expected value of the observable given the final state after the simulation. 
+
+
 ```python
 from qilisdk.analog import Z, PauliZ
 from qilisdk.common import QuantumObject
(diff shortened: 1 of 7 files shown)


changes/68.feature.md
```diff
@@ -0,0 +1,22 @@
...
+cost_function = ObservableCostFunction(h)
+
+## or
+
+cost_function = ObservableCostFunction(PauliZ(0))
+
+## or
+
+cost_function = ObservableCostFunction(QuantumObject(np.array([[1, 0], [0, -1]])))
+```
```
fedonman: Could we add more info to the changelog?
- How the cost is calculated based on a Hamiltonian/QuantumObject.
- Mention in the code example, how the cost function can be used.

src/qilisdk/backends/qutip_backend.py
```diff
@@ -111,13 +112,16 @@ def _execute_sampling(self, functional: Sampling) -> SamplingResult:
         measurements = sorted(measurements_set)
 
         sim = CircuitSimulator(qutip_circuit)
-        sim.initialize(init_state)
+        # sim.initialize(init_state)
```
fedonman: If we don't need this line, delete it.

fedonman (commented):
Two minor comments and is good to go.

fedonman (approved):

PR #70, opened 2025-09-05
adding basic state tomography

amirazzam2000:
Adding ``StateTomography``functional. This functional can be used to retrieve the full reconstructed state vector at the end of the quantum circuit. 

Example usage: 
```python
from qilisdk.digital import *
from qilisdk.functionals import StateTomography
from qilisdk.backends import CudaBackend

circuit = Circuit(2)
circuit.add(H(0))
circuit.add(CNOT(0, 1))

backend = CudaBackend()
backend.execute(StateTomography(circuit))
```

### DIFF
--- a/src/qilisdk/backends/cuda_backend.py
+++ b/src/qilisdk/backends/cuda_backend.py
@@ -29,13 +29,13 @@ from qilisdk.common.qtensor import QTensor, ket, tensor_prod
 from qilisdk.digital.exceptions import UnsupportedGateError
 from qilisdk.digital.gates import RX, RY, RZ, U1, U2, U3, Adjoint, BasicGate, Controlled, H, M, S, T, X, Y, Z
 from qilisdk.functionals.sampling_result import SamplingResult
-from qilisdk.functionals.state_tomography import StateTomography
 from qilisdk.functionals.state_tomography_result import StateTomographyResult
 from qilisdk.functionals.time_evolution_result import TimeEvolutionResult
 
 if TYPE_CHECKING:
     from qilisdk.digital.circuit import Circuit
     from qilisdk.functionals.sampling import Sampling
+    from qilisdk.functionals.state_tomography import StateTomography
     from qilisdk.functionals.time_evolution import TimeEvolution
 
 
--- a/src/qilisdk/backends/cuda_backend.py
+++ b/src/qilisdk/backends/cuda_backend.py
@@ -210,6 +210,18 @@ class CudaBackend(Backend):
     def _execute_state_tomography(
         self, functional: StateTomography, initial_state: QTensor | None = None, tol: float = -1e12
     ) -> StateTomographyResult:
+        """execute a state tomography protocol.
+
+        Args:
+            functional (StateTomography): the state tomography functional.
+            initial_state (QTensor | None, optional): the initial state of the circuit. Defaults to None.
+                NOTE: this is only used for the quantum reservoir and is not a public attribute.
+            tol (float, optional): the tolerance when computing if the state is positive semi-definite. Defaults to -1e12.
+                TODO (ameer): Move this somewhere else -> will do that in the reservoir pr.
+
+        Returns:
+            StateTomographyResult: the final results of the state tomography
+        """
         logger.info("Executing state tomography")
         self._apply_digital_simulation_method()
         kernel = self._get_cuda_kernel(functional.circuit, initial_state, ignore_measurement=True)
--- a/src/qilisdk/backends/cuda_backend.py
+++ b/src/qilisdk/backends/cuda_backend.py
@@ -14,6 +14,7 @@
 from __future__ import annotations
 
 from enum import Enum
+from itertools import product
 from typing import TYPE_CHECKING, Callable, Type, TypeVar
 
 import cudaq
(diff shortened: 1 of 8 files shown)


src/qilisdk/backends/cuda_backend.py
```diff
@@ -216,6 +207,102 @@ def get_schedule(key: str) -> Callable:
...
 
+    def _execute_state_tomography(
+        self, functional: StateTomography, initial_state: QTensor | None = None, tol: float = -1e12
+    ) -> StateTomographyResult:
+        """execute a state tomography protocol.
+
+        Args:
+            functional (StateTomography): the state tomography functional.
+            initial_state (QTensor | None, optional): the initial state of the circuit. Defaults to None.
+                NOTE: this is only used for the quantum reservoir and is not a public attribute.
```
fedonman, line 219: Why? Users could be able to set an initial_state (simulators allow this, even if qililab doesn't allow it yet).

src/qilisdk/backends/cuda_backend.py
```diff
@@ -216,6 +207,102 @@ def get_schedule(key: str) -> Callable:
             ),
         )
 
+    def _execute_state_tomography(
+        self, functional: StateTomography, initial_state: QTensor | None = None, tol: float = -1e12
+    ) -> StateTomographyResult:
```
fedonman, line 213: `initial_state` should be an attribute of `StateTomography`.

For `tol`, could you explain why we need it?

src/qilisdk/backends/qutip_backend.py
```diff
@@ -259,6 +261,42 @@ def get_coeff(t: float) -> Number:
             ),
         )
 
+    def _execute_state_tomography(
+        self, functional: StateTomography, initial_state: QTensor | None = None
+    ) -> StateTomographyResult:
```
fedonman, line 258: Same comment on `initial_state` as in `CudaBackend`.

src/qilisdk/functionals/state_tomography_result.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def __init__(self, state: QTensor) -> None:
+        """
+        Args:
+            state (QTensor): the state at the end of the execution.
+
+        Raises:
+            ValueError: if the final state is not a ket of a density matrix
+        """
+        self._final_state = state
```
fedonman, line 42: Could be shortened to `_state`. We know it's final, since it is in the results class.

src/qilisdk/functionals/state_tomography_result.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        Gets the number of qubits involved in the measurement.
+
+        Returns:
+            int: The number of qubits measured.
+        """
+        return self._nqubits
+
+    @property
+    def final_state(self) -> QTensor:
```
fedonman, line 72: Same here, could be shortened to `state`.

src/qilisdk/functionals/state_tomography_result.py
```diff
@@ -0,0 +1,128 @@
...
+        Args:
+            state (QTensor): the state at the end of the execution.
+
+        Raises:
+            ValueError: if the final state is not a ket of a density matrix
+        """
+        self._final_state = state
+
+        # Assign nqubits to attribute
+        self._nqubits = self._final_state.nqubits
```
fedonman, line 45: No need to store an attribute. It can be just a variable in the method. 

Property `nqubits` (do we need it?) can `return self.state.nqubits` directly.

src/qilisdk/functionals/state_tomography_result.py
```diff
@@ -0,0 +1,128 @@
...
+        if state.is_ket():
+            probs = np.abs(state.dense.flatten()) ** 2
+        elif state.is_density_matrix():
+            probs = np.real(np.diag(state.dense))
+        else:
+            raise ValueError("State must be a ket or density matrix")
+
+        bitstrings = [format(i, f"0{self._nqubits}b") for i in range(len(state.dense))]
+
+        self._probabilities = dict(zip(bitstrings, probs))
```
fedonman, line 59: Let's write a test checking these calculations. 

Also. there are ways to get the probabilities in pure-sparse format without densifying, but we need a solid test to verify correctness.

src/qilisdk/backends/cuda_backend.py
```diff
@@ -216,6 +207,102 @@ def get_schedule(key: str) -> Callable:
...
+        rho = 0.5 * (rho + rho.conj().T)
+        rho /= np.trace(rho)
+        rho = 0.5 * (rho + rho.conj().T)
+        rho /= np.trace(rho)
+
+        eigvals, eigvecs = np.linalg.eigh(rho)
+        if any(e < tol for e in eigvals):
+            eigvals_clipped = np.clip(eigvals, 0, None)
+            rho = eigvecs @ np.diag(eigvals_clipped) @ eigvecs.conj().T
+            rho /= np.trace(rho)
```
fedonman, line 277: 1) We should add tests
2) Could you explain the implementation? Is this for density matrices? For vector states, CUDAQ offers the `get_state` method: https://nvidia.github.io/cuda-quantum/latest/using/examples/executing_kernels.html#get-state

fedonman (changes requested):
Thanks for splitting this up to a separate PR. I left some comments.

fedonman (commented):
Also, check some analysis of AI on hot paths (code that is slow) and optimizing:
1) the state tomography flow: https://chatgpt.com/s/t_68bb0fe897e0819193f518f908c7015b
2) the probabilities from state calculation:
https://chatgpt.com/s/t_68bb105d8cac8191ae30f2d60d003415

src/qilisdk/backends/cuda_backend.py
amirazzam2000: because I didn't want to make the tomography very different from sampling. I wanted them to be something like sister methods so it's easy to switch between them. but we can add initial states to both of them and hardcode it to |0> in the case of the hardware execution

src/qilisdk/functionals/state_tomography_result.py
amirazzam2000: I was planning to add the tests with the reservoir pr, but we can split them if you think it's better

src/qilisdk/backends/cuda_backend.py
amirazzam2000: yes get state doesn't work with density matrices which is why i had to build it from scratch

amirazzam2000: but on that note, we know that the time evolution is not working for denisty matrices which is preventing us from having reservoir simulation with cudaq anyway, so one option would be to return a not supported error for reservoirs with cudaq and just use this get_state method here

fedonman:
Closing it as no longer relevant.

PR #77, opened 2025-09-18
[ALGP-324] updating schedule structure

amirazzam2000:
Implemented schedule plotting with theme options consistent with those available for circuits, along with new schedule style classes to improve customization. A new `LinearSchedule` has been added, implementing linear interpolation between user-defined coefficients across the schedule. In addition, both `Schedule` and `LinearSchedule` now enforce that the total duration `T` must be divisible by the time step `dt`.

### DIFF
--- a/changes/77.feature.md
+++ b/changes/77.feature.md
@@ -1,7 +1,1 @@
-
-Features
-
-- Added schedule plotting, with theme options similar to that of the circuit. 
-- Adding schedule style classes for added customizability 
-- Adding ``LinearSchedule`` which implements linear interpolation between the user defined coefficient during the schedule.
-- Updated ``Schedule`` and ``LinearSchedule`` to enforce that T is divisible by dt.
+Implemented schedule plotting with theme options consistent with those available for circuits, along with new schedule style classes to improve customization. A new `LinearSchedule` has been added, implementing linear interpolation between user-defined coefficients across the schedule. In addition, both `Schedule` and `LinearSchedule` now enforce that the total duration `T` must be divisible by the time step `dt`.
--- a/changes/77.feature.md
+++ b/changes/77.feature.md
@@ -1,0 +1,7 @@
+
+Features
+
+- Added schedule plotting, with theme options similar to that of the circuit. 
+- Adding schedule style classes for added customizability 
+- Adding ``LinearSchedule`` which implements linear interpolation between the user defined coefficient during the schedule.
+- Updated ``Schedule`` and ``LinearSchedule`` to enforce that T is divisible by dt.
--- a/src/qilisdk/analog/linear_schedule.py
+++ b/src/qilisdk/analog/linear_schedule.py
@@ -1,0 +1,119 @@
+# Copyright 2025 Qilimanjaro Quantum Tech
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import annotations
+
+from qilisdk.analog.hamiltonian import Hamiltonian
+from qilisdk.analog.schedule import Schedule
+from qilisdk.common.variables import Number, Parameter, Term
+from qilisdk.yaml import yaml
+
+
+@yaml.register_class
+class LinearSchedule(Schedule):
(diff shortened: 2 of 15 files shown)


src/qilisdk/analog/schedule.py
```diff
@@ -52,7 +53,12 @@ def __init__(
         Raises:
             ValueError: If the provided schedule references Hamiltonians that have not been defined.
         """
-
+        if abs(T % dt) > 1e-12:  # noqa: PLR2004
```
fedonman: If `T` and `dt` are integers `T % dt != 0` should suffice.

fedonman: So probably move this check after the `isinstance(T, int)` and `isinstance(dt, int)` checks

src/qilisdk/analog/schedule.py
```diff
@@ -493,3 +499,123 @@ def __next__(self) -> Hamiltonian:
...
+            raise ValueError("Something unexpected happened while retrieving the coefficient.")
+        alpha: float = (t - prev_idx) / (next_idx - prev_idx)
+        return (1 - alpha) * prev_expr + alpha * next_expr
+
+    def __getitem__(self, time_step: int) -> Hamiltonian:
+        ham = Hamiltonian()
+        for ham_label in self._hamiltonians:
+            coeff = self.get_coefficient(time_step * self.dt, ham_label)
+            ham += coeff * self._hamiltonians[ham_label]
+        return ham.get_static_hamiltonian()
```
fedonman: I think there is no circular dependency, so let's move this to a separate file `linear_schedule.py`.

fedonman: Also, expose `LinearSchedule` to the `__init__.py` so it can be imported like `from qilisdk.analog import LinearSchedule`.

src/qilisdk/utils/visualization/style.py
```diff
@@ -0,0 +1,176 @@
...
+    )
+    math_fontfamily: str | None = Field(default=None, description="Math text family, e.g. 'dejavusans', 'cm', or None.")
+
+    # Ticks
+    xtick_fontsize: int = 12
+    ytick_fontsize: int = 12
+    tick_color: Optional[str] = None  # If None, use theme.on_background
+
+    # Misc
+    tight_layout: bool = True
```
fedonman: Should we move common fields like font-related, to a common `Style` model, and the rest inherit from it?

changes/77.feature.md
```diff
@@ -0,0 +1,7 @@
+
+Features
+
+- Added schedule plotting, with theme options similar to that of the circuit. 
+- Adding schedule style classes for added customizability 
+- Adding ``LinearSchedule`` which implements linear interpolation between the user defined coefficient during the schedule.
+- Updated ``Schedule`` and ``LinearSchedule`` to enforce that T is divisible by dt.
```
fedonman: Don't write changelogs with bullet points nor add titles like `Features`, because it will be messed up by `towncrier` when doing the release. Use paragraph-style texts with no title.

```suggestion
Implemented schedule plotting with theme options consistent with those available for circuits, along with new schedule style classes to improve customization. A new `LinearSchedule` has been added, implementing linear interpolation between user-defined coefficients across the schedule. In addition, both `Schedule` and `LinearSchedule` now enforce that the total duration `T` must be divisible by the time step `dt`.
```

fedonman (changes requested):
Nice!!! Small changes requested.

src/qilisdk/analog/schedule.py
amirazzam2000: yeah fair that was there from when they were floats cuz the modulus is not accurate always. I can change it to 0 now that they are ints.

src/qilisdk/utils/visualization/style.py
amirazzam2000: yeah I can group common features into a single parent class 👍

src/qilisdk/analog/linear_schedule.py
```diff
@@ -0,0 +1,119 @@
...
+        if prev_expr is None and next_expr is not None:
+            return next_expr
+        if next_expr is None and prev_expr is not None:
+            return prev_expr
+
+        # linear interpolation (keeps expressions if they are Terms/Parameters)
+        if next_idx is None or prev_idx is None or prev_expr is None or next_expr is None:
+            raise ValueError("Something unexpected happened while retrieving the coefficient.")
+        alpha: float = (t - prev_idx) / (next_idx - prev_idx)
+        return (1 - alpha) * prev_expr + alpha * next_expr
```
fedonman: These two methods seem to be 90% the same. Can we extract the common code? What is the difference between the two? Also, could you add tests for the LinearSchedule?

fedonman: Just saw that the tests are inside `test_schedule.py` Let's move those in separate file as well.

fedonman (commented):
Small comment and is good to go 💯

fedonman (approved):
💯 💯

PR #78, opened 2025-09-23
[ALGP-328] Fixing Time Evolution bugs

amirazzam2000:
We moved the dt and T in the `Schedule` class back to being floats instead of integers. Added helpful methods to the hamiltonian class (`commutator`, `anitcommutator`, `vector_norm`, `frobenius_norm`, `trace`)


src/qilisdk/analog/hamiltonian.py
```diff
@@ -636,6 +729,63 @@ def looks_like_number(text: str) -> bool:
...
+    def commutator(self, h: Hamiltonian) -> Hamiltonian:
+        """compute the commutator of the current hamiltonian with another hamiltonian (h)
+
+        Args:
+            h (Hamiltonian): the second hamiltonian.
+
+        Returns:
+            Hamiltonian: the commutator.
+        """
+        h0 = copy.copy(self)
```
fedonman: Do we need to copy here?

src/qilisdk/analog/hamiltonian.py
```diff
@@ -636,6 +729,63 @@ def looks_like_number(text: str) -> bool:
...
+    def anticommutator(self, h: Hamiltonian) -> Hamiltonian:
+        """compute the anticommutator of the current hamiltonian with another hamiltonian (h)
+
+        Args:
+            h (Hamiltonian): the second hamiltonian.
+
+        Returns:
+            Hamiltonian: the anticommutator.
+        """
+        h0 = copy.copy(self)
```
fedonman: Do we need to copy here?

src/qilisdk/backends/qutip_backend.py
```diff
@@ -77,8 +77,9 @@ class QutipBackend(Backend):
     ideal for local development, CI pipelines, and educational notebooks.
     """
 
-    def __init__(self) -> None:
+    def __init__(self, nsteps: int | None = None) -> None:
         """Instantiate a new :class:`QutipBackend`."""
+        self.nsteps = nsteps or 10000
```
fedonman: If we want a default value of 10_000 add it directly to the signature.
```suggestion
    def __init__(self, nsteps: int = 10_000) -> None:
        """Instantiate a new :class:`QutipBackend`."""
        self.nsteps = nsteps
```

fedonman: Also add the `nsteps` to the docstring, explaining what it does.

fedonman (changes requested):
Made some minor comments. Also add a changelog entry!

fedonman (approved):

PR #81, opened 2025-10-06
[ALGP-334] Updating SpeQtrum to match the GraphQL structure

amirazzam2000:
Updated SpeQtrum models to latest schema

### DIFF
--- a/src/qilisdk/speqtrum/speqtrum.py
+++ b/src/qilisdk/speqtrum/speqtrum.py
@@ -24,7 +24,6 @@ import httpx
 from loguru import logger
 from pydantic import TypeAdapter
 
-import qilisdk
 from qilisdk.functionals import Sampling, TimeEvolution, VariationalProgram
 from qilisdk.settings import get_settings
 from qilisdk.speqtrum.experiments import ExperimentFunctional, RabiExperiment, T1Experiment
--- a/src/qilisdk/speqtrum/speqtrum.py
+++ b/src/qilisdk/speqtrum/speqtrum.py
@@ -37,6 +37,7 @@ from .speqtrum_models import (
     JobId,
     JobInfo,
     JobStatus,
+    JobType,
     RabiExperimentPayload,
     SamplingPayload,
     T1ExperimentPayload,
@@ -332,7 +333,7 @@ class SpeQtrum:
             type=ExecuteType.SAMPLING,
             sampling_payload=SamplingPayload(sampling=sampling),
         )
-        json = {"device_id": device_id, "payload": payload.model_dump_json(), "job_type": "digital", "meta": {}}
+        json = {"device_id": device_id, "payload": payload.model_dump_json(), "job_type": JobType.DIGITAL, "meta": {}}
         logger.debug("Executing Sampling on device {}", device_id)
         with httpx.Client() as client:
             response = client.post(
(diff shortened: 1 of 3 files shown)


src/qilisdk/speqtrum/speqtrum.py
```diff
@@ -332,7 +332,7 @@ def _submit_sampling(self, sampling: Sampling, device_id: int) -> int:
             type=ExecuteType.SAMPLING,
             sampling_payload=SamplingPayload(sampling=sampling),
         )
-        json = {"device_id": device_id, "payload": payload.model_dump_json(), "meta": {}}
+        json = {"device_id": device_id, "payload": payload.model_dump_json(), "job_type": "digital", "meta": {}}
```
fedonman: Did we add this to the `execute` endpoint? It seems redundant since payload has this information. In any case, if we do need it, pass the enum `ExecuteType` here. E.g. `ExecuteType.SAMPLING`.

amirazzam2000: yeah I agree, but this was added by access to the api. I am adding it for now, but on Wednesday we should talk to them and probably remove it from the api because as you said it's redundant.

fedonman: I just thought they might need it for the website, which makes sense since the payload isn't deserialized by Public API.

amirazzam2000: but in theory the website shouldn't use the submit endpoint.

fedonman: It's a field of the job [https://www.notion.so/qilimanjaro/Agreement-PublicAPI-GraphQL-contract-2407eec14c538023ac4df6ae2858e3a3?source=copy_link#2407eec14c53[…]cdd8b1389850](https://www.notion.so/qilimanjaro/Agreement-PublicAPI-GraphQL-contract-2407eec14c538023ac4df6ae2858e3a3?source=copy_link#2407eec14c5380db9071cdd8b1389850)

amirazzam2000: good point, maybe it's easier for them to get the information this way. But then we can consider removing the redundant Execution Type in the payload

fedonman: No we need it in the `qilisdk-speqtrum`

amirazzam2000: but we can get that information from access, they should be able to pass it down to us

fedonman: No `qilisdk-speqtrum` doesn't know anything about jobs

fedonman: and payload is important in code anyway for pydantic to serialize/deserialize

amirazzam2000: I guess for now it's fine to keep it like this

fedonman: we just have to make sure that the JobType and ExecuteType are the same

fedonman: both get the value from the enum

fedonman: not an arbitrary string.

amirazzam2000: yeah I can change that.

fedonman (approved):

PR #82, opened 2025-10-08
[QHC-1100] Fix Circuit's draw method

elygoner:
Fixed some bugs:
* Wires now end at the same spot
* There isn't an extra margin after the end of the circuit
* All except  for one of the gate superposition issues have been fixed.
     - The remaining one in question emerges when you have long controlled gates. I Don't see a good way of fixing it without some major changes in the class workflow


amirazzam2000 (changes requested):
Amazing job!! it looks really good so far

I found an issue with this example:

Image

here is the code: 

```
from qilisdk.digital import * 

circ = Circuit(4)
circ.add(H(0))
circ.add(CNOT(0,1))
circ.add(CNOT(0,2))
circ.add(CNOT(0,3))
circ.add(RZ(0, phi=np.pi/2))
circ.add(X(1))
circ.add(X(1))
# circ.add(CNOT(0,1))
circ.add(CNOT(0,2))
circ.add(CNOT(0,3))
circ.add(H(0))

circ.draw()
```

fedonman (approved):
If many changes need to fix the other issues, I would merge this to have a better version in the release, and continue in another PR for the rest.

elygoner:
There is some wierd stuff with the style margins, but apart from that everything works correctly, I think.
If you find any other bugs, tell me.

fedonman (changes requested):
I run the following code:

```
circuit = Circuit(5)
for i in range(5):
    circuit.add(X(i))
for i in range(4):
    circuit.add(CNOT(4, i))
    circuit.add(U3(i+1, theta=i, phi=i*2, gamma=i*3))
    circuit.add(RX(0, theta=i).controlled(i+1))
for i in range(5):
    circuit.add(M(i))

circuit.draw()
```

and I get the following plot as a result. The order of gates is wrong.

image

For comparison, this was plotted before:

image

fedonman:
I understand they are equivalent circuits, but we shouldn't change the order of the gates that the user has decided to have. At least not by default, we could have a `layout` property in the styling with values `normal`/`compact`. But the default behaviour should always respect the insertion order.

amirazzam2000 (approved):
Amazing!! I tested a couple of examples and they all seem to work perfectly 👍 
thanks a lot Ely!! 💯

src/qilisdk/utils/visualization/style.py
```diff
@@ -146,4 +148,5 @@ class ScheduleStyle(Style):
     )
 
     # Misc
+    compact_depth: bool = Field(default=True, description="If True , minimizes the layers to highlight circuit depth")
```
amirazzam2000: you don't need to add this to the Schedule style this is a circuit specific thing

amirazzam2000 (changes requested):

src/qilisdk/utils/visualization/style.py
amirazzam2000: Also I tested the property compact_depth and it doesn't seem to have any effect

src/qilisdk/utils/visualization/circuit_renderers.py
```diff
@@ -15,7 +15,7 @@
 from __future__ import annotations
 
 from fractions import Fraction
-from typing import TYPE_CHECKING, Final, Iterable
+from typing import TYPE_CHECKING, Dict, Final, Iterable, List
```
fedonman: Don't use typing's `List` and `Dict`, it's legacy, we always use built-in `list` and `dict` for annotations.

src/qilisdk/utils/visualization/style.py
```diff
@@ -95,6 +95,8 @@ class CircuitStyle(Style):
     target_r: float = Field(default=0.12, description="Radius (inches) of ⊕ target circle and SWAP half-width.")
     control_r: float = Field(default=0.05, description="Radius (inches) of a filled control dot.")
 
+    compact_depth: bool = Field(default=True, description="If True , minimizes the layers to highlight circuit depth")
```
fedonman: I would prefer this to be a `layout` with literal fields of possible values. That way 1. is more clear and consistent with the rest of the fields, and 2. we can add more layout options in the future. E.g.

```
layout: Literal["normal", "compact"] = Field(
        default="normal", description="..."
    )

Comment (fedonman) [reply src/qilisdk/utils/visualization/style.py]: Also, the default value must be `False`, or the equivalent `normal` in the proposed refactor.

**Review by fedonman — CHANGES_REQUESTED** [review]
Awesome work! Made some comments.

File: src/qilisdk/utils/visualization/style.py (hunk 1 above)
Comment (fedonman) [reply src/qilisdk/utils/visualization/style.py]: I agree on removing it from the Schedule. For the second comment, it worked correctly for me.

File: src/qilisdk/utils/visualization/style.py (hunk 3 above)
Comment (elygoner) [reply src/qilisdk/utils/visualization/style.py]: That one last one, we should discus. @amirazzam2000 Disagrees with that assessment.

Comment (amirazzam2000) [reply src/qilisdk/utils/visualization/style.py]: Yes I wanted to talk to Vyron about this. I think it's more useful to see the compact circuit because it represents the order of operations better.

Comment (fedonman) [reply src/qilisdk/utils/visualization/style.py]: @amirazzam2000 @elygoner

Please, don't confuse the **order of operations** with the visualization. `draw()` simply provides a visualization of the circuit so humans can easily understand it. The order of operations is determined **after** transpilation and **after** compilation to qprogram and **after** compilation of QProgram to qpysequence. And it heavily depends on the hardware, and the native gates we have chosen. E.g. three single-qubit gates might appear consecutive one after the other in the circuit, but on hardware they will be done simultaneously as a single rotation. Or, you might think a CNOT(0, 2) affects only the control and target qubits, but on Soprano chip the qubit 1 must be parked, so no gate on qubit 1 can be executed at that time.

Coming back to the visualization, as a user I want the **default** behavior to be the most predictable one, which is the order of insertion of the gates in the circuit. Take the following example.

```
circuit = Circuit(5)
circuit.add(H(1))
circuit.add(CNOT(0, 2))
circuit.add(RX(2, theta=0.5))
```

With `normal` layout I get the predictable:

image

With `compact` layout I get:

image

Which is different that the insertion order (and actually even different than the actual order of operations in hardware...) . If the user wanted this circuit he would do this instead:

```
circuit = Circuit(5)
circuit.add(CNOT(0, 2))
circuit.add(H(1))
circuit.add(RX(2, theta=0.5))
```

There are numerous examples like this. Again, we are not removing the `compact` choice, it is there if someone wants it.

**_Default behavior should be the most predictable_** is a software design principle: https://en.wikipedia.org/wiki/Principle_of_least_astonishment

amirazzam2000: okay yeah, we can have that as the default behavior and add in the docs that you can have a compact view of the circuit in case you need it. 👍 

@elygoner do you mind adding a line in the docs on how to apply this compact drawing? if you need help with that text me and I can help you

fedonman (approved):
great work 💯

PR #86, opened 2025-10-22
Add job name to SpeQtrum

amirazzam2000:
adding ``job_name`` to speqtrum's ``submit`` method so the user is able to give their jobs custom names.

### DIFF
--- a/changes/86.feature.md
+++ b/changes/86.feature.md
@@ -1,1 +1,1 @@
-adding ``job_name`` to speqtrum's submit so the user is able to give their jobs custom names.
+adding ``job_name`` to speqtrum's ``submit``method so the user is able to give their jobs custom names.
--- a/changes/86.feature.md
+++ b/changes/86.feature.md
@@ -1,0 +1,1 @@
+adding ``job_name`` to speqtrum's submit so the user is able to give their jobs custom names.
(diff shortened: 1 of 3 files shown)


src/qilisdk/speqtrum/speqtrum.py
```diff
@@ -288,7 +294,9 @@ def wait_for_job(
             logger.debug("Job {} still {}, sleeping {}s", id, current.status.value, poll_interval)
             time.sleep(poll_interval)
 
-    def submit(self, functional: PrimitiveFunctional | ExperimentFunctional, device: str) -> int:
+    def submit(
+        self, functional: PrimitiveFunctional | ExperimentFunctional, device: str, job_name: str = "MyJob"
```
fedonman: Should we force the default name from QiliSDK-side, or send `None` as default and let SpeQtrum decide the default naming?

fedonman: With the current implementation, SpeQtrum can't decide: we are always sending a job name.

amirazzam2000: okay I will change it 👍

changes/86.feature.md
```diff
@@ -0,0 +1 @@
+adding ``job_name`` to speqtrum's submit so the user is able to give their jobs custom names.
```
fedonman: "Added", `submit` method

fedonman (approved):
Awesome! Do the minor edit in the changelog, before merging.

PR #95, opened 2025-11-05
[SDK 93] fixing schedule issues

amirazzam2000:
Fixed `Schedule.add_hamiltonian` so time-dependent coefficient functions populate every discrete time step.

Added regression tests to cover function-driven schedules and validate that only parameters appear in the generated coefficients.


fedonman (approved):
💯 💯

PR #96, opened 2025-11-06
[SDK 94]  Adding an OpenFermion QubitOperator converter.

amirazzam2000:
Added a ``QiliSDK`` to ``OpenFermion`` converter. This allows to translate ``QubitOperator`` from ``OpenFermion`` to ``QiliSDK``'s ``Hamiltonian`` Objects and vice versa.

This optional can be installed using ``pip install qilisdk[openfermion]``.

here is an example of the usage: 
```python
from openfermion.hamiltonians import jellium_model
from openfermion.transforms import fourier_transform, jordan_wigner
from openfermion.utils import Grid

from qilisdk.utils.openfermion import openfermion_to_qilisdk, qilisdk_to_openfermion

# Let's look at a very small model of jellium in 1D.
grid = Grid(dimensions=1, length=3, scale=1.0)
spinless = True

# Get the momentum Hamiltonian.
momentum_hamiltonian = jellium_model(grid, spinless)
momentum_qubit_operator = jordan_wigner(momentum_hamiltonian)
momentum_qubit_operator.compress()

# Fourier transform the Hamiltonian to the position basis.
position_hamiltonian = fourier_transform(momentum_hamiltonian, grid, spinless)
position_qubit_operator = jordan_wigner(position_hamiltonian)
position_qubit_operator.compress()

qilisdk_ham = openfermion_to_qilisdk(position_qubit_operator)
openfermion_ham = qilisdk_to_openfermion(qilisdk_ham)

```


src/qilisdk/utils/openfermion/openfermion.py
```diff
@@ -0,0 +1,45 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from openfermion import QubitOperator
+
+from qilisdk.analog import Hamiltonian, PauliI, PauliX, PauliY, PauliZ
+
+
+def openfermion_to_qili(of_ham: QubitOperator) -> Hamiltonian:
```
fedonman: No abbreviations please! 🥹`qili` should be `qilisdk`, `of_ham` I think is confusing, why not `operator` or `qubit_operator` as the class? Otherwise, full name `of_hamiltonian`.
```suggestion
def openfermion_to_qilisdk(qubit_operator: QubitOperator) -> Hamiltonian:
```

src/qilisdk/utils/openfermion/openfermion.py
```diff
@@ -0,0 +1,45 @@
...
+
+    return Hamiltonian(
+        {
+            (tuple((pauli_map[op](q)) for q, op in term) if len(term) > 0 else (PauliI(0),)): coeff
+            for term, coeff in of_ham.terms.items()
+        }
+    )
+
+
+def qili_to_openfermion(qili_ham: Hamiltonian) -> QubitOperator:
```
fedonman:
```suggestion
def qilisdk_to_openfermion(hamiltonian: Hamiltonian) -> QubitOperator:
```

fedonman: Also, will these classes (QubitOperator/Hamiltonian) be the only ones that we will ever convert? Because if there is a chance to add conversions between other classes we should name the functions to be more specific, e.g. `qubit_operator_to_hamiltonian` or something similar. E.g. OpenFermion has the `FermionOperator` class as well.

changes/96.feature.md
```diff
@@ -0,0 +1 @@
+Added a ``QiliSDK`` to ``OpenFermion`` converter. This allows to translate ``QubitOperator`` from ``OpenFermion`` to ``QiliSDK``'s ``Hamiltonian`` Objects and vice versa.
```
fedonman: Mention the specific utility functions here, and add a code example (could be the same as in the test case). Also mention that this is an optional extra, and how to install it. (could be better to start with this actually)

tests/utils/openfermion/test_openfermion.py
```diff
@@ -0,0 +1,46 @@
...
+
+    # Get the momentum Hamiltonian.
+    momentum_hamiltonian = jellium_model(grid, spinless)
+    momentum_qubit_operator = jordan_wigner(momentum_hamiltonian)
+    momentum_qubit_operator.compress()
+
+    # Fourier transform the Hamiltonian to the position basis.
+    position_hamiltonian = fourier_transform(momentum_hamiltonian, grid, spinless)
+    position_qubit_operator = jordan_wigner(position_hamiltonian)
+    position_qubit_operator.compress()
```
fedonman, line 24: Are the two `.compress()` calls needed? What's their functionality?

fedonman (changes requested):
Nice work and useful! Left some comments.

src/qilisdk/utils/openfermion/openfermion.py
amirazzam2000: I think this should be the only one to convert. Because it's the only one that can be mapped to something that can be executed on our hardware eventually.

tests/utils/openfermion/test_openfermion.py
amirazzam2000: they are compressing hamiltonians in different representations, so one is being done before the jordan wigner transformation and one after. This is a copy from an example in openfermion's docs, i wanted to use a real example.

fedonman (approved):

PR #98, opened 2025-11-06
[SDK-96] Use `complex_precision` setting

fedonman:
- Users can now set the environment variable `QILISDK_COMPLEX_PRECISION` to either `COMPLEX_64` or `COMPLEX_128`, making it easy to globally select the complex dtype used across dense and sparse matrices; if the variable is unset we continue to default to the project-wide precision defined in `settings.py`.
- Fixed the Hamiltonian `to_matrix()` and `to_qtensor()` routines to skip unnecessary sparse→dense→sparse conversions by caching single-qubit factors, resulting in noticeably faster tensor construction for medium- and large-qubit systems.


NataliaSirera (approved):

src/qilisdk/analog/hamiltonian.py
```diff
@@ -184,28 +201,45 @@ def __rtruediv__(self, _: Number | PauliOperator | Hamiltonian) -> Hamiltonian:
 class PauliZ(PauliOperator):
     # __slots__ = ()
     _NAME: ClassVar[str] = "Z"
-    _MATRIX: ClassVar[np.ndarray] = np.array([[1, 0], [0, -1]], dtype=complex)
+    _MATRIX: ClassVar[np.ndarray] = np.array([[1, 0], [0, -1]], dtype=np.complex128)
```
amirazzam2000: Why are these not following the dynamic typing?

src/qilisdk/digital/gates.py
```diff
@@ -34,6 +35,10 @@
 TBasicGate = TypeVar("TBasicGate", bound="BasicGate")
 
 
+def _complex_dtype() -> np.dtype:
```
amirazzam2000, line 38: why don't we make this a method in settings so we don't have to create it in each file?

tests/analog/test_hamiltionian.py
```diff
@@ -18,13 +18,40 @@
...
+        settings.complex_precision = Precision.COMPLEX_64
+        assert PauliX(0).matrix.dtype == np.dtype(np.complex64)
+    finally:
+        settings.complex_precision = original_precision
+
+
+def test_hamiltonian_matrix_dtype_updates_with_settings_change():
+    settings = get_settings()
+    original_precision = settings.complex_precision
+    try:
```
amirazzam2000: why do we have a try statement in the test?

src/qilisdk/analog/hamiltonian.py
amirazzam2000, line 358: I want to test this method locally, I will notify you when my tests are done!

amirazzam2000: I would add two checks here:

1. confirm that the length of the terms list is equal to the number of qubits in the hamiltonian. 
2. no two pauliOperators are acting on the same qubit. 

Because this method assumes that you are passing it a single PauliOperator per-qubit, so if there are two pauliOperators acting on the same qubits the results are are wrong.

amirazzam2000 (changes requested):

src/qilisdk/digital/gates.py
Lumorti: I guess this is just a shorthand, so maybe it's not worth putting it in settings. I do agree that it's weird to have it defined in multiple locations, but maybe it's also weird to have to import a one line function every time

tests/analog/test_hamiltionian.py
Lumorti: I'll remove it

src/qilisdk/analog/hamiltonian.py
Lumorti: Added now

amirazzam2000 (approved):

PR #101, opened 2025-11-13
[SDK-104] Implement CircuitTranspiler and DecomposeMultiControlledGatesPass

fedonman:
Introduced the first version of the `CircuitTranspiler` alongside the `DecomposeMultiControlledGatesPass` that decomposes any multi-controlled single-qubit gate, enabling both CUDA and QuTiP backends to execute circuits containing such constructs. The pass is wired directly into both backends today while the full transpiler pipeline is staged for future expansion.


src/qilisdk/backends/cuda_backend.py
```diff
@@ -137,13 +138,14 @@ def _execute_sampling(self, functional: Sampling) -> SamplingResult:
         kernel = cudaq.make_kernel()
         qubits = kernel.qalloc(functional.circuit.nqubits)
 
-        for gate in functional.circuit.gates:
+        transpiled_circuit = DecomposeMultiControlledGatesPass().run(functional.circuit)
```
amirazzam2000, line 141: I don't think we should run this by default, we should allow the user to specify the passes in the sampling functional or in the circuit. But this will complicate the simulation and add lots of overhead.

src/qilisdk/backends/qutip_backend.py
```diff
@@ -117,18 +118,17 @@ def _execute_sampling(self, functional: Sampling) -> SamplingResult:
...
         init_state = tensor(*[basis(2, 0) for _ in range(functional.circuit.nqubits)])
 
         measurements_set = set()
         for m in functional.circuit.gates:
             if isinstance(m, M):
                 measurements_set.update(list(m.target_qubits))
-
         measurements = sorted(measurements_set)
 
+        transpiled_circuit = DecomposeMultiControlledGatesPass().run(functional.circuit)
```
amirazzam2000, line 130: same here.

amirazzam2000 (changes requested):

src/qilisdk/backends/cuda_backend.py
fedonman: We need to run this automatically for now, because otherwise circuits with multi-controlled gates cannot be simulated in either backends.

Transpiler will be exposed to the users when we have added the rest of the passes. (check changelog)

amirazzam2000 (approved):
thanks vyron!

PR #107, opened 2025-12-02
[SDK-114] Wrong color of linked elements in light mode

Lumorti:
Fixed a display bug in which inline code blocks in headers were invisible. This was due to the previous rule only accounting for "a" elements in headers and not ".pre" span elements. The reason there was a ".pre" block in the header is because of a broken class link, which will be fixed along with others in a different PR.


fedonman (approved):
Nice catch! 💯

PR #108, opened 2025-12-02
[SDK-120] Docs: Broken class links

Lumorti:
Fixed a number of broken/missing API links in the docs and in general made the docs consistent with the API reference.

Also:
 - Remove references to optimize() since it has since been moved into execute()
 - Fixed a broken output box
 - Fixed several lists that weren't displaying correctly
 - Removed an unused import from a code block in core.rst
 - Various other typos, such as missing "a"s before certain terms (keeping consistency with previous uses of said terms)
 - Wrapped some very long lines (as per the style throughout the rest of the files)


fedonman (approved):
Thank you @Lumorti 💯

PR #109, opened 2025-12-04
[SDK-148] Improve Versioning Navbar

fedonman:
Docs publishing now points the root of the site to the latest release docs and keeps "main" as a clearly labeled development build, with the version picker reorganized into separate Releases and Development sections so visitors land on the most current tag while still finding the dev docs easily.


amirazzam2000 (approved):
nice! thank vyron!!

PR #113, opened 2025-12-09
[SDK-118] Build QAOA Algorithm

Lumorti:
- Feature: Added a QAOA ansatz, as well as corresponding documentation. This ansatz combines a problem Hamiltonian and a mixer Hamiltonian, performing the alternating time evolution of both, scaled by some parameters. By adjusting the values of said parameters using a Variational functional, the energy of the problem Hamiltonian is minimized thus a solution to the original problem is found.

- Feature: Hamiltonian class now has a get_commuting_partitions() method, which generates partitions each containing only commuting Pauli strings. This is useful for time evolution, since each partition can then be trivially exponentiated without Trotterization.

 - Feature: All parameterizable gates (i.e. RX, RY, RZ, U1, U2, U3) now accept Terms rather than only Parameters, allowing for more general expressions. This is useful if one wants two gates in the same circuit parameterized by the same variable, but with different scaling, e.g. Rz(2*theta) and RZ(3*theta).

 - Bugfix: Parameters with equal bounds (i.e. min == max) are no longer optimized.


src/qilisdk/digital/ansatz.py
```diff
@@ -198,3 +202,213 @@ def _build_circuit(self) -> None:
...
+                mixer_type=None,
+                trotter_steps=1,
+            )
+            ansatz.draw()
+
+    """
+
+    def __init__(
+        self,
+        nqubits: int,
```
amirazzam2000: I don't think we need to have this, we can gather the number of qubits from the mixer and problem hamiltonians.

src/qilisdk/digital/gates.py
```diff
@@ -229,8 +234,17 @@ def __init__(self, target_qubits: tuple[int, ...], parameters: dict[str, Paramet
 
         self._target_qubits: tuple[int, ...] = target_qubits
         self._parameters: dict[str, Parameter] = parameters or {}
+        self._parameter_transforms: dict[str, Term] = parameter_transforms or {}
         self._matrix: np.ndarray = self._generate_matrix()
 
+        # Check that anything mentioned in parameter_transforms is also in parameters
+        for term in self._parameter_transforms:
```
amirazzam2000, line 243: we should also check that all variables in the terms are Parameters and not BasicVariable (aka. checking that term.is_parameterized_term() == True)

src/qilisdk/digital/gates.py
```diff
@@ -722,6 +736,35 @@ def _generate_matrix(self) -> np.ndarray:  # noqa: PLR6301
         return np.array([[1, 0], [0, np.exp(1j * np.pi / 4)]], dtype=complex)
 
 
+def _process_param(
```
amirazzam2000, line 748: why are we updating values by reference here instead of just returning the `params_to_init` and `terms_to_init` dicts?

amirazzam2000: okay I actually later noticed that this is more efficient for gates with two parameters

src/qilisdk/digital/ansatz.py
Lumorti: I thought about it, but in the end I left it to keep with the syntax of the HardwareEfficientAnsatz. But if you think it's better without I'll get rid of it!

src/qilisdk/digital/gates.py
Lumorti: Sure thing

src/qilisdk/digital/ansatz.py
amirazzam2000: yeah cuz I think this might cause a lot of confusion, and in the hardware efficient ansatz we have no other way of getting this information.

amirazzam2000 (approved):

PR #114, opened 2025-12-10
[SDK-121] Add Zero Tolerance

Lumorti:
Tolerances for checking if something is zero have been unified as a new global setting: "zero_tolerance".


src/qilisdk/settings.py
```diff
@@ -62,6 +62,10 @@ class QiliSDKSettings(BaseSettings):
         default="urn:qilimanjaro.tech:public-api:beren",
         description="Audience claim expected in the JWT used for authentication. [env: QILISDK_SPEQTRUM_AUDIENCE]",
     )
+    zero_tolerance: float = Field(
```
fedonman: I think it's better to rename this to `atol`, absolute tolerance, which is the standard naming, plus it will be used to comparing equality in general, not just to zero.

We could also introduce the relative tolerance, `rtol` parameter, for the methods that need it like `np.allclose`, although I don't know if it's an overkill or not.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -506,14 +507,14 @@ def _format_coeff(c: complex) -> str:
         return " ".join(parts)
 
     @classmethod
-    def from_qtensor(cls, tensor: QTensor, tol: float = 1e-10, prune: float = 1e-12) -> Hamiltonian:
+    def from_qtensor(cls, tensor: QTensor, tol: float = get_settings().zero_tolerance, prune: float = get_settings().zero_tolerance) -> Hamiltonian:
```
fedonman: There is a subtle bug with this: Python evaluates default arguments once at **import time**, so `tol` would be locked to whatever `get_settings().zero_tolerance` was during module import and won’t reflect later config changes.

The safe, recommended pattern is `tol: float | None = None` (same for prune) and inside set `tol = tol or get_settings().zero_tolerance`.

In case we wanted the setting to be a fixed constant, we only need to evaluate it once, so make it explicit with a module-level `DEFAULT_ATOL = get_settings().zero_tolerance` and inside the method do `tol = tol or DEFAULT_ATOL`. (I think in this case we don't, since in other parts of the PR, the value is read dynamically at runtime)

src/qilisdk/core/qtensor.py
```diff
@@ -433,19 +434,19 @@ def to_density_matrix(self) -> QTensor:
             raise ValueError("Invalid object for density matrix conversion.")
 
         tr = float(np.real(rho.trace()))
-        if tr == 0.0:
+        if abs(tr) < get_settings().zero_tolerance:
             raise ValueError("Cannot normalize density matrix with zero trace.")
         return QTensor(rho.data / tr)  # keep it sparse
 
-    def is_density_matrix(self, tol: float = 1e-8) -> bool:
+    def is_density_matrix(self, tol: float = get_settings().zero_tolerance) -> bool:
```
fedonman: here too.

src/qilisdk/core/qtensor.py
```diff
@@ -471,13 +472,13 @@ def is_density_matrix(self, tol: float = 1e-8) -> bool:
                 return False
         return lam_min >= -tol
 
-    def is_hermitian(self, tol: float = 1e-8) -> bool:
+    def is_hermitian(self, tol: float = get_settings().zero_tolerance) -> bool:
```
fedonman: here too.

fedonman (changes requested):
Great work Luke! Left some comments.

src/qilisdk/analog/hamiltonian.py
Lumorti: Ah, good catch! Changed.

src/qilisdk/settings.py
Lumorti: Yes I was wondering about the name, I agree that atol makes more sense.

src/qilisdk/settings.py
```diff
@@ -62,10 +62,14 @@ class QiliSDKSettings(BaseSettings):
...
     )
-    zero_tolerance: float = Field(
+    atol: float = Field(
         default=1e-10,
         description="Numerical tolerance below which values are considered zero. [env: QILISDK_ZERO_TOLERANCE]",
     )
+    rtol: float = Field(
+        default=1e-5,
+        description="Numerical relative tolerance below which values are considered zero. [env: QILISDK_RELATIVE_TOLERANCE]",
+    )
```
fedonman, line 76: The environment variables are by default `QILISDK_` + the name of the parameter, so `QILISDK_ATOL` and `QILISDK_RTOL`.

Lumorti: Ah, oops!

fedonman: You could also add constraints to the Field, e.g.
```
gt=0.0,
lt=1e-3
```

fedonman: And Pydantic will raise a validation error if a user tries to set e.g. a negative value.

fedonman (approved):
Image

src/qilisdk/settings.py
Lumorti: Okay, added! For now I left it as 0 <= atol, rtol <= 1, just in case someone wants to be really brutal setting it to 0.1 for example.

PR #115, opened 2025-12-10
[SDK-122] dense to dense()

Lumorti:
The property "dense" of QTensor has been changed to a method (i.e. now "QTensor.dense()"), since it could be expensive in the case of a large sparse object, and it being a property would make it seem like it's just accessing a cached dense object.


fedonman (approved):
Image

PR #118, opened 2025-12-10
[SDK-154] Updating Schedule Coefficients and Tlist

amirazzam2000:
Reworked analog schedule timing so `T` derives from coefficients or an explicit max time and `tlist` is generated from `dt`, with all backends/renderers consuming the same steps (and tests adjusted accordingly); the interpolator now validates overlapping intervals, keeps interval endpoints only, and is exported via `qilisdk.core`; QuTiP builds Hamiltonians via `to_qtensor(...).dense()` with correct dimensions


src/qilisdk/analog/schedule.py
```diff
@@ -145,25 +147,23 @@ def coefficients(self) -> dict[str, Interpolator]:
     @property
     def T(self) -> float:
         """Total annealing time of the schedule."""
-        return max(self.tlist)
+        if self._max_time is not None:
+            return self._get_value(self._max_time)
+        max_t = self._get_coefficients_max_time()
+        return self._get_value(max_t)
```
Lumorti, line 153: Should this also cache `max_t`? Or is `self._max_time` reserved for a user-specified explicit max time? 

```suggestion
        self._max_time = self._get_coefficients_max_time()
        return self._get_value(self._max_time)
```

src/qilisdk/core/interpolator.py
```diff
@@ -120,18 +120,36 @@ def __init__(
...
+                self._get_value(t) if not isinstance(t, tuple) else self._get_value(min(t, key=self._get_value))
+            ),
+        )
+
+        for i in range(len(fixed_times) - 1):
+            ti = fixed_times[i]
+            tj = fixed_times[i + 1]
+            t0 = self._get_value(ti) if not isinstance(ti, tuple) else self._get_value(ti[1])
+            t1 = self._get_value(tj) if not isinstance(tj, tuple) else self._get_value(tj[0])
+            if t0 == self._get_value(t1):
```
Lumorti: Is this checking for exact equality, or just that the time difference should be non-zero? If it's the latter, then maybe use the new atol setting:
```suggestion
            from qilisdk.settings import get_settings
            if abs(t0 - self._get_value(t1)) < get_settings().atol:
```
(obviously moving the import to the top of the file)

Lumorti: Also, why is `_get_value` needed here on t1? hasn't it already been `_get_value`'d in line 134?

src/qilisdk/core/interpolator.py
```diff
@@ -120,18 +120,36 @@ def __init__(
...
+                raise ValueError(f"The time point {self._get_value(t0)} is defined twice.")
+            if self._get_value(t0) > self._get_value(t1):
```
Lumorti: Same here, are the `_get_value`s needed?

tests/analog/test_schedule.py
```diff
@@ -116,7 +116,7 @@ def test_schedule_constructor_invalid_schedule_reference():
 def test_len_schedule(T, dt):
     """The length of a Schedule is T/dt (as an integer)."""
     sched = Schedule(dt=dt, total_time=T)
-    assert len(sched) == 2
+    assert len(sched) == len(sched.tlist)
```
Lumorti, line 121: (not related to this line, just the file)
Are there any tests for checking the backends with the schedules? If not, might be worth adding a tiny one for both (i.e. that the object inits correctly, not that the result is correct or anything like that).

uv.lock
```diff
@@ -2633,7 +2633,7 @@ wheels = [
 
 [[package]]
 name = "qilisdk"
-version = "0.1.7"
+version = "0.1.8"
```
Lumorti, line 2636: This shouldn't be there, no? We haven't had the 0.1.8 release yet

Lumorti (changes requested):
All seems fine to me, most of the comments are just suggestions, although the uv.lock version change I assume is an accident and should probably be fixed

uv.lock
amirazzam2000: So we have already updated the version tag in main, because this is the pre-release version. So this change here is inherited from main, and it should be fine.

Lumorti: Ah okay, got it, I didn't realise we changed the version number already, I thought it was only done when we create the release, my bad

src/qilisdk/analog/schedule.py
amirazzam2000: no in this case we shouldn't update the self._max_time because we want it to evaluate at each iteration. So basically the self._max_time will force the schedule to be of that time, whereas in this case we want to check what is the maximum time in the schedule.

tests/analog/test_schedule.py
amirazzam2000: There is a test for sure in the qutip backend, but I am not sure if cuda backend has one.

Lumorti (approved):
Nice!

PR #119, opened 2025-12-11
[SDK-156] Update TimeEvolution code blocks in the documentation to use the new schedule structure

amirazzam2000:
Updated TimeEvolution documentation examples (backends, functionals, cost_functions) to use the new Schedule signature with explicit driver/problem Hamiltonians, coefficient mappings, and linear interpolation import.


Lumorti (changes requested):
I see `dt=0.5` is used in the schedule, should that be using the `dt` defined above, or is it different?

Also, isn't `steps` unused now?

NataliaSirera (approved):

Lumorti (approved):
Nice!

PR #120, opened 2025-12-11
[SDK-161] Fix CUDA Integer Bug

Lumorti:
Fixed a bug in which the CUDA backend would throw an error if given a gate with an integer parameter, e.g. RX with theta=1.


amirazzam2000 (approved):
thanks 🙏

PR #124, opened 2025-12-16
[SDK-107] Fix SpeQtrum Optional Dependency

amirazzam2000:


src/qilisdk/speqtrum/speqtrum.py
```diff
@@ -257,12 +257,33 @@ def auth_flow(self, request: httpx.Request) -> Generator[httpx.Request, httpx.Re
...
+        Raises:
+            RuntimeError: If the user credentials are not present in the environment and were not provided in this constructor.
+
+        Note:
+            The resulting tokens are stored securely in the OS keyring so that future
+            :class:`SpeQtrum` constructions require no explicit credentials.
+        """
         logger.debug("Initializing SpeQtrum client")
+        if username is not None or apikey is not None:
+            self.login(username=username, apikey=apikey)
```
fedonman: If we do this, then we can no longer login passing no parameters and using the environment variables.

Image

src/qilisdk/speqtrum/speqtrum.py
```diff
@@ -282,9 +303,8 @@ def _create_client(self) -> httpx.Client:
             event_hooks={"response": [_ensure_ok]},
         )
 
-    @classmethod
     def login(
```
fedonman: This should be private method if the user now logins when instantiating.

fedonman (changes requested):

fedonman:
Left some comments.

Also, I am not sure if `login` should be made instance method after all. `logout` is still class method, and seems right for both to be of the same type.

There will also be the misconception that "login" is associated with the specific instance, whereas, it should be associated with the keyring.

amirazzam2000:
> Left some comments.
> 
> Also, I am not sure if `login` should be made instance method after all. `logout` is still class method, and seems right for both to be of the same type.
> 
> There will also be the misconception that "login" is associated with the specific instance, whereas, it should be associated with the keyring.

The reason for this change is that calling a class method wouldn't trigger the check to see if the optional module for speqtrum is loaded or not. This check is only triggered when the class is instantiated. And I think overall it's more intuitive to skip the log in method and rely on the constructor to log in and store the credentials.

fedonman:
> > Left some comments.
> > Also, I am not sure if `login` should be made instance method after all. `logout` is still class method, and seems right for both to be of the same type.
> > There will also be the misconception that "login" is associated with the specific instance, whereas, it should be associated with the keyring.
> 
> The reason for this change is that calling a class method wouldn't trigger the check to see if the optional module for speqtrum is loaded or not. This check is only triggered when the class is instantiated. And I think overall it's more intuitive to skip the log in method and rely on the constructor to log in and store the credentials.

The optional trigger is easily fixed, I will do a PR now.

image

I had the same misconception at some point we had discussed it, but login and logout should be distinct actions. It makes sense for them to be class methods because login is “global” in the sense that once you are logged in, multiple instances can share the login information (for example, through the keyring).

If login is moved to `__init__`, the user is forced to provide credentials every time and a request is sent to the server each time. This also removes any clear distinction between the user’s intent to log in and the intent to simply create an instance using an already logged-in user.

fedonman:
https://github.com/qilimanjaro-tech/qilisdk/pull/125

amirazzam2000:
okay then I will close this PR and we can go ahead with your approach :+1:

PR #125, opened 2025-12-18
[SDK-107] Fix Optional Dependencies

fedonman:
Optional dependency handling is now more robust: missing optional symbols are exported as attribute-aware stubs so both calling them and accessing members (e.g. `SpeQtrum.login()`) raises an informative `OptionalDependencyError` with the correct `pip install qilisdk[extra]` hint, and optional imports fall back to stubs even when a dependency is installed but fails during import.

The test suite no longer errors during collection when optional extras aren’t installed; optional-feature tests (CUDA, QuTiP, SpeQtrum, OpenFermion) are skipped via `pytest.importorskip(..., exc_type=ImportError)` unless their dependencies can be imported.

### DIFF
--- a/src/qilisdk/_optionals.py
+++ b/src/qilisdk/_optionals.py
@@ -70,16 +70,21 @@ class _OptionalDependencyStub:
     raise an informative OptionalDependencyError rather than AttributeError.
     """
 
-    def __init__(self, *, symbol_name: str, feature_name: str) -> None:
+    def __init__(self, *, symbol_name: str, feature_name: str, import_error: Exception | None = None) -> None:
         self._symbol_name = symbol_name
         self._feature_name = feature_name
+        self._import_error = import_error
         self.__name__ = symbol_name
         self.__qualname__ = symbol_name
 
     def _raise(self) -> NoReturn:
-        raise OptionalDependencyError(
+        message = (
             f"Using {self._symbol_name} requires installing the '{self._feature_name}' optional feature: `pip install qilisdk[{self._feature_name}]`\n"
         )
+        if self._import_error is None:
+            raise OptionalDependencyError(message)
+        detail = f"{type(self._import_error).__name__}: {self._import_error}"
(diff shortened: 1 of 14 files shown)


src/qilisdk/_optionals.py
```diff
@@ -63,6 +63,43 @@ class ImportedFeature:
...
+    """
+
+    def __init__(self, *, symbol_name: str, feature_name: str, import_error: Exception | None = None) -> None:
+        self._symbol_name = symbol_name
+        self._feature_name = feature_name
+        self._import_error = import_error
+        self.__name__ = symbol_name
+        self.__qualname__ = symbol_name
+
+    def _raise(self) -> NoReturn:
```
amirazzam2000, line 80: Why are we using NoReturn here instead of just None?

amirazzam2000 (changes requested):
I don't think these changes fixed the issue, because still if they user tries to call login on speqtrum while the optional dependency is not installed they will still not get an error. The only way to get the error is to instantiate speqtrum.

```
>>> from qilisdk.speqtrum import SpeQtrum
>>> SpeQtrum.login()
  File "<stdin>", line 1, in <module>
AttributeError: 'function' object has no attribute 'login'
>>> SpeQtrum
<function import_optional_dependencies.<locals>.make_stub.<locals>._stub at 0x7f3dbe106c00>
>>> SpeQtrum()
  File "<stdin>", line 1, in <module>
  File "/home/ameer/Documents/Qili/test_env/lib/python3.12/site-packages/qilisdk/_optionals.py", line 88, in _stub
    raise OptionalDependencyError(
qilisdk._optionals.OptionalDependencyError: Using SpeQtrum requires installing the 'speqtrum' optional feature: `pip install qilisdk[speqtrum]`
```

amirazzam2000 (approved):
Okay sorry after checking again there was an error on my side while testing the branch. The functionality seems to be working fine 🙏

PR #126, opened 2026-01-05
[SDK-160] QiliSim

Lumorti:
- Added a new simulation backend: QiliSim. This is an internal simulator written in C++. It offers improved performance compared to the QuTiP CPU simulator and can even be faster than the CUDA backend for some small test cases. Both circuit simulation and time evolution are supported. There are a number of time evolution methods: direct exponention, Anrolid/Krylov exponentiation, RK4 integration, and Monte Carlo (which can be combined with any of the previous methods).

- Changed the build system from hatchling to scikit-build, to allow for compilation of the C++ backend via support of CMake.

- Added method to generate a randomized circuit from a set of single- and two-qubit gates, providing a useful tool for benchmarking.


DZPM:
Luke pointed that `src/qilisdk/backends/qilisim.cpp` (C++ instead of Python) is not being analyzed by Sonarqube.

I tried to make it work, unsuccessfully. Everything obvious I checked looks ok:
- you can have python and c++ in the same repo
- our "Team" plan allows that
- our "Automatic analysis" allows that
- we have no exclusions, or .githignore, on cpp files

I'll go deeper, but it will take me a while :pray: so ignore it for now

Looks like it could be fixed by:
- github settings (I'd do a test PR, ignore it!) :crossed_fingers: 
- via a CI Analysis, instead of the automatic one (I'd prefer to avoid that)

docs/fundamentals/backends.rst
```diff
@@ -234,4 +243,56 @@ It is the most lightweight option, ideal for local development or environments w
         final_state=QTensor(shape=2x1, nnz=2, format='csr')
         [[0.05506547-0.00516502j]
         [0.3364973 -0.94005887j]]
-    )
\ No newline at end of file
+    )
+
+QiliSim Backend
```
amirazzam2000: We should probably mention that it's is built in and doesn't need additional installing like the other simulators

src/qilisdk/backends/qilisim.py
```diff
@@ -0,0 +1,182 @@
...
+            elif isinstance(obs, QTensor):
+                aux_obs = obs
+            else:
+                logger.error("Unsupported observable type {}", obs.__class__.__name__)
+                raise ValueError(f"unsupported observable type of {obs.__class__}")
+            if aux_obs is not None:
+                observables.append(aux_obs)
+
+        # Execute the time evolution
+        result = self.qili_sim.execute_time_evolution(
```
amirazzam2000, line 143: I think ideally this should be processed inside the cpp simulator, so like just pass all the python objects down and do the rest there. Maybe for now just add a TODO so that we can keep it in mind.

src/qilisdk/digital/circuit.py
```diff
@@ -177,3 +179,59 @@ def draw(self, style: CircuitStyle = CircuitStyle(), filepath: str | None = None
...
+        Generate a random quantum circuit from a given set of gates.
+
+        Args:
+            single_qubit_gates (set[Gate]): A set of single-qubit gate classes to choose from.
+            two_qubit_gates (set[Gate]): A set of two-qubit gate classes to choose from.
+            ngates (int): The number of gates to include in the circuit.
+
+        """
+        gate_list: list[type[BasicGate]] = list(single_qubit_gates) + list(two_qubit_gates)
+        for _ in range(ngates):
```
amirazzam2000, line 219: it might be good to check that we don't add the same gate twice because it will then result in identity

.github/workflows/tests.yml
```diff
@@ -42,8 +42,17 @@ jobs:
...
+        run: uv venv --clear
+
+      - name: Activate the environment
+        run: source .venv/bin/activate
+
+      - name: Build QiliSDK
+        run: uv build --wheel
+
+      - name: Install QiliSDK
+        run: uv pip install dist/*.whl
```
fedonman: Why do we need this change in this specific workflow?

changes/126.feature.1.md
```diff
@@ -0,0 +1 @@
+Added method to generate a randomized circuit from a set of single- and two-qubit gates, providing a useful tool for benchmarking.
```
fedonman: Be specific: which method is it? How is it used? Add a code example.

changes/126.feature.md
```diff
@@ -0,0 +1 @@
+Added a new simulation backend: QiliSim. This is an internal simulator written in C++. It offers improved performance compared to the QuTiP CPU simulator and can even be faster than the CUDA backend for some small test cases. Both circuit simulation and time evolution are supported. There are a number of time evolution methods: direct exponention, Anrolid/Krylov exponentiation, RK4 integration, and Monte Carlo (which can be combined with any of the previous methods).
```
fedonman, line 1: Add code examples for sampling and time evolution.

docs/fundamentals/backends.rst
```diff
@@ -2,10 +2,11 @@ Backends
 ========
 
 The :mod:`~qilisdk.backends` module provides concrete execution engines for running :mod:`~qilisdk.functionals` (quantum processes).  
-Currently, two backends are supported:
+Currently, three backends are supported:
 
 - :class:`~qilisdk.backends.cuda_backend.CudaBackend`  
 - :class:`~qilisdk.backends.qutip_backend.QutipBackend`  
+- :class:`~qilisdk.backends.qilisim.QiliSim`
```
fedonman: Add it first on the list 😉

docs/fundamentals/backends.rst
```diff
@@ -62,6 +63,10 @@ drivers to be present on the system.
      - ``qutip``
      - `QuTiP <https://qutip.org/>`_
      - CPU based; no special hardware needs.
+   * - :class:`QiliSim <qilisdk.backends.qilisim.QiliSim>`
```
fedonman: here too

docs/fundamentals/backends.rst
```diff
@@ -84,6 +89,10 @@ The table below summarizes which primitive :mod:`~qilisdk.functionals` each back
      - ✓
      - ✓
      - ✓
+   * - :class:`QiliSim <qilisdk.backends.qilisim.QiliSim>`
```
fedonman: here too

docs/fundamentals/backends.rst
```diff
@@ -234,4 +243,56 @@ It is the most lightweight option, ideal for local development or environments w
         final_state=QTensor(shape=2x1, nnz=2, format='csr')
         [[0.05506547-0.00516502j]
         [0.3364973 -0.94005887j]]
-    )
\ No newline at end of file
+    )
+
```
fedonman, line 300: Here too

docs/fundamentals/digital.rst
```diff
@@ -120,6 +120,22 @@ Quantum circuits can be built using the :class:`~qilisdk.digital.circuit.Circuit
...
 
+**Generating Random Circuits**
+
+You can also generate a random circuit with a specified number of gates using the :meth:`~qilisdk.digital.circuit.Circuit.randomize` method:
+
+.. code-block:: python
+
+    from qilisdk.digital import Circuit, H, X, CNOT
+
+    circuit = Circuit(nqubits=3)
```
fedonman: Since essentially `randomize()` is a constructor, it should be a class method that returns a random circuit:

```
circuit = Circuit.randomize(
    nqubits=3,
    single_qubit_gates={H, X},
    two_qubit_gates={CNOT},
    ngates=10
)
```

In this case, I would also rename it to just `random`: `Circuit.random(...)`

src/qilisdk/digital/circuit.py
```diff
@@ -177,3 +179,59 @@ def draw(self, style: CircuitStyle = CircuitStyle(), filepath: str | None = None
...
+            if gate_class.PARAMETER_NAMES:
+                for param_name in gate_class.PARAMETER_NAMES:
+                    val = random.uniform(-np.pi, np.pi)
+                    params[param_name] = Parameter(
+                        label=param_name + str(val), value=val, domain=Domain.REAL, bounds=(val, val)
+                    )
+
+            # If you're reading this and thinking "why don't we just do self.add(gate_class(*qubits, **params))"?
+            # It's because mypy gets angry about it, since then we don't know the type of the variable at runtime.
+            if gate_class == CNOT:
```
fedonman: Feel free to ignore mypy errors if it makes sense.

src/qilisdk/digital/circuit.py
```diff
@@ -177,3 +179,59 @@ def draw(self, style: CircuitStyle = CircuitStyle(), filepath: str | None = None
         renderer.plot()
         if filepath:
             renderer.save(filepath)
+
+    def randomize(
+        self, single_qubit_gates: set[type[BasicGate]], two_qubit_gates: set[type[BasicGate]], ngates: int
+    ) -> None:
```
fedonman: Convert to class method, see comment above.

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
     "loguru>=0.7.3",
     "matplotlib>=3.10.7",
     "mkl-service>=2.4.2; sys_platform != 'darwin'",
+    "pybind11>=3.0.1",
```
fedonman: Do we need this dependency here? I thought adding it to the build dependencies is enough.

src/qilisdk/backends/qilisim.cpp
```diff
@@ -0,0 +1,1695 @@
...
+            std::string: The gate name.
+        */
+        std::string control_string = "";
+        for (int i = 0; i < int(control_qubits.size()); ++i) {
+            control_string += "c";
+        }
+        return control_string + gate_type;
+    }
+
+    std::string get_id() const {
```
fedonman, line 269: What about parameterized gates? Wouldn't this implementation return the same id for e.g. two RX gates that have same target qubit, but different parameter theta?

src/qilisdk/backends/qilisim.cpp
```diff
@@ -0,0 +1,1695 @@
...
+                    mat_entries.emplace_back(Triplet(r, c, val));
+                }
+            }
+        }
+        SparseMatrix mat(dim, dim);
+        mat.setFromTriplets(mat_entries.begin(), mat_entries.end());
+        return mat;
+    }
+
+    SparseMatrix create_superoperator(const SparseMatrix& currentH, const std::vector<SparseMatrix>& jump_operators) {
```
fedonman, line 764: This was suggested from AI: Vectorization stacks columns but the Lindblad superoperator is built with Kronecker ordering for row-stacked vec, so direct/arnoldi with jumps will be mathematically wrong. qilisim.cpp (line 635), qilisim.cpp (line 682)

vectorize stacks columns (r + c*rows), i.e. column‑major vec (vec_col). For that convention, the identity is vec(AXB) = (B^T ⊗ A) vec(X), so the Hamiltonian part of the Liouvillian should be -i(I ⊗ H - H^T ⊗ I) and the jump term should be (L* ⊗ L) vec(ρ). In create_superoperator the code builds H ⊗ I and I ⊗ H^T, and uses L ⊗ L* (row‑vec ordering), which corresponds to vec_row, not vec_col. That means you’re applying a row‑vec Liouvillian to a column‑stacked vector, effectively evolving the transpose‑ordered system; for real symmetric matrices this can “look” correct, but for complex Hamiltonians or jumps it gives different dynamics. See qilisim.cpp (line 635), qilisim.cpp (line 682).
For reference:
```
vec_col(AXB) = (B^T ⊗ A) vec_col(X)
vec_row(AXB) = (A ⊗ B^T) vec_row(X)
```
Only the direct/arnoldi open‑system paths use this superoperator, so the bug is hidden in current tests.

src/qilisdk/backends/qilisim.cpp
```diff
@@ -0,0 +1,1695 @@
...
+            }
+        }
+
+        // Make sure probabilities sum to 1
+        if (std::abs(total_prob - 1.0) > atol_) {
+            throw py::value_error("Probabilities do not sum to 1 (sum = " + std::to_string(total_prob) + ")");
+        }
+
+        // Sample from these probabilities
+        std::map<std::string, int> counts = sample_from_probabilities(prob_entries, n_qubits, n_shots);
```
fedonman, line 1483: Measurement gates are ignored and sampling always returns the full `nqubits` bitstrings. This is different from sampling in QuTiP / CUDA / Qililab backends.

E.g for the following circuit

```
circuit = Circuit(2)
circuit.add(X(0))
circuit.add(X(1))
circuit.add(M(0))
```

results should be 

```
SamplingResult(
  nshots=1000,
  samples={'1': 1000}
)
```

but instead QiliSim returns:

```
SamplingResult(
  nshots=1000,
  samples={'11': 1000}
)
```

src/qilisdk/backends/qilisim.cpp
```diff
@@ -0,0 +1,1695 @@
...
+            num_threads = 1;
+        }
+        Eigen::setNbThreads(num_threads);
+
+        // Sanity checks
+        if (n_qubits <= 0) {
+            throw py::value_error("shots must be positive.");
+        }
+        if (n_shots <= 0) {
+            throw py::value_error("nqubits must be positive.");
```
fedonman: error messages are swapped.

src/qilisdk/backends/qilisim.cpp
```diff
@@ -0,0 +1,1695 @@
...
+                    }
+                    counts[binary_strings[state_index]]++;
+                    break;
+                }
+            }
+        }
+        return counts;
+    }
+
+    SparseMatrix get_vector_from_density_matrix(SparseMatrix& rho_t) {
```
fedonman, line 1013: Also AI: Pure-state extraction from a density matrix reads a row instead of a column (no conjugation), producing the conjugate state for complex amplitudes and changing unitary evolution/expectations.

Pure‑state extraction uses a row, not a column: In a pure density matrix ρ = |ψ><ψ|, any non‑zero column is proportional to |ψ>, while any non‑zero row is proportional to ⟨ψ| (i.e., ψ* as a column). get_vector_from_density_matrix picks a non‑zero diagonal entry and then copies row i into a column, returning |ψ*> instead of |ψ> when amplitudes are complex. Example: |ψ> = (|0> + i|1>)/√2 produces a row that normalizes to (|0> − i|1>)/√2. This conjugation flips phases and can change expectation values (e.g., sign flips for Y‑type observables). It’s triggered when a pure density matrix is converted into a statevector for efficiency. See qilisim.cpp (line 884), qilisim.cpp (line 1553).

src/qilisdk/backends/qilisim.cpp
```diff
@@ -0,0 +1,1695 @@
...
+
+        // Sanity checks
+        if (hamiltonians.size() == 0) {
+            throw py::value_error("At least one Hamiltonian must be provided");
+        }
+        if (step_list.size() == 0) {
+            throw py::value_error("At least one time step must be provided");
+        }
+        for (size_t h_ind = 0; h_ind < hamiltonians.size(); ++h_ind) {
+            if (parameters_list[h_ind].size() != step_list.size()) {
```
fedonman, line 1689: Are `hamiltonians` and `parameters_list` guaranteed to have same length? Otherwise, an index out of bounds error might occur.

fedonman (changes requested):
Great implementation Luke! :100: Made some comments.

src/qilisdk/backends/qilisim.py
```diff
@@ -0,0 +1,185 @@
...
+
+if TYPE_CHECKING:
+    from qilisdk.functionals.sampling import Sampling
+    from qilisdk.functionals.sampling_result import SamplingResult
+    from qilisdk.functionals.time_evolution import TimeEvolution
+    from qilisdk.functionals.time_evolution_result import TimeEvolutionResult
+    from qilisdk.noise_models.noise_model import NoiseModel
+
+
+class QiliSim(Backend):
```
amirazzam2000, line 33: Can we rename this to QiliBackend for consistency. And also expose it through the __init__

amirazzam2000 (changes requested):

src/qilisdk/backends/qilisim.py
fedonman: QiliRuntime contains a QiliBackend as well though.

fedonman: We can try to think an original name. I would leave it as it is for now.

.github/workflows/tests.yml
Lumorti: `uv build` is now required for the tests since it requires compiling the C++ backend, which as far as I could tell `uv sync` doesn't do, it just gets dependencies, so now we do a full build and install. For the code quality workflow this isn't needed since we don't need to actually run any code.

src/qilisdk/digital/circuit.py
Lumorti: Okay, good to know!

src/qilisdk/digital/circuit.py
Lumorti: good idea! added

src/qilisdk/backends/qilisim.py
amirazzam2000: Regarding this, I would suggest renaming this to `QiliBackend` and the one in QiliRuntime to `QililabBackend` that way we avoid confusion, and I would rather that the user facing library has consistent names.

src/qilisdk/digital/circuit.py
Lumorti: Done!

changes/126.feature.1.md
Lumorti: Got it!

.github/workflows/tests.yml
Lumorti: Actually, maybe uv sync does build, I will test this a bit more now

docs/fundamentals/digital.rst
Lumorti: Yes now that you say it, I also prefer this! Changed

docs/fundamentals/backends.rst
Lumorti: Ah yes, originally I left this because I wasn't sure whether it would be optional or not, but then forgot to update it once we agreed that not

docs/fundamentals/backends.rst
Lumorti: Done!

docs/fundamentals/backends.rst
Lumorti: Done!

docs/fundamentals/backends.rst
Lumorti: Done!

docs/fundamentals/backends.rst
Lumorti: Done!

.github/workflows/tests.yml
Lumorti: Actually yes it turns out sync *does* run the CMake, at some point in an earlier test it seemed like it wasn't but I guess I misunderstood! Reverted the file

fedonman: Yes, `uv sync` builds everything locally in edit mode (similar to `pip install -e .`), while `uv build` creates the files for distribution.

src/qilisdk/backends/qilisim.cpp
Lumorti: good catch

src/qilisdk/backends/qilisim.cpp
Lumorti: ah yes, oops

src/qilisdk/backends/qilisim.cpp
Lumorti: I'll add a check and a test

src/qilisdk/backends/qilisim.py
Lumorti: Yeah that's fair, for now it was just like this because it's copying exactly the qutip backend python for simplicity. Also, moving this to the c++ would require handling a few extra different qilisdk classes, and probably wouldn't be any more efficient, but I agree it would be nice to have as little as possible inside qilisim.py

src/qilisdk/backends/qilisim.cpp
Lumorti: Actually, right now it would be impossible for this to occur given the current python wrapper, since parameter_list comes from coeff and coeff is constructed with the same length as hamiltonians, but I'll put a check in the C++ just for safety.

changes/126.feature.md
Lumorti: Done!

src/qilisdk/backends/qilisim.cpp
Lumorti: QiliSim now correctly handles this, test added

src/qilisdk/backends/qilisim.cpp
Lumorti: very good that you noticed this

pyproject.toml
Lumorti: I think we want it there since we install it with pip, in the CMake it's just checking that it's present but I don't know if it would install it if it wasn't, I'll try removing it and see if it works

Lumorti: It does seem  to work find without, I'll remove it and see if the github actions build correctly

src/qilisdk/backends/qilisim.cpp
Lumorti: I have tried so many different tests with weird complex Hamiltonians and all of them pass. So maybe for now we leave this until we can find a reproducible example. Regardless I have made the above changes assuming that the AI is correct and it doesn't affect any of the new or old tests. Probably when we try to do noise models (and actually start testing jump operators) we will determine if there is a deeper problem.

src/qilisdk/digital/circuit.py
```diff
@@ -177,3 +179,105 @@ def draw(self, style: CircuitStyle = CircuitStyle(), filepath: str | None = None
         renderer.plot()
         if filepath:
             renderer.save(filepath)
+
+    @classmethod
+    def random(
+        cls, nqubits: int, single_qubit_gates: set[type[BasicGate]], two_qubit_gates: set[type[BasicGate]], ngates: int
+    ) -> "Circuit":
```
fedonman: Minor: Replace `"Circuit"` with `Self` from `typing` module.

src/qilisdk/digital/circuit.py
```diff
@@ -177,3 +179,105 @@ def draw(self, style: CircuitStyle = CircuitStyle(), filepath: str | None = None
...
+            elif gate_class == RZ:
+                new_circuit.add(RZ(qubits[0], **params))
+            elif gate_class == U1:
+                new_circuit.add(U1(qubits[0], **params))
+            elif gate_class == U2:
+                new_circuit.add(U2(qubits[0], **params))
+            elif gate_class == U3:
+                new_circuit.add(U3(qubits[0], **params))
+            else:
+                continue
```
fedonman: Can't we create the gates in a more generic way? I can see two issues with the if/elif/else chain:
1. You could easily forget to add a gate. In fact, the `CZ` is missing.
2. You can't possibly handle all combinations of `Controlled`, `Exponential` and `Adjoint` variations. E.g. I would like to do `Circuit.random(one_qubit_gates={Exponential[X]}, two_qubit_gates={Controlled[H]})`.

.github/workflows/code_quality.yml
```diff
@@ -53,3 +53,13 @@ jobs:
 
       - name: Mypy
         run: uv run mypy src --ignore-missing-imports
+
+      - name: Cpp Linter
+        uses: cpp-linter/cpp-linter-action@b6edc0625e3941baa1797f4b4326adeab6890c97
```
fedonman: Remove the hash and add back the version. I will add the `cpp-linter` to the allowed providers so SonarQube won't complain.

src/qilisdk/backends/qilisim.cpp
```diff
@@ -0,0 +1,1824 @@
...
+        }
+
+        // Make sure probabilities sum to 1
+        // const double probability_tolerance = 1e-5;
+        // if (std::abs(total_prob - 1.0) > probability_tolerance) {
+        if (std::abs(total_prob - 1.0) > atol_) {
+            std::stringstream ss;
+            ss << std::setprecision(10) << total_prob;
+            throw py::value_error("Probabilities do not sum to 1 (sum = " + ss.str() + ")");
+        }
```
fedonman, line 1480: This throws if I add any exponential gate so either it is not handled correctly by the simulator or our exponential definition is wrong?

Other backends don't support exponential either so it's not critical but let's check it.

In any case, if we don't support exponentials, we should catch the existence earlier and throw an `UnsupportedGateError` like other backends do.

```
circuit = Circuit(1)
circuit.add(X(0).exponential())
circuit.add(M(0))
```

gives

```
ValueError: Probabilities do not sum to 1 (sum = 3.762195691)
```

fedonman (commented):
Thanks for the quick fixes! Left some final polishing comments. Also, check the SonarQube comments (the 4 issues, I think you could address them, no worries about the duplicated code as it is exclusively in tests).

src/qilisdk/digital/circuit.py
fedonman: This is not critical. If you can't figure it out, just make sure all of our basic gates are present in the list.

fedonman: Actually, don't worry about that for now, I think it will be too complicated. Just add the missing gates and is good to go.

src/qilisdk/digital/circuit.py
Lumorti: Okay thanks, good to know

src/qilisdk/digital/circuit.py
Lumorti: Yeah I also hate this bit, originally I just did new_circuit.add(gate_class(...)) but then mypy didn't like this since then it isn't type safe. Maybe I should go back to that, if we just agree to ignore that mypy error

.github/workflows/code_quality.yml
Lumorti: Sure thing

src/qilisdk/digital/circuit.py
Lumorti: I have moved back to using `new_circuit.add(gate_class(*qubits, **params))  # type: ignore`, so now it is more general but we have to ignore the mypy error.

src/qilisdk/backends/qilisim.cpp
Lumorti: This is very curious, since we don't really "handle" the gates in QiliSim, we call their Gate._generate_matrix() from within the cpp, and then just use that matrix. I'll check now and add tests

Lumorti: Okay, so the problem with exponential gate is that they don't preserve normalization, so I will need to add renormalization between gate applications, which might slow normal circuits down. I will try to add this in a efficient way if I can.

fedonman: Then they aren't unitary? Should we change the `_generate_matrix()` directly?

fedonman (approved):
Great work!! This is a huge milestone 👏 💯

amirazzam2000 (approved):
Thanks a lot Luke!! 🎉🎉

PR #127, opened 2026-01-07
[SDK-131] Adding Noise Models

amirazzam2000:
Introduced noise model abstractions with digital, analog, and parameter noise classes.

Example:
```python
from qilisdk.noise_models import (
    DissipationNoise,
    KrausNoise,
    NoiseModel,
    ParameterNoise,
)
from qilisdk.digital import X
import numpy as np


op = QTensor(np.array([[1.0, 0.0], [0.0, 1.0]]))
model = NoiseModel()

model.add(KrausNoise(kraus_operators=[op], affected_qubits=[0], affected_gates=[X]))

# OR 
model.add(DissipationNoise(jump_operators=[op]))

# OR 
model.add(ParameterNoise(affected_parameters=["theta"], noise_std=0.05))
```


Lumorti (approved):
yup (for future reference: we already looked through this together in person)

Lumorti (changes requested):
we forgot to add the template of execute/usage of the noise model

Lumorti (changes requested):
could you also update the example in the docs? And also maybe a very small test

Lumorti (approved):
actually nevermind

PR #128, opened 2026-01-08
[SDK-185] Add Code Coverage

fedonman:
Added CI-driven code coverage reporting via Codecov. The Codecov configuration requires CI to pass, sets project coverage to auto with a 1% threshold, enforces 100% patch coverage with zero threshold, and only posts comments when changes occur. The test workflow now generates coverage XML and JUnit reports using pytest with coverage flags, and it uploads both coverage and test results to Codecov (only on Python 3.13 to avoid duplicate uploads). Finally, coverage and test report artifacts (`coverage.xml` and `junit.xml`) are added to `.gitignore` to keep them out of the repository.


DZPM (approved):
Great, thanks!

PR #129, opened 2026-01-09
[SDK-188] Improve YAML Tag Handling

fedonman:
YAML class registration now supports decorator usage with a `shared` option that auto-assigns shared tags as `{module}.{name}` instead of `{name}`.


src/qilisdk/yaml.py
```diff
@@ -187,8 +187,31 @@ def type_constructor(constructor, node):
...
+        if cls is None:
+
+            def decorator(target_cls):  # noqa: ANN202
+                return self.register_class(target_cls, shared=shared)
+
+            return decorator
+        if not getattr(cls, "yaml_tag", None):
+            cls.yaml_tag = f"!{cls.__module__.split('.')[0]}.{cls.__name__}" if shared else f"!{cls.__name__}"
+        return super().register_class(cls)
+
```
elygoner, line 212: With  the shared = True thing you are expecting new incorporations to the qilisdk / qililab teams to know or check what classes have the same names. This is a bad move that can cause problems in the future that won't be detected right away. It's a bad move, and one that we expicitly said we didn't want to do in the reunion we had.

elygoner (commented):
I thought I was supposed to do this, so I guess that was a waste of time. Anyway this pr uses a style of solution that we explicitly said we wouldn't use even if it was the easy option.

src/qilisdk/yaml.py
fedonman: It's the most painless compromise since shared names are few. (currently just 2-3)

Sadly, If we enforce the `{module}.{name}` naming everywhere, not only will we need to do many changes in the current runcards and calibration files (the error you got), but also in the near future (e.g. when `waveforms` module move to QiliSDK which will happen in February). This was not considered in the reunion.

elygoner: That's why I porposed adding explicit yaml_tag in the elements of the runcard/calibration_file. Also, why are we moving those to qilisdk and why hasn't any one told me???

elygoner: Honestly, I don't care. If you want to do something I was supposed to do and do it differently than what we agreed and make me waste my time, do it. But I won't be the one to implement the qililab part. I'm out.

PR #130, opened 2026-01-09
Qhc 1236 yaml serialisation bug

elygoner:


elygoner:
So, I had some problems with the implementation we proposed. It did serialize correctly, but it didn't build platform.
- I changed the automatic tag to do "!library.class" instead of module because I think i'ts prettier.
- With the solution we proposed, we had to use the same tag structure or specify a specific tag if we wanted to create a non generic to_yaml. I've implemented that. *IT IS IMPORTANT THAT ANY FUTURE to_yaml FUNCTION ADHERES TO THE TAG OR IT WONT WORK*
- In qililab, anything that goes on the runcard (waveforms) need a tag to allow the build_platform method to work.

PR #132, opened 2026-01-09
[SDK-186] adding doc-strings to Noise Models

amirazzam2000:
Added and aligned docstrings for noise model classes and backend execution methods, including noise model parameter documentation.

### DIFF
--- a/changes/132.doc.md
+++ b/changes/132.doc.md
@@ -1,0 +1,1 @@
+Added and aligned docstrings for noise model classes and backend execution methods, including noise model parameter documentation.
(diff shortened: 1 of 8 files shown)


Lumorti:
Ah, I also did this when doing the cuda noise models, especially since I had to tweak some existing classes very slightly and also add new noise models based on the base classes.

If you want, after we merge the cuda noise models to main, we can then update this PR and just use it to check that I didn't miss anything.

amirazzam2000:
This issue will be resolved in the CudaBackend noise model PR #131

PR #135, opened 2026-01-13
[HotFix] Fix uv.lock and ruff issues

fedonman:
Fixes uv.lock and ruff issues after corrupted dependencies update.


Lumorti (approved):
Nice!

PR #136, opened 2026-01-15
[SDK-182] Refactor Qilisim

Lumorti:
The Qilisim C++ code has been significantly refactored, improving compilation times, readability and performance.


src/qilisdk_cpp/backends/qilisim/qilisim.h
```diff
@@ -0,0 +1,128 @@
...
+
+    // time_evolution/direct.cpp
+    SparseMatrix iter_direct(const SparseMatrix& rho_0, double dt, const SparseMatrix& currentH, const std::vector<SparseMatrix>& jump_operators, bool is_unitary_on_statevector) const;
+
+   public:
+    // circuit_sampling/execute_sampling.cpp
+    py::object execute_sampling(const py::object& functional, const py::dict& solver_params);
+
+    // time_evolution/execute_time_evolution.cpp
+    py::object execute_time_evolution(const py::object& initial_state, const py::object& Hs, const py::object& coeffs, const py::object& steps, const py::object& observables, const py::object& jumps, bool store_intermediate_results, const py::dict& solver_params);
```
amirazzam2000: Can we make this as well accept a functional like in sampling? 
The reason I am asking for this is that it might be useful for the reservoirs that we have a public facing method (execute_time_evolution) that accepts functionals, and then does the translation to the C++ objects and then there is an internal method (internal_time_evolution) that takes these C++ objects and does the computation. Then the results are returned to execute_time_evolution that will wrap them again in python objects and returns them.

Same for the circuit sampling. 

This way when we build the reservoirs we will need to run both circuits and hamiltonian time evolutions, so the internal solvers can pass C++ objects to each other without having to pass through the python interface.

amirazzam2000 (approved):
Thank you luke!! it looks really nice 🙏

fedonman (approved):
Looks great! 💯 💯

PR #141, opened 2026-01-26
[SDK-199] Added density matrix support to qilisim for circuit simulation

Lumorti:
Added internal support for density matrix initial states when simulating circuits in QiliSim, as well as Monte-Carlo simulation.

### DIFF
--- a/src/qilisdk/backends/qilisim.py
+++ b/src/qilisdk/backends/qilisim.py
@@ -23,6 +23,7 @@ from qilisim_module import QiliSimCpp  # ty:ignore[unresolved-import]
 from qilisdk.backends.backend import Backend
 
 if TYPE_CHECKING:
+    from qilisdk.core import QTensor
     from qilisdk.functionals.sampling import Sampling
     from qilisdk.functionals.sampling_result import SamplingResult
     from qilisdk.functionals.time_evolution import TimeEvolution
@@ -108,7 +109,9 @@ class QiliSim(Backend):
             "atol": atol,
         }
 
-    def _execute_sampling(self, functional: Sampling, noise_model: NoiseModel | None = None) -> SamplingResult:
+    def _execute_sampling(
+        self, functional: Sampling, noise_model: NoiseModel | None = None, initial_state: QTensor | None = None
+    ) -> SamplingResult:
         """
         Execute a quantum circuit and return the measurement results.
 
@@ -120,7 +123,7 @@ class QiliSim(Backend):
 
         """
         logger.info("Executing Sampling with {} shots", functional.nshots)
-        result = self.qili_sim.execute_sampling(functional, noise_model, self.solver_params)
+        result = self.qili_sim.execute_sampling(functional, noise_model, initial_state, self.solver_params)
         logger.success("Sampling finished")
         return result
 
--- a/src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
+++ b/src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
@@ -23,14 +23,14 @@
 #include <omp.h>
 #endif
 
-void sampling(std::vector<Gate>& gates, 
+void sampling(const std::vector<Gate>& gates, 
               const std::vector<bool>& qubits_to_measure, 
               int n_qubits, 
               int n_shots, 
-              SparseMatrix& initial_state,
+              const SparseMatrix& initial_state,
               DenseMatrix& state,
               std::map<std::string, int>& counts, 
-              QiliSimConfig& config) {
+              const QiliSimConfig& config) {
     /*
     Execute a sampling functional using a simple statevector simulator.
 
@@ -54,9 +54,6 @@ void sampling(std::vector<Gate>& gates,
     */
 
     // Set the number of threads
-    if (config.num_threads <= 0) {
-        config.num_threads = 1;
-    }
 #if defined(_OPENMP)
     omp_set_num_threads(config.num_threads);
     Eigen::setNbThreads(config.num_threads);
--- a/src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
+++ b/src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
@@ -27,9 +27,10 @@ void sampling(std::vector<Gate>& gates,
               const std::vector<bool>& qubits_to_measure, 
               int n_qubits, 
               int n_shots, 
-              QiliSimConfig& config, 
               SparseMatrix& initial_state,
-              std::map<std::string, int>& counts) {
+              DenseMatrix& state,
+              std::map<std::string, int>& counts, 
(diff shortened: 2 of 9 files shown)


src/qilisdk/backends/qilisim.py
```diff
@@ -108,7 +109,9 @@ def __init__(
             "atol": atol,
         }
 
-    def _execute_sampling(self, functional: Sampling, noise_model: NoiseModel | None = None) -> SamplingResult:
+    def _execute_sampling(
```
amirazzam2000, line 112: can we do this in the C++ not the python? so the method `execute_sampling` (qilisim.cpp) passes a hardcoded initial state to `sampling` (from digital/sampling.cpp) and then from the python stand point there would be no visible changes

src/qilisdk_cpp/backends/qilisim/qilisim.cpp
```diff
@@ -83,10 +71,20 @@ py::object QiliSimCpp::execute_sampling(const py::object& functional, const py::
     std::vector<Gate> gates = parse_gates(functional.attr("circuit"), config.atol);
     std::vector<bool> qubits_to_measure = parse_measurements(functional.attr("circuit"));
 
+    // The initial state
+    long dim = 1L << n_qubits;
+    SparseMatrix initial_state_cpp;
+    if (initial_state.is_none()) {
+        initial_state_cpp = SparseMatrix(dim, 1);
```
amirazzam2000, line 78: so essentially we hard-code this if branch and remove the parameter passing

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -50,12 +56,29 @@ void sampling(std::vector<Gate>& gates, const std::vector<bool>& qubits_to_measu
...
 #endif
 
     // Start with the zero state
     long dim = 1L << n_qubits;
-    DenseMatrix state = DenseMatrix::Zero(dim, 1);
-    state(0, 0) = 1.0;
+    DenseMatrix state = initial_state;
+    bool is_statevector = (state.cols() == 1 && state.rows() == dim);
+
+    // If it's a density matrix, check if it's pure
```
amirazzam2000, line 68: keep in mind that even if the state is pure, if i provide a density matrix i would ideally like to have a density matrix as a result.

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -107,10 +139,25 @@ void sampling(std::vector<Gate>& gates, const std::vector<bool>& qubits_to_measu
         }
 
         // Apply the gate (Sparse-Dense multiplication, OpenMP parallel if enabled)
-        state = gate_matrix * state;
+        if (is_statevector || monte_carlo) {
+            state = gate_matrix * state;
```
amirazzam2000, line 144: can we offer a way to retrieve this state? even tho execute sampling doesn't use it

amirazzam2000 (changes requested):

src/qilisdk/backends/qilisim.py
Lumorti: This is an internal method so the front-end isn't affected, and it actually let's me test it

It's also in the C++, both at the sampling.cpp level and the qilisim.cpp level

src/qilisdk_cpp/backends/qilisim/qilisim.cpp
Lumorti: I think it's useful for us/theory to have the ability to test density matrix / initial state stuff from the Python level. Also, otherwise we have to implement a separate testing suite in C++ specifically for the C++ stuff

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
Lumorti: Yes, but we don't return a state in Sampling (I still think we should) so it's not needed unlike with Time Evolution

Edit: internally, sure

amirazzam2000: well that's what i was trying to say in the daily, is that it would be cool if the internal sampling method can return the state and take an initial state, both of which could be ignored by the execute_sampling method. Just for the future cuz it will be needed.

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
Lumorti: Sure

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
Lumorti: Yeah I added this, have a look

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -50,12 +59,30 @@ void sampling(std::vector<Gate>& gates, const std::vector<bool>& qubits_to_measu
...
 #if defined(_OPENMP)
     omp_set_num_threads(config.num_threads);
+    Eigen::setNbThreads(config.num_threads);
 #endif
 
     // Start with the zero state
     long dim = 1L << n_qubits;
-    DenseMatrix state = DenseMatrix::Zero(dim, 1);
-    state(0, 0) = 1.0;
+    state = initial_state;
```
amirazzam2000, line 64: do we need to implicitly copy the initial_state? because I am afraid if you are passing the pointer then you are updating the initial state alongside the state. And maybe it's better to just copy the information in initial_state. Or another option could be to not pass two parameters (state and initial_state) and just pass state, and then it will update that variable.

Lumorti: It's not a pointer, it's a reference. C++ references (defined using "&" rather than "*") can only be set on initialization, it's basically just "give a variable a different name". The following code will not change x:

```C++
int x = 3;
int& y = x;
int z = y;
z = 4;
std::cout << x << " " << y << " " << z << std::endl;
// prints: 3 3 4
```

C++ is nicer than Python, it's verrrry hard to accidentally change something you didn't mean to change, unless you are explicitly using pointers, then what you described could happen, that you set two pointers equal to each other and then changing the value stored at either pointer would affect the same data. But I generally avoid pointers, I think there are none in QiliSim so far. 

Regardless, I've made them const now, so the compiler will check to make sure nothing changes them even though they're passed by reference.

Lumorti: But in general any SparseMatrix/DenseMatrix should be passed by (constant) reference, to avoid making a copy and doubling the memory usage

amirazzam2000 (approved):
thanks a lot Luke!! 🙏

PR #142, opened 2026-01-28
[SDK-137] increase tests

Lumorti:
Increased test coverage of QiliSDK to 100% and fixed a few small bugs found in the process.


src/qilisdk/analog/hamiltonian.py
```diff
@@ -137,6 +137,8 @@ def __hash__(self) -> int:
         return hash((self._NAME, self._qubit))
 
     def __eq__(self, other: object) -> bool:
+        if isinstance(other, Hamiltonian):
+            return other == self
```
Lumorti, line 157: Added because Pauli == Hamiltonian wasn't defined, but Hamiltonian == Pauli was defined, so now the former calls the latter

src/qilisdk/analog/hamiltonian.py
```diff
@@ -416,6 +418,8 @@ def __eq__(self, other: object) -> bool:
             )
         if isinstance(other, PauliOperator):
             other = other.to_hamiltonian()
+        if isinstance(other, QTensor):
+            other = Hamiltonian.from_qtensor(other)
```
Lumorti, line 474: Hamiltonian == QTensor wasn't defined (but was for other similar objects)

src/qilisdk/analog/hamiltonian.py
```diff
@@ -643,14 +647,13 @@ def parse(cls, hamiltonian_str: str) -> Hamiltonian:
...
-                if first.isdigit() or first in {"(", ".", "+", "-"}:
-                    return True
+                # Make sure it's not empty
+                if text:
+                    # If the first char is digit, '(', '.', '+', '-', or '0',
+                    # or if 'j' is present, assume it's numeric
+                    first = text[0]
+                    if first.isdigit() or first in {"(", ".", "+", "-"}:
+                        return True
                 return "j" in text
```
Lumorti, line 709: Rewrite because otherwise the first check was unreachable, if text is empty then `return j in text` returns False anyway

src/qilisdk/analog/hamiltonian.py
```diff
@@ -758,12 +761,14 @@ def trace(self) -> Number:
...
+        d = 2**n
         t = self._elements.get((PauliI(0),), 0)
         if isinstance(t, Parameter):
-            return t.evaluate()
+            return t.evaluate() * d
         if isinstance(t, Term):
-            return t.evaluate({})
-        return t
+            return t.evaluate({}) * d
+        return t * d
```
Lumorti, line 823: Trace of a Hamiltonian in Pauli form should be d * coeff_of_identity, not just coeff_of_identity. For example, The Hamiltonian H = I(0) should have a trace of 2, not 1.

src/qilisdk/core/interpolator.py
```diff
@@ -178,6 +181,39 @@ def _generate_tlist(self) -> list[PARAMETERIZED_NUMBER]:
...
+
+                # Compute the time scale
+                self._time_scale_cache = self._get_value(self._max_time) / max_t
+
+            # Otherwise set the scale to 1.0
+            else:
+                self._time_scale_cache = 1.0
+
+        return self._time_scale_cache
+
```
Lumorti, line 216: Time scale caching is now handled in a single place, previously there was a lot of code repetition of checking/recalculating the cache, so now it's centralized

src/qilisdk/core/interpolator.py
```diff
@@ -552,24 +566,32 @@ def _linear_value(
...
             return v1 * alpha + v0 * (1 - alpha)
 
-        if prev_expr is None and next_expr is not None:
+        # this is done in order to prevent setting the indices to none and causing type errors
+        has_prev = insert_pos >= 1
+        has_next = insert_pos < len(self._tlist)
+        prev_idx = self._tlist[insert_pos - 1] if has_prev else 0
+        prev_expr = self._time_dict[prev_idx] if has_prev else 0
+        next_idx = self._tlist[insert_pos] if has_next else 0
+        next_expr = self._time_dict[next_idx] if has_next else 0
```
Lumorti, line 573: As mentioned in the comment, previous these vars could be either None or a float/int/Term, but ty was giving errors since it wasn't figuring out that there was no way that it could reach the end of the function with prev_idx being None. So now the ids/expr are never None and we have separate vars to determine whether they have been set or not

src/qilisdk/core/variables.py
```diff
@@ -1248,11 +1248,6 @@ def __eq__(self, other: object) -> bool:
             return self.value == other
         return False
 
-    def __ne__(self, other: object) -> bool:
-        if isinstance(other, (float, int)):
-            return self.value != other
-        return NotImplemented
-
```
Lumorti, line 1255: I check around online, apparently explicitly adding a __ne__ isn't needed unless you want to do something special, by default the __ne__ is the negation of the __eq__

src/qilisdk/digital/circuit_transpiler_passes/decompose_multi_controlled_gates_pass.py
```diff
@@ -143,7 +143,7 @@ def _sqrt_of(gate: BasicGate) -> BasicGate:
         return RZ(q, phi=math.pi / 8.0)
 
     # Build the 2x2 unitary matrix for gate
-    if isinstance(gate, (U2, U3, H, BasicGate)):
+    if isinstance(gate, BasicGate) and gate.nqubits == 1:
```
Lumorti, line 146: This was just making consistent with a different check elsewhere in the file, also since U3, U2 and H extend BasicGate anyway it's redundant, no?

src/qilisdk/digital/ansatz.py
```diff
@@ -294,8 +294,8 @@ def __init__(
 
         nqubits = problem_hamiltonian.nqubits
 
-        if layers <= 0:
-            raise ValueError("layers must be >= 1")
+        if layers < 0:
+            raise ValueError("layers must be >= 0")
```
Lumorti, line 298: This was making it consistent with the other ansatz, but Ameer and I did chat about whether the minimum number of layers should be zero or one

src/qilisdk/digital/gates.py
```diff
@@ -250,9 +250,7 @@ def __init__(
             # Check that anything mentioned in parameter_transforms is also in parameters
             for param in self._parameter_transforms[term].variables():
                 if param.label not in self._parameters:
-                    raise ValueError(
-                        f"Parameter transform '{term}' contains parameter '{param.label}' which is not defined in the parameters list."
-                    )
+                    raise InvalidParameterNameError
```
Lumorti, line 258: This was used in some places and not other, made it consistent across the file

src/qilisdk/digital/gates.py
```diff
@@ -413,17 +409,21 @@ def __init__(self, *control_qubits: int, basic_gate: TBasicGate) -> None:
...
         P = np.array([[0, 0], [0, 1]], dtype=complex)
-        for _ in range(len(self.control_qubits) - 1):
+        for _ in range(len(self.control_qubits) - len(self.basic_gate.control_qubits) - 1):
             P = np.kron(P, np.array([[0, 0], [0, 1]], dtype=complex))
         # Extend the projector to the full space (control qubits ⊗ target qubit). It acts as P_control ⊗ I_target.
-        I_target = np.eye(1 << len(self.target_qubits), dtype=complex)
+        I_target = np.eye(1 << self.basic_gate.nqubits, dtype=complex)
         # The controlled gate is then:
         controlled = I_full + np.kron(P, self.basic_gate.matrix - I_target)
         return controlled
```
Lumorti, line 434: This now correctly generates the matrices for controlled CNOTs, which in the code are never explicitly forbidden. Basically I_target should be the same size as the basic gate matrix, not the number of target qubits (e.g. CNOT has a 4x4 base matrix, but only one target qubit)

src/qilisdk/utils/openqasm2.py
```diff
@@ -154,7 +154,7 @@ def from_qasm2(qasm_str: str) -> Circuit:
         #   Group 1: gate name (e.g., "h", "rx", "cx")
         #   Group 2: optional parameters (inside parentheses)
         #   Group 3: operand list (e.g., "q[0]" or "q[0], q[1]")
-        m = re.match(r"^(\w+)(?:\(([^)]*)\))?\s+(.+);$", line)
+        m = re.match(r"^(\w+)(?:\(([^)]*)\))?\s+(.+);", line)
```
Lumorti, line 157: This previous wouldn't match lines with comments at the end, e.g. 
`h q[0]; // this is a comment`

fedonman (approved):
Amazing work!!! Thanks a lot Luke. 💯 🙏

src/qilisdk/core/interpolator.py
```diff
@@ -387,13 +417,9 @@ def add_time_point(
...
-        if self._max_time is not None and self._tlist is not None:
-            if self._time_scale_cache is None:
-                self._time_scale_cache = self._get_value(self._max_time) / self._get_value(
-                    max(self._tlist, key=self._get_value)
-                )
-            time /= self._time_scale_cache
+            raise ValueError(
+                "Coefficient must be a number, Parameter, Term, or callable that returns one of these types."
+            )
         self._time_dict[time] = coeff
```
amirazzam2000: the code that has been removed was applying the time scaling on time but now it's not being applied anymore no?

src/qilisdk/core/variables.py
```diff
@@ -1635,10 +1630,6 @@ def __repr__(self) -> str:
                     )
             elif isinstance(e, BaseVariable):
                 if self._is_constant(e):
-                    if self.operation in {Operation.ADD, Operation.SUB} and self[e] == 0:
```
amirazzam2000, line 1638: why did we remove these lines? these prevent printing + 0 or * 1

src/qilisdk/core/interpolator.py
Lumorti: Ah yes, very good catch!! Don't know how that happened, also scary that wasn't caught in a test

src/qilisdk/core/variables.py
Lumorti: These were completely unreachable, the above code (lines 161-1620) already removes the constant if it's 0 and ADD/SUB or 1 and MUL/DIV

amirazzam2000: alright then if you checked that they are unreachable we can remove them!

src/qilisdk/core/interpolator.py
amirazzam2000: can we add a test for this method and check again that this time scaling is okay everywhere? I will do some tests on my end because this part of the code is a bit tricky and has many patches from errors encountered during testing.

Lumorti: Ah wait no, I remember, I think this was causing unexpected behaviour. 

Elsewhere in the code, the original time_dict the user gave is treated as immutable, like if they gave {0: 0: 10: 1} then even if they later set max time to be 5, the time_dict would still be the same. Time_dict is always treated as the "raw" time values, such that then when tlist is generated it is generated/scaled from the raw times. If we scale when we add a point, then when we later scale whilst generating the tlist we will scale twice.

Imagine I init with the dict {0: 0, 10: 1}, then set max time to 5, then add the point {15: 2}. 
Shouldn't my new time_dict be {0: 0, 10: 1, 15: 2}?
Because with the original code it would be {0: 0, 7.5: 2, 10:1}.

amirazzam2000: okay then there might have been an uncaught error in the original code but the expected behaviour is the following: 

- user creates dict {0: 0, 10: 1}
- user sets max time to 5
- user adds point {2.5: 0.5}
- then the original time dict should be {0: 0, 5: 0.5, 10: 1}

The thing is that it's more intuitive for the user to add points in their new time scale than having to refer to the original time scale at creation.

amirazzam2000: for this I think we just need to change this line to:
`  self._time_dict[time / self._time_scale] = coeff`

Lumorti: I changed it in one of the commits! Have a look, also there's the test test_add_time_point_with_scaling which passes, so have a look at that and make sure that's the expected behaviour

amirazzam2000: well I assume that with the new implementation of time_scale that we only need the line that says `time /= self._time_scale` because internally it will do all the computation you do up there, otherwise return 1

Lumorti: So the _time_scale I think relies on having an up-to-date t_list, everywhere _time_scale is called there is a check for t_list I think

amirazzam2000: but in _time_scale you do: 
```
  if self._time_scale_cache is None:
        # Make sure we have a max time set
        if self._max_time is not None:
            # Generate the tlist if we haven't already
            if self._tlist is None:
                self._tlist = self._generate_tlist()

```
which is essentially duplicate with the code you have between lines 423 - 425.

Lumorti: Ah true, I forgot I included it in _time_scale. Fixed in the latest commit

amirazzam2000 (approved):
perfect! Amazing work Luke 🙌 Thanks!

PR #143, opened 2026-02-02
[SDK-201] initial reservoir proposal

amirazzam2000:
Added quantum reservoir computing support through the new`QuantumReservoir`, `ReservoirLayer`, and `ReservoirInput` APIs. Backends can now execute reservoir workflows with per-layer inputs, optional pre/post encoding stages, and optional intermediate/final state output.

Example:

```python
from qilisdk.analog import Schedule, Z
from qilisdk.backends import QutipBackend
from qilisdk.core import ket
from qilisdk.digital import Circuit, RX

theta = ReservoirInput("theta", 0.0)
pre = Circuit(1)
pre.add(RX(0, theta=theta))

layer = ReservoirLayer(
    evolution_dynamics=Schedule(
        hamiltonians={"h": Z(0)},
        coefficients={"h": {(0.0, 1.0): 1.0}},
        dt=0.1,
    ),
    observables=[Z(0)],
    input_encoding=pre,
)

reservoir = QuantumReservoir(
    initial_state=ket(0),
    reservoir_layer=layer,
    input_per_layer=[{"theta": 0.1}, {"theta": 0.4}],
    store_intermediate_states=True,
)

result = QutipBackend().execute(reservoir)
print(result.expected_values)
```


src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,204 @@
...
+
+
+class QuantumReservoir:
+    def __init__(
+        self,
+        intial_state: QTensor,
+        reservoir_pass: ReservoirPass,
+        input_per_layer: list[dict[str, float]],
+        store_final_state: bool = False,
+        store_intermideate_states: bool = False,
```
DZPM:
```suggestion
        store_intermediate_states: bool = False,
```

DZPM:
```suggestion
        store_intermediate_states: bool = False,
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,204 @@
...
+        nshots: int = 0,
+    ) -> None:
+        self._intial_state = intial_state
+        self._reservoir_pass = reservoir_pass
+        self._input_per_layer = input_per_layer
+        self._store_final_state = store_final_state
+        self._store_intermideate_states = store_intermideate_states
```
DZPM:
```suggestion
        self._store_intermediate_states = store_intermediate_states
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,204 @@
...
+    @property
+    def input_per_layer(self) -> list[dict[str, float]]:
+        return self._input_per_layer
+
+    @property
+    def store_final_state(self) -> bool:
+        return self._store_final_state
+
+    @property
+    def store_intermideate_states(self) -> bool:
```
DZPM:
```suggestion
    def store_intermediate_states(self) -> bool:
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,204 @@
...
+        return self._store_intermideate_states
```
DZPM:
```suggestion
        return self._store_intermediate_states
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+        super().__init__(
+            label=label,
+            value=value,
+            domain=domain,
+            bounds=bounds,
+            trainable=False,
+        )
+
+
+class ReservoirPass(Parameterizable):
```
fedonman: `Pass` is a bit confusing for me since I instantly associate it with a chain of events (passes) like within a compiler. Whereas we always repeat the same. Maybe `ReservoirStep` or `ReservoirLayer` or just `Reservoir`?

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+    """Single reservoir layer template.
+
+    A reservoir pass can contain up to three ordered stages:
+    ``pre_processing -> reservoir_dynamics -> post_processing``.
+    It also stores the observables measured after each pass and optional qubits to reset.
+    """
+
+    def __init__(
+        self,
+        reservoir_dynamics: Schedule,
```
fedonman: `reservoir` is redundant. Could rename to just `dynamics`
```suggestion
        dynamics: Schedule,
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+        measured_observables: list[QTensor | Hamiltonian | PauliOperator],
```
fedonman: `measured` is reduntant. We should rename to `observables` same as TimeEvolution. Also,
1. why do we allow `QTensor` here, whereas in other functionals we do not?
2. why do we allow PauliOperators? User never interacts with that class, since e.g. `X(0)` is converted to `Hamiltonian` directly. 
```suggestion
        observables: list[QTensor | Hamiltonian | PauliOperator],
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+        post_processing: Circuit | None = None,
+        pre_processing: Circuit | None = None,
```
fedonman: These are nice namings, just an idea: If we want to be more close to ML terminology these could be renamed to `input_encoding` and `output_encoding`? I saw the literature in Notion is using these terms.

In any case, they should be in order, `pre_processing` before `post_processing`.
```suggestion
        pre_processing: Circuit | None = None,
        post_processing: Circuit | None = None,
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+            post_processing: Optional single-qubit post-processing circuit.
+            pre_processing: Optional single-qubit pre-processing circuit.
+            qubits_to_reset: Optional qubits reset between consecutive layers.
+        """
+
+        self._pre_processing: Circuit | None = pre_processing
+        self._reservoir_dynamics: Schedule = reservoir_dynamics
+        self._post_processing: Circuit | None = post_processing
+        self._qubits_to_reset: list[int] | None = qubits_to_reset
+        super().__init__()
```
fedonman, line 90: move `super().__init__()` to the first line.

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+
+        for observable in measured_observables:
+            if isinstance(observable, PauliOperator):
+                self._qtensor_observables.append(observable.to_hamiltonian().to_qtensor(self._nqubits))
+            elif isinstance(observable, Hamiltonian):
+                self._qtensor_observables.append(observable.to_qtensor(self._nqubits))
+            elif isinstance(observable, QTensor):
+                self._qtensor_observables.append(self._process_qtensor(observable))
+
+        self._measured_observables = measured_observables
```
fedonman: Following the previous question, isn't a bit weird to store duplicate list of observables just with a different representation? We could store a single type (e.g. `Hamiltonian`) and the user could call `.to_qtensor()` or the other way around. Also, as mentioned, I think `PauliOperator` is not used by users.

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+            self._pre_processing.set_parameters(
+                {k: v for k, v in parameters.items() if k in self._pre_processing.get_parameter_names()}
+            )
+        self._reservoir_dynamics.set_parameters(
+            {k: v for k, v in parameters.items() if k in self._reservoir_dynamics.get_parameter_names()}
+        )
+        if self._post_processing:
+            self._post_processing.set_parameters(
+                {k: v for k, v in parameters.items() if k in self._post_processing.get_parameter_names()}
+            )
```
fedonman: How do we ensure that the parameters of the pre-processing circuit are distinct from those of the post-processing circuit?

For example, if both circuits have an `RX` gate as their first parameterizable gate, the corresponding parameters will share the same name. How can I modify the value of only one of them without affecting the other?

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+    @property
+    def reservoir_dynamics(self) -> Schedule:
+        """Main dynamics stage for each pass."""
+        return self._reservoir_dynamics
+
+    def get_parameter_values(self) -> list[float]:
+        """Return current parameter values of the pre-processing and dynamics blocks."""
+        return (
+            self._pre_processing.get_parameter_values() if self._pre_processing else []
+        ) + self._reservoir_dynamics.get_parameter_values()
```
fedonman: We should include `post_processing` circuit here too.

src/qilisdk/functionals/quantum_reservoirs.py
fedonman: This is not fixed.

src/qilisdk/functionals/quantum_reservoirs.py
fedonman: This is not fixed.

src/qilisdk/core/parameterizable.py
```diff
@@ -43,10 +43,18 @@ def get_parameter_names(self) -> list[str]:
...
+        return [k for k, param in self._parameters.items() if param.is_trainable]
+
     def get_parameters(self) -> dict[str, RealNumber]:
         """Return a mapping from parameter labels to their current numerical values."""
         return {label: param.value for label, param in self._parameters.items()}
 
+    def get_trainable_parameters(self) -> dict[str, RealNumber]:
+        """Return a mapping from parameter labels to their current numerical values."""
+        return {label: param.value for label, param in self._parameters.items() if param.is_trainable}
+
```
fedonman: I do not particularly like adding these explicit methods.

1. For consistency, we would also need to add the corresponding `set_*` versions.
2. If we begin specializing methods in this way, then every time we introduce a new property to `Parameter`, we would need to add matching methods. For example, if I add a `fixed: bool` property to indicate that a parameter cannot change value, would we then need to introduce `get_fixed_parameters()`, `get_fixed_parameter_values()`, and so on?
3. This approach significantly increases the number of public methods, and these would need to be implemented in **every** class that implements `Parameterizable`.

If we want to support filtering based on properties, then either:

1. The user can filter the returned list themselves (which is straightforward), or preferably:
2. We can provide a `where: Callable[[Parameter], bool] | None = None` argument to the standard methods, similar to what we implemented in the `list_devices` and `list_jobs` methods of the `SpeQtrum` client. For example:

`params = circuit.get_parameters(lambda x: x.is_trainable)`

src/qilisdk/core/variables.py
```diff
@@ -1160,18 +1161,28 @@ def __init__(
...
             )
         self._value = value
-        self.set_bounds(bounds[0], bounds[1])
+        self._trainable = trainable
+        if not trainable:
+            if bounds[0] is not None or bounds[1] is not None:
+                logger.warning(f"Ignoring the bounds of the un-trainable parameter {label}")
+            self.set_bounds(None, None)
+        else:
+            self.set_bounds(bounds[0], bounds[1])
```
fedonman: `is_trainable = False` means an Optimizer won't change the value of the parameter, but the user can still change it right? And we should respect the provided bounds.

src/qilisdk/backends/cuda_backend.py
```diff
@@ -627,6 +631,45 @@ def _execute_time_evolution(self, functional: TimeEvolution) -> TimeEvolutionRes
             intermediate_states=intermediate_states,
         )
 
+    def _execute_quantum_reservoir(self, functional: QuantumReservoir) -> QuantumReservoirResult:
+        state = copy(functional.initial_state).to_density_matrix()
```
fedonman: Why do we need `copy` here? `.to_density_matrix()` returns a new object.

src/qilisdk/backends/qutip_backend.py
```diff
@@ -107,6 +113,45 @@ def __init__(self, nsteps: int = 10_000) -> None:
         }
         logger.success("QutipBackend initialised")
 
+    def _execute_quantum_reservoir(self, functional: QuantumReservoir) -> QuantumReservoirResult:
+        state = copy(functional.initial_state).to_density_matrix()
```
fedonman: same here, is `copy` needed?

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+            trainable=False,
+        )
+
+
+class ReservoirPass(Parameterizable):
+    """Single reservoir layer template.
+
+    A reservoir pass can contain up to three ordered stages:
+    ``pre_processing -> reservoir_dynamics -> post_processing``.
+    It also stores the observables measured after each pass and optional qubits to reset.
```
fedonman: Both `pass` and `layer` are mentioned here.

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,404 @@
...
+
+    Each element in ``input_per_layer`` is applied to the underlying
+    :class:`ReservoirPass` before executing one pass of dynamics and measurement.
+    """
+
+    def __init__(
+        self,
+        initial_state: QTensor,
+        reservoir_pass: ReservoirPass,
+        input_per_pass: list[dict[str, float]],
```
fedonman: We should also find a better name for this, it is confusing as it is now since it doesn't contain `parameters`. Maybe `input_parameters_per_step` or `input_parameters_per_layer` or just `input_parameters` or `parameter_assignments`?

fedonman (changes requested):
Great job! I left several comments, some related to naming. Since this is a novel representation, we should aim for maximum clarity. Other comments concern specific implementation details.

src/qilisdk/functionals/quantum_reservoirs.py
amirazzam2000: so I added this here to improve performance in later steps, because usually you will be measuring these observables at each reservoir step (and we have like 1000-4000 steps normally). So it's good to cache these values ahead of time.

as per the PauliOperators, they were supported everywhere before, the doc-strings might have not been updated (I changed them now). But we do offer them in the library and it feels odd not to support them. Like yes users usually use the hamiltonian version, but if a user uses the Pauli version there is nothing wrong with that.

src/qilisdk/core/parameterizable.py
amirazzam2000: the structure of the parameterizable has been completely changed. I didn't add a general filter i only added a trainable toggle for now.

src/qilisdk/core/variables.py
amirazzam2000: yes so trainable = false that means the optimizer won't touch this parameter. regarding the bounds I think if the user uses it as an input, i wouldn't see why the parameters would matter. Because the user puts in whatever value they want. the bounds were more for the training to restrict the range of the parameters.

src/qilisdk/functionals/quantum_reservoirs.py
amirazzam2000: I changed it to input_per_layer because i think the users are not gonna use Parameter directly they will use the alias ReservoirInput. so I think using the word input would be consistent with the interface

src/qilisdk/functionals/quantum_reservoirs.py
amirazzam2000: this is a good catch i will look into it!! thanks :pray:

src/qilisdk/core/variables.py
fedonman: If the user does:

```
from qilisdk.core import Parameter

parameter = Parameter(label="scale", value=0.0, bounds=(0.0, 1.0), trainable=False)
parameter.set_value(5)
```

it is expected that the `set_value` to raise an error, regardless if the parameter is trainable or not.

amirazzam2000: okay I have an idea to fix it! I will apply it now

src/qilisdk/functionals/quantum_reservoirs.py
fedonman: If these refer to `ReservoirInput` then we should name it `inputs_per_layer` (plural)

But, my main question is: What happens if the user uses normal parameters? E.g. What happens if I give the following circuit as `pre_proprocessing`?

```
pre_processing = Circuit(2)
pre_processing.add(RX(0, theta=0.5))
pre_processing.add(RX(1, theta=0.5))
```

instead of

```
pre_processing = Circuit(2)
pre_processing.add(RX(0, theta=ReservoirInput("theta_0", 0.5)))
pre_processing.add(RX(1, theta=ReservoirInput("theta_1", 0.5)))
```

This will be the most natural way of the user to define a circuit since he is used to it.

Additionally, what happens if I mix and match parameters? E.g.

```
pre_processing = Circuit(2)
pre_processing.add(RX(0, theta=0.5))
pre_processing.add(RX(1, theta=ReservoirInput("theta_1", 0.5)))
```

amirazzam2000: so you can mix and match and it should be fine, the difference is that your parameter unless defined as untrainable will not be used as an input to the reservoir. And I think the user should be very aware of what they are using as reservoir input because they should have planned it before started coding. So it shouldn't be accidental. But basically any untrainable parameter in your program will be taken as a reservoir input. You don't have to use it as a reservoir input it can also be used as a fixed value, but you will have the option to set it in the inputs_per_layer.

fedonman: So, from my understanding, I can have N parameters in a circuit, and choose M < N of them to be `ReservoirInput` and only these will be used in the reservoir loop. Is this correct? What happens to the rest? Will stay the same throughout the whole execution?

> so you can mix and match and it should be fine

We should verify this with tests.

> But basically any untrainable parameter in your program will be taken as a reservoir input.

This defeats the purpose of being intentional by the user. IMO we should enforce by checking `is_instance(parameter, ReservoirInput)` instead of checking `parameter.is_trainable == True`.

The fact that `ReservoirInput` is not trainable, does not mean that all non-trainable parameters can be `ReservoirInput`.

fedonman: Side note. Why `trainable` matters here? I cannot see any optimizer or variational program involved.

src/qilisdk/functionals/quantum_reservoirs.py
fedonman: Users shouldn't have to use the class PauliOperator directly though. They should always use them through the Hamiltonian class. It's the same analogy that we don't accept `Gate | Circuit` but just `Circuit`, even if a user can create a gate individually.

fedonman: I think the `PauliOperator | Hamiltonian` is a remnant from the days when we didn't convert the X/Y/Z/I methods directly to Hamiltonian. If it's left in other places too, it's by mistake. We should remove all such occurrences. (we can do it in a separate PR to speed up the merge of this one, that's fair)

src/qilisdk/functionals/quantum_reservoirs.py
amirazzam2000: >  Will stay the same throughout the whole execution?

yes, the idea is that you can externally train them with a variational program. But we need to modify the cost function a bit because the reservoirs work a bit differently and the output is not only at the end but also at each iteration. 

> The fact that ReservoirInput is not trainable, does not mean that all non-trainable parameters can be ReservoirInput.

Yeah I see your point, I will change it. 

> Side note. Why trainable matters here? I cannot see any optimizer or variational program involved.

I think this is something we have been discussing adding for a while, but in this case I wanted to reuse the infrastructure that we have built for parameter but I wanted it to not be trainable in case the reservoir is passed through a variational program.

src/qilisdk/functionals/quantum_reservoirs.py
amirazzam2000: okay you convinced me, but we should do it in a separate PR because i think this PR is big enough and I don't want to complicate it further. I will create a linear issue :+1:

changes/143.feature.md
```diff
@@ -0,0 +1,46 @@
+Added the quantum reservoir functional with CUDA and QuTiP backend support, plus documentation coverage.
```
Lumorti: Nit: the word coverage probably isn't needed.
```suggestion
Added the quantum reservoir functional with CUDA and QuTiP backend support, plus documentation.
```

changes/143.feature.md
Lumorti, line 1: General comment, right now it's quite verbose to make a reservoir, I get that it's still quite a new field, but how does it act if I don't give some of the requirements? Like are there any standards that people always use (e.g. maybe the preprocessing is always just a generic rotation) and thus maybe if someone doesn't give any arguments it still creates a "reasonable" reservoir? I'm only at the start of looking through it right now, so maybe later I see that there are defaults everywhere, but just wondering.

changes/143.feature.md
```diff
@@ -0,0 +1,46 @@
...
+    input_per_layer=[
+        {"phi_1": 0.2, "gamma_1": 0.1 },
+        {"phi_1": 0.3, "gamma_1": 0.2 },
+        {"phi_1": 0.4, "gamma_1": 0.3 },
+        {"phi_1": 0.5, "gamma_1": 0.4 },
+        ],
+    store_final_state=True,
+    store_intermediate_states=True,
+)
+backend = QutipBackend()
```
Lumorti: I thought you mentioned you were using a mix of Qutip and QiliSim?

src/qilisdk/core/parameterizable.py
```diff
@@ -30,36 +32,110 @@ def __init__(self) -> None:
...
+        return list(self.get_parameters(trainable=trainable, parameter_filter=parameter_filter).keys())
 
-    def get_parameters(self) -> dict[str, RealNumber]:
-        """Return a mapping from parameter labels to their current numerical values."""
-        return {label: param.value for label, param in self._parameters.items()}
+    def get_parameters(
+        self,
+        trainable: bool | None = None,
+        parameter_filter: Callable[[Parameter], bool] | None = None,
+    ) -> dict[str, RealNumber]:
```
fedonman, line 124: Why are we adding both `trainable: bool` filtering and `parameter_filter: Callable`? My understanding was that we introduced `trainable` as a quick compromise because you didn’t want to add callable-based filtering in this PR.

We should pick one:

a) Keep only `trainable`.

b) Keep only `parameter_filter`, and rename it to `where` for consistency with the rest of the codebase. (I’d prefer b).

This applies to all other get/set methods as well.

docs/fundamentals/functionals.rst
Lumorti, line 1: Is the reservoir not a complex functional, since it includes sampling and time evolution (implicitly) inside of it? I get that it doesn't explicitly take a functional as an argument which does make it different from the variational program, but in terms of vibe I feel like it's closer to a variational program than to a "pure" functional like Sampling or Time Evolution. On that note, should we at some point separate the more complex functionals to a different class? Like having a class called CompoundFunctional or QuantumAlgorithm, to avoid confusion as our list of complex functionals grows.

docs/fundamentals/functionals.rst
```diff
@@ -199,6 +203,192 @@ we can execute it on the Qutip backend:
...
+------------------
+
+The :class:`~qilisdk.functionals.quantum_reservoirs.QuantumReservoir` functional executes a reservoir pipeline over a
+sequence of inputs. Each layer applies an optional pre-processing circuit, an analog reservoir dynamics block
+(:class:`~qilisdk.analog.schedule.Schedule`), and an optional post-processing circuit, followed by measurements of one
+or more observables. The optional ``qubits_to_reset`` list can be used to reset selected qubits between layers.
+
+The reservoir inputs are represented using :class:`~qilisdk.functionals.quantum_reservoirs.ReservoirInput` parameters.
+These behave like standard :class:`~qilisdk.core.variables.Parameter` objects but are marked as non-trainable so they can
+be driven by input data sequences rather than optimization loops.
```
Lumorti, line 218: Is it needed to mention the fact that they're non-trainable? I feel like this is an internal implementation detail and could confuse someone, like maybe they then think that they can't change their values or or that they act differently or something.
```suggestion
```

docs/fundamentals/functionals.rst
```diff
@@ -199,6 +203,192 @@ we can execute it on the Qutip backend:
...
+Quantum reservoir execution is supported by the :class:`~qilisdk.backends.cuda_backend.CudaBackend` and
+:class:`~qilisdk.backends.qutip_backend.QutipBackend`.
```
Lumorti: This is duplicate info with the backends doc page, which is fine to have, but we need to remember to update this as well when we add support to QiliSim

docs/fundamentals/functionals.rst
```diff
@@ -199,6 +203,192 @@ we can execute it on the Qutip backend:
...
+
+**Parameters**
+
+- **initial_state** (:class:`~qilisdk.core.qtensor.QTensor`): Initial state of the reservoir.
+- **reservoir_layer** (:class:`~qilisdk.functionals.quantum_reservoirs.ReservoirLayer`): Defines pre/post-processing,
+  reservoir dynamics, observables, and reset policy.
+- **input_per_layer** (List[Dict[str, float]]): Input values to apply at each layer, keyed by input parameter names
+  (typically the labels of :class:`~qilisdk.functionals.quantum_reservoirs.ReservoirInput`).
```
Lumorti, line 226: Typically? Isn't this always the case? Even if there is circuit-parameter-naming weirdness it will still always be the names of the parameters that are used as the keys, no?


```suggestion
- **input_per_layer** (List[Dict[str, float]]): Input values to apply at each layer, keyed by input parameter names.
```

src/qilisdk/core/parameterizable.py
```diff
@@ -30,36 +32,110 @@ def __init__(self) -> None:
...
+        local_params = getattr(self, "_parameters", {})
+        yield from local_params.items()
+        for child in self._iter_parameter_children():
+            yield from child._iter_parameter_items()  # noqa: SLF001
+
+    def _parameter_map(self) -> dict[str, Parameter]:
+        params: dict[str, Parameter] = {}
+        for label, parameter in self._iter_parameter_items():
+            params[label] = parameter
+        return params
```
fedonman: Why do we need to build a new dictionary every time we iterate over the parameters? Couldn’t the other methods either:

a) call `_iter_parameter_items` directly and consume the iterator, or
b) store this dictionary as an attribute so it’s created only once?

docs/fundamentals/functionals.rst
```diff
@@ -199,6 +203,192 @@ we can execute it on the Qutip backend:
...
+- **store_final_state** (bool, optional): Store the final state after the last layer.
+- **store_intermediate_states** (bool, optional): Store the state after each layer.
```
Lumorti:
```suggestion
- **store_final_state** (bool, optional): Whether to store the final state after the last layer.
- **store_intermediate_states** (bool, optional): Whether to store the state after each layer.
```

src/qilisdk/core/parameterizable.py
```diff
@@ -30,36 +32,110 @@ def __init__(self) -> None:
...
+        """Yield parameterizable children to compose this object's parameter interface.
+
+        Returns:
+            Iterable[Parameterizable]: Child objects that contribute parameters to this instance.
+        """
+        return ()
+
+    def _iter_parameter_items(self) -> Iterable[tuple[str, Parameter]]:
+        """Yield ``(label, parameter)`` items exposed by this object."""
+        local_params = getattr(self, "_parameters", {})
```
fedonman: Avoid using `getattr` here, since we already have direct access to the `_parameters` attribute.

src/qilisdk/backends/backend.py
Lumorti, line 1: Implementation is very nice! I thought it would be much more code but it's surprisingly compressed

src/qilisdk/backends/qutip_backend.py
```diff
@@ -35,8 +35,8 @@
 from qilisdk.settings import get_settings
 
 if TYPE_CHECKING:
+    from qilisdk.functionals import TimeEvolution
```
Lumorti, line 38: In general I do find we're quite inconsistent with the imports, like sometimes we do the full path (i.e. `from qilisdk.functionals.time_evolution`) and sometimes the level above ((i.e. `from qilisdk.functionals`) . Is there a performance hit from doing it like this? If not, maybe we should generally use this notation since then it forces us to make sure the init.py imports the children correctly. Or could we even take it one level higher and do `from qilisdk import TimeEvolution, Sampling` etc.?

src/qilisdk/core/parameterizable.py
```diff
@@ -96,20 +187,24 @@ def set_parameter_bounds(self, ranges: dict[str, tuple[float, float]]) -> None:
...
+            available_parameters[label].set_bounds(bound[0], bound[1])
 
     def get_constraints(self) -> list[ComparisonTerm]:
         """Get all constraints on the parameters.
 
         Returns:
             list[ComparisonTerm]: A list of comparison terms involving the parameters of the Object.
         """
-        return self._parameter_constraints
+        constraints = list(getattr(self, "_parameter_constraints", []))
```
fedonman: Same here, we should access `_parameter_constraints` directly.

src/qilisdk/core/parameterizable.py
```diff
@@ -30,36 +32,110 @@ def __init__(self) -> None:
...
+        parameter_filter: Callable[[Parameter], bool] | None = None,
+    ) -> dict[str, Parameter]:
+        params = self._parameter_map()
+        if trainable is None:
+            filtered = params
+        else:
+            filtered = {label: param for label, param in params.items() if param.is_trainable is trainable}
+        if parameter_filter is None:
+            return filtered
+        return {label: param for label, param in filtered.items() if parameter_filter(param)}
```
Lumorti: This is very nice to have!

src/qilisdk/core/qtensor.py
```diff
@@ -445,6 +448,58 @@ def to_density_matrix(self) -> QTensor:
...
+        reference_norm = max(self.norm(order="fro"), get_settings().atol)
+        relative_correction = (repaired_state - self).norm(order="fro") / reference_norm
+        if relative_correction > max_relative_correction:
+            raise ValueError(
+                "Repairing the density matrix required a large correction "
+                f"(relative Frobenius correction={relative_correction:.3e})."
+            )
+
+        if not repaired_state.is_density_matrix():
+            raise ValueError("Repaired state is still not a valid density matrix.")
```
Lumorti: Careful with this, `is_density_matrix` could be expensive since it does an eigendecomposition, the repair is cheap (since it's just rho + rho_dag / 2) but checking positivity could be a bottleneck if you do a lot of repairs.

src/qilisdk/core/qtensor.py
```diff
@@ -698,3 +753,45 @@ def expect_val(operator: QTensor, state: QTensor) -> Number:
...
+        ValueError: If ``state`` is not a density matrix.
+        ValueError: If any requested qubit index is out of range.
+
+    Returns:
+        QTensor: The density matrix with the requested qubits reset to ``|0><0|``.
+    """
+    if len(qubits_to_reset) == 0:
+        return state
+    if not state.is_density_matrix():
+        raise ValueError("`reset_qubits` requires a density matrix input state.")
```
Lumorti, line 748: See above comment, also if you have already called repair you might be able to skip the check here

src/qilisdk/core/qtensor.py
```diff
@@ -445,6 +448,58 @@ def to_density_matrix(self) -> QTensor:
             raise ValueError("Cannot normalize density matrix with zero trace.")
         return QTensor(rho.data / tr)  # keep it sparse
 
+    def repair_density_matrix(self, max_relative_correction: float = _MAX_RESERVOIR_STATE_CORRECTION) -> QTensor:
```
fedonman: I don’t think this should be a public instance method on `QTensor`.

1. It takes an argument whose default is clearly reservoir-specific (judging by the name).
2. It mostly just delegates to other public instance methods (e.g. `.to_density_matrix()`), except for a special-case correction.

In practice, this is an “improved” `to_density_matrix()`. We should avoid exposing two overlapping public methods.

We should either:
a) Fold this into the public `to_density_matrix()` by adding a `max_relative_correction` argument, or
b) Move it to `utils` or the class that uses it.

src/qilisdk/core/qtensor.py
```diff
@@ -698,3 +753,45 @@ def expect_val(operator: QTensor, state: QTensor) -> Number:
...
+        part_2 = []
+        for avail_qubit in range(aux_state.nqubits):
+            if qubit > avail_qubit:
+                part_1.append(avail_qubit)
+            elif qubit < avail_qubit:
+                part_2.append(avail_qubit)
+        state_part_1 = aux_state.ptrace(part_1)
+        state_part_2 = aux_state.ptrace(part_2)
+        aux_state = tensor_prod([state_part_1, zero, state_part_2])
+    return aux_state
```
Lumorti: I'm not too familiar with this, but is there no way to reset multiple at once? Like say I have 5 qubits and I want to reset 2 and 3, couldn't I do that in a single pass rather than having to do 2 and then 3? And if the qubits to reset aren't adjacent, doing a reindex first to make them adjacent (or even moving all qubits to be reset to the start, then just doing `tensor_prod(zero, everything_else)` and then reindex back.

But that might be a lot of work, I don't know how much of a bottleneck this section is, I'm just wary about excess `tensor_prod`s

src/qilisdk/cost_functions/model_cost_function.py
```diff
@@ -99,7 +99,7 @@ def _compute_cost_time_evolution(self, results: TimeEvolutionResult) -> Number:
                 )
                 variable_map = {v: int(state[i]) for i, v in enumerate(self.model.variables())}
                 evaluate_results = self.model.evaluate(variable_map)
-                total_cost += sum(v for v in evaluate_results.values()) * _prob
+                total_cost += sum(evaluate_results.values()) * _prob
```
Lumorti, line 102: Huh, surprised the linters never complained about this before

src/qilisdk/digital/circuit.py
```diff
@@ -250,6 +304,21 @@ def prepend(self, circuit: Circuit) -> None:
...
+        """
+        dim = 1 << self.nqubits
+        operator = np.eye(dim, dtype=_complex_dtype())
+        for gate in self.gates:
+            operator = _apply_gate_left(operator, gate, self.nqubits)
+        return operator
+
+    def to_qtensor(self) -> QTensor:
+        return QTensor(self.to_matrix())
+
```
Lumorti, line 321: Very useful to have, nice!

src/qilisdk/digital/circuit.py
```diff
@@ -113,10 +163,9 @@ def set_parameter_values(self, values: list[float]) -> None:
         Raises:
             ParametersNotEqualError: If the number of provided values does not match the expected number of parameters.
         """
-        if len(values) != self.nparameters:
+        if len(values) != len(self.get_parameter_names(trainable=trainable, parameter_filter=parameter_filter)):
             raise ParametersNotEqualError
```
fedonman: Isn't this checked at the `Parameterizable` interface now?

src/qilisdk/digital/gates.py
```diff
@@ -196,11 +216,15 @@ def set_parameter_values(self, values: list[float]) -> None:
...
             raise ParametersNotEqualError
 
-    def get_parameter_bounds(self) -> dict[str, tuple[float, float]]:  # noqa: PLR6301
-        return {}
+    def get_parameter_bounds(
+        self,
+        trainable: bool | None = None,
+        parameter_filter: Callable[[Parameter], bool] | None = None,
+    ) -> dict[str, tuple[float, float]]:
+        return super().get_parameter_bounds(trainable=trainable, parameter_filter=parameter_filter)
```
Lumorti: General question, why do we always have to re-implement the parameterizable methods? Shouldn't they be left the same as the original in almost all cases? That's the point of inheriting from a common base class, no?

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,387 @@
...
+        """
+        super().__init__(
+            label=label,
+            value=value,
+            domain=Domain.REAL,
+            bounds=(None, None),
+            trainable=False,
+        )
+
+    __hash__ = Parameter.__hash__
```
Lumorti: Is this not automatically inferred?

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,387 @@
...
+
+
+class ReservoirLayer(Parameterizable):
+    """Single reservoir layer template.
+
+    A reservoir layer can contain up to three ordered stages:
+    ``pre_processing -> reservoir_dynamics -> post_processing``.
```
Lumorti: Should this be more consistent with the actual names?


```suggestion
    ``input_encoding -> evolution_dynamics -> output_encoding``.
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,387 @@
...
+            yield self._output_encoding
+
+    def __len__(self) -> int:
+        """
+        Get the total number of steps in the reservoir pass (maximum of 3)
+
+        Returns:
+            int: The number of steps in the reservoir pass.
+        """
+        return int(self._input_encoding is not None) + int(self._output_encoding is not None) + 1
```
Lumorti, line 234: Is this the most intuitive form for `len(reservoir)`? Is there actually a use case for this? Since if the user does this and gets "2" as the result, they still then have to check whether it's missing the input or the output encoding. So maybe `len` should instead be the sum of the various internals since then it lets the user gauge the vague complexity of their reservoir.
```suggestion
    def __len__(self) -> int:
        """
        Get the total number of steps in the reservoir pass (maximum of 3)

        Returns:
            int: The number of steps in the reservoir pass.
        """
        input_length = len(self._input_encoding) if self._input_encoding else 0
        evolution_length = len(self._evolution_dynamics)
        output_length = len(self._output_encoding) if self._input_encoding else 0
        return input_length + evolution_length + output_length
```
Granted this might require adding `len` methods to these classes if they don't already exist.

src/qilisdk/digital/circuit.py
```diff
@@ -250,6 +304,21 @@ def prepend(self, circuit: Circuit) -> None:
...
+            GateHasNoMatrixError: if any gate does not define a matrix (e.g., measurement).
+        """
+        dim = 1 << self.nqubits
+        operator = np.eye(dim, dtype=_complex_dtype())
+        for gate in self.gates:
+            operator = _apply_gate_left(operator, gate, self.nqubits)
+        return operator
+
+    def to_qtensor(self) -> QTensor:
+        return QTensor(self.to_matrix())
```
fedonman, line 320: Shouldn’t we construct the `QTensor` directly, instead of first building the dense (N^2 x N^2) NumPy matrix?

src/qilisdk/digital/circuit.py
```diff
@@ -19,32 +19,65 @@
...
+    perm_output = [out_positions[q] for q in range(nqubits)]
+    perm = perm_output + list(range(nqubits, 2 * nqubits))
+
+    return np.transpose(contracted, axes=perm).reshape(operator.shape)
+
+
 @yaml.register_class
 class Circuit(Parameterizable):
-    def __init__(self, nqubits: int) -> None:
+    def __init__(self, nqubits: int, parameter_prefix: str | None = None) -> None:
```
fedonman: This shouldn't be Circuit-specific. It should be in `Parameterizable` level. And ideally, shouldn't be an argument of `__init__()` method, because everyone will inherit it and will mess up the `__init__()` signatures. It could be a public method that sets it. E.g.

```
class Parameterizable:
   def __init__(self):
        self._prefix = ""
   
   def set_prefix(self, prefix: str) -> None:
        self._prefix = prefix
```

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,387 @@
...
+        """Number of shots used by backend executions."""
+        return self._nshots
+
+    @property
+    def input_parameter_names(self) -> list[str]:
+        """Input-driven (non-trainable) parameter names expected per layer."""
+        return self._reservoir_layer.input_parameter_names
+
+    def _iter_parameter_children(self) -> Iterator[Parameterizable]:
+        yield self._reservoir_layer
```
Lumorti, line 348: Does this definitely work? Like does this then yield over the __iter__ of reservoir layer? Or does it just yield the reservoir layer object without iterating over the children?

src/qilisdk/functionals/quantum_reservoirs_result.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def _complex_dtype() -> np.dtype:
+    """Return the configured complex dtype for reservoir result arrays."""
+    return get_settings().complex_precision.dtype
+
+
+@yaml.register_class
+class QuantumReservoirResult(TimeEvolutionResult):
+    """Container for expectation values and states produced by a reservoir simulation."""
```
Lumorti, line 30: Is this just a placeholder or is it really just a relabelling of the time evolution result? I thought for sure there would be some unique outputs, like maybe the list of final parameter values etc.

tests/integration/backends/test_all_backend_integration.py
```diff
@@ -479,3 +507,37 @@ def test_time_dependent_hamiltonian_density_mat(backend):
...
+    monkeypatch.setattr(
+        "qilisdk.backends.qutip_backend.QutipBackend._execute_time_evolution",
+        lambda self, f: TimeEvolutionResult(final_state=final_density),
+    )
+
+    result = backend._execute_quantum_reservoir(functional)
+
+    assert result.final_state is not None
+    assert len(result.expected_values) == 2
+    assert len(result.intermediate_states) == 2
```
Lumorti, line 543: I would say these aren't really backend integration tests, since they don't actually use the backends. Maybe they should be moved to a different file.

src/qilisdk/functionals/quantum_reservoirs.py
```diff
@@ -0,0 +1,387 @@
...
+            value=value,
+            domain=Domain.REAL,
+            bounds=(None, None),
+            trainable=False,
+        )
+
+    __hash__ = Parameter.__hash__
+
+
+class ReservoirLayer(Parameterizable):
```
fedonman, line 61: Maybe `ReservoirInput` and `ReservoirLayer` should be moved into a dedicated module, e.g. `reservoirs`.

They’re core building blocks in the same way that `Circuit`, `Schedule`, and `Hamiltonian` are the building blocks for the `Sampling` and `TimeEvolution` functionals.

tests/unit/digital/test_circuit.py
```diff
@@ -20,9 +20,28 @@
...
+    expanded = np.kron(gate.matrix, np.eye(2 ** (nqubits - k), dtype=gate.matrix.dtype))
+    order = list(gate.qubits) + remaining
+    tensor = expanded.reshape([2] * nqubits + [2] * nqubits)
+
+    out_pos = {q: i for i, q in enumerate(order)}
+    perm_out = [out_pos[q] for q in range(nqubits)]
+    perm_in = [nqubits + out_pos[q] for q in range(nqubits)]
+    perm = perm_out + perm_in
+
+    return np.transpose(tensor, axes=perm).reshape(2**nqubits, 2**nqubits)
```
Lumorti: I feel like maybe we should just precalculate the correct gate expansion for the cases we want, since adding this function means we test one implementation versus another. If we do this, we should also add tests for this implementation, which would require us to compare versus another implementation, and so on.

Lumorti (changes requested):
Very nice work! Just a few questions and a few minor comments/fixes.

src/qilisdk/functionals/quantum_reservoirs.py
fedonman: Is this needed?

fedonman (changes requested):
Left some more comments.

changes/143.feature.md
amirazzam2000: I think for that we can build some pre-defined reservoir funcitonals and the user can just use that directly.

changes/143.feature.md
amirazzam2000: it's not a mix, the circuit part for all simulators just constructs the unitary of the circuit and multiplies it by the state. so it's the same for all backends. but this is pure QTensor.

docs/fundamentals/functionals.rst
amirazzam2000: I agree, but here is a counter point: 
- The inclusion of Time Evolution and Sampling here is to do the operations we have already coded and know how to do. Like yes these two functionals are basic and they do a simple straight forward operation. But the idea here is that you want to reuse workflows that were built for these two functionals and not explicity include them as an input to your functional. do you know what i mean?

src/qilisdk/functionals/quantum_reservoirs.py
amirazzam2000: The len was implemented for the iterator and this proposal breaks how this workflow works. I need to loop over the operations not the components of these operations. Like I don't care if i have a circuit or a schedule i just need to know that I have something.

src/qilisdk/core/parameterizable.py
amirazzam2000: I kept it as a shortcut because it's very frequently used.

src/qilisdk/core/parameterizable.py
amirazzam2000: well this is because we are not guaranteed that the parameters have not changed since the last time this was called. maybe new parameters were added.

src/qilisdk/digital/circuit.py
amirazzam2000: well but the circuit is the class that needs to know about the prefix. how am i supposed to add this to the circuit if it's on the parameter level?

src/qilisdk/functionals/quantum_reservoirs.py
Lumorti: Ah fair enough then

src/qilisdk/core/parameterizable.py
fedonman: It's not good practice to have multiple filtering arguments, especially if they can overlap. Let's just keep the most generic one and if you want to provide a shortcut you could offer a predefined callable e.g. `is_trainable` that can be imported.

src/qilisdk/core/parameterizable.py
fedonman: Then, we can go with a)? The point is, do you need the dictionary in memory?

src/qilisdk/digital/circuit.py
fedonman: Not the `Parameter`, the `Parameterizable`. (where Circuit inherits from)

amirazzam2000: ahhhhhh yes makes sense, i can move it.

docs/fundamentals/functionals.rst
amirazzam2000: well this is more to highlight that it needs to be a ReservoirInput object, so not any random parameter.

src/qilisdk/backends/qutip_backend.py
amirazzam2000: The performance should be the same, it's just sometimes they are auto-imported before we explicitly expose them through the __init__.py of the folder so then they are imported with the full path.

src/qilisdk/digital/circuit.py
amirazzam2000: This would require some refactoring, I will add a linear task to change the gates to return a qtensor and refactor this method to be directly a qtensor.

src/qilisdk/digital/gates.py
amirazzam2000: yeah some of these methods can be removed. I will do it in the future pr that will fix the circuit/gate parameterization.

src/qilisdk/functionals/quantum_reservoirs_result.py
amirazzam2000: for now the result objects are equivalent. The reservoir itself doesn't do any optimization so there is no need to return the final parameter values.

tests/integration/backends/test_all_backend_integration.py
amirazzam2000: They do use backend.execute and backend._execute_quantum_reservoir. Why do you say they are not integration?

src/qilisdk/functionals/quantum_reservoirs.py
amirazzam2000: this process is recursive, so it will iterate over the children of the reservoir layer.

fedonman (approved):
Great work!! 💯

Lumorti (approved):
Nice changes, also good job noticing some typos (e.g PauliI calling itself a PauliX). Congrats on finally getting reservoirs in QiliSDK!! 🏞️

PR #146, opened 2026-02-02
[SDK-183] Add Noise Models to QiliSim

Lumorti:
Noise models are now supported in QiliSim. Usage is the same as with the CudaBackend:
```python
from qilisdk.digital import Circuit, X
from qilisdk.noise import NoiseModel, BitFlip
from qilisdk.functionals import Sampling

circuit = Circuit(nqubits=1)
circuit.add(X(0))

noise_model = NoiseModel()
noise_model.add(BitFlip(probability=1.0))

backend = QiliSim(seed=42, num_threads=1, noise_model=noise_model)
result = backend.execute(Sampling(circuit, nshots=100))

assert result.samples == {"0": 100}
```


fedonman (approved):
Looks good! Great work! 💯 💯 

For future reference:
1. Could we add unit tests? I can see that we have added integration testing which guarantees overall output behavior but unit tests guarantee internals. (edge cases etc.)
2. Could we add code coverage of unit tests for the cpp project?

Lumorti:
> Looks good! Great work! 💯 💯
> 
> For future reference:
> 
>     1. Could we add unit tests? I can see that we have added integration testing which guarantees overall output behavior but unit tests guarantee internals. (edge cases etc.)
> 
>     2. Could we add code coverage of unit tests for the cpp project?

For sure we should discuss adding testing to the c++ code. The problem is that right now it's hard to unit test c++ functions since we use pytest, so either we would have to expose the c++ function using pybind or write a separate test suite in c++ to call stuff.

amirazzam2000 (approved):
thanks luke! 🙌

PR #152, opened 2026-02-10
[SDK-209] fixing cudaq backend issues

amirazzam2000:
Fixed a bug in the cuda backend with time evolution that was erasing information from the hamiltonian.


Lumorti (approved):
Nice!

PR #154, opened 2026-02-12
[SDK-214] Add Support for CUDA 13.X

fedonman:
CUDA 13 is now officially supported alongside CUDA 12, and the optional dependency system has been refactored to support these mutually exclusive CUDA variants in a clean and explicit way. The internal optional-import mechanism now supports "ANY-of" dependency groups, allowing the `cuda` feature to be satisfied by either the CUDA 12 or CUDA 13 distribution. If neither is installed, CUDA symbols (e.g., `CudaBackend`) resolve to informative stubs that raise an `OptionalDependencyError` with clear installation instructions indicating the valid extras.

Two new mutually exclusive extras are introduced:

* `cuda12` → installs the CUDA 12 backend stack
* `cuda13` → installs the CUDA 13 backend stack

For convenience, aggregated extras are also provided:

* `all-cu12` → installs all optional features plus CUDA 12
* `all-cu13` → installs all optional features plus CUDA 13

For backwards compatibility, the legacy `cuda` and `all` specifiers are temporarily preserved and default to CUDA 12:

* `cuda` → aliases to `cuda12`
* `all` → equivalent to `all-cu12`

This ensures existing environments and CI pipelines continue to work unchanged while users transition to explicit CUDA version selection. Conflicts between incompatible extras (e.g., `cuda12` vs `cuda13`, `all-cu12` vs `all-cu13`) are enforced via `uv` configuration to prevent invalid combinations.

User installation examples:

* Core only (no CUDA):
  `pip install qilisdk`

* CUDA 12 (explicit):
  `pip install "qilisdk[cuda12]"`
  or
  `uv sync --extra cuda12`

* CUDA 13 (explicit):
  `pip install "qilisdk[cuda13]"`
  or
  `uv sync --extra cuda13`

* Full stack with CUDA 12:
  `pip install "qilisdk[all-cu12]"`
  or
  `uv sync --extra all-cu12`

* Full stack with CUDA 13:
  `pip install "qilisdk[all-cu13]"`
  or
  `uv sync --extra all-cu13`

* Backwards-compatible installs (defaulting to CUDA 12):
  `pip install "qilisdk[cuda]"`
  `pip install "qilisdk[all]"`

Developer installation examples:

* Development + CUDA 12:
  `uv sync --group dev --extra cuda12`
  or
  `pip install -e ".[dev,cuda12]"`

* Development + CUDA 13:
  `uv sync --group dev --extra cuda13`
  or
  `pip install -e ".[dev,cuda13]"`

* Development without CUDA (CPU-only CI):
  `uv sync --group dev`

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -43,7 +43,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install the project
-        run: uv sync --group dev --all-extras
+        run: uv sync --group dev --extra all-cu13
         
       - name: Ruff
         run: uv run ruff check --output-format=github .
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,7 +43,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install qilisdk
-        run: uv sync --group dev --all-extras
+        run: uv sync --group dev --extra all-cu13
 
       - name: Run tests
         run: uv run pytest tests/unit/ --cov=qilisdk --cov-report=xml --cov-report=term-missing --junitxml=junit.xml -o junit_family=legacy tests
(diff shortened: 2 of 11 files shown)


Lumorti (changes requested):
Tested locally and it works. The informative error messages are very nice!

Only comment would be that maybe it should be "all-cuda13" and not "all-cu13", we should be consistent with the naming. Unless this is conforming to some convention, I know you mentioned that many libraries do similar things.

fedonman:
> Tested locally and it works. The informative error messages are very nice!
> 
> Only comment would be that maybe it should be "all-cuda13" and not "all-cu13", we should be consistent with the naming. Unless this is conforming to some convention, I know you mentioned that many libraries do similar things.

Yeah, I left it as cu12/cu13 which is the standard way of mentioning cuda versions in packages, e.g. https://pypi.org/project/cuda-quantum-cu12/

Lumorti (approved):
Alright, nice!

PR #155, opened 2026-02-12
fixing cuda-backend issues

amirazzam2000:
Fixed issues with empty elements in cuda hamiltonian that caused issues during simulation.


fedonman (approved):
💯 💯

Lumorti (approved):
Nice, and good unit tests! Hopefully with this we won't have to touch the CUDAQ objects again for a while 🤞

PR #157, opened 2026-02-16
[SDK-217] fixing the noise models with QiliSim not changing with changing rates

amirazzam2000:
Fixed multiple backend issues in analog time evolution: CUDA now supports `QTensor` observables (via Hamiltonian conversion), CUDA Lindblad jump-operator dimension handling was corrected for global/per-qubit noise, and QiliSim now correctly applies Lindblad rates (`jump_operators_with_rates`) and expands global single-qubit jump operators per qubit so dephasing strength affects results as expected.


fedonman (approved):
LGTM, great job!. But I think we are missing tests. `test_qilisim_backend.py` imports only `Dephasing`, we should test all noise types to all backends. I'll create a Linear task for this.

PR #158, opened 2026-02-20
[SDK-220] Fix Serialization issues with hash_cache

amirazzam2000:
Fixed YAML serialization for `@yaml.register_class` objects so transient hash cache fields (`_hash_cache`) is  always dumped as `null` (`None`) without mutating the in-memory object state.


NataliaSirera (approved):
LGTM

PR #159, opened 2026-02-20
[SDK-222] Updating qilisdk object hashing to be deterministic

amirazzam2000:
Reworked custom hashing to a deterministic `hashlib.blake2b`-based `qilisdk.utils.hashing.hash(...)` interface (no Python builtin hash dependency in the hashing pipeline), migrated Variable/Term/ComparisonTerm/QTensor/Hamiltonian/Pauli hashing to it, and added consistency tests for hashing behavior (stability, equality-consistency, and order-insensitive structures).

### DIFF
--- a/src/qilisdk/core/variables.py
+++ b/src/qilisdk/core/variables.py
@@ -24,6 +24,7 @@ from loguru import logger
 
 from qilisdk.core.exceptions import EvaluationError, InvalidBoundsError, NotSupportedOperation, OutOfBoundsException
 from qilisdk.settings import get_settings
+from qilisdk.utils.hashing import hash as qili_hash
 from qilisdk.yaml import yaml
 
 from .types import Number, QiliEnum, RealNumber
@@ -917,7 +918,7 @@ class BaseVariable(ABC):
 
     def __hash__(self) -> int:
         if self._hash_cache is None:
-            self._hash_cache = hash((self._label, self._domain.value, self._bounds))
+            self._hash_cache = qili_hash(self._label)  # , self._domain.value, self._bounds[0], self._bounds[1])
         return self._hash_cache
 
     def __eq__(self, other: object) -> bool:
@@ -1773,7 +1774,7 @@ class Term:
         )
 
     def __hash__(self) -> int:
-        return hash((frozenset(self._elements.items()), self.operation))
+        return qili_hash(self.operation.value, self._elements)
 
     def __eq__(self, other: object) -> bool:
         if not isinstance(other, Term):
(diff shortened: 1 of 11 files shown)


tests/unit/utils/test_hashing.py
Lumorti, line 1: Should there maybe be a few simple tests for QTensor, Hamiltonian and Variable? Just to double to check that everything works as expected there.

Could be in this file, or a new file like test_hashing_qtensor.py etc.

src/qilisdk/core/variables.py
```diff
@@ -917,7 +918,7 @@ def __pow__(self, a: int) -> Term:
 
     def __hash__(self) -> int:
         if self._hash_cache is None:
-            self._hash_cache = hash((self._label, self._domain.value, self._bounds))
+            self._hash_cache = qili_hash(self._label)  # , self._domain.value, self._bounds[0], self._bounds[1])
```
Lumorti, line 921: Why not include this extra information? Is it just for performance? Right now wouldn't this give a hash collision if you had two variables with the same name but different values? I get that shouldn't really happen, but since it is allowed.

tests/unit/utils/test_hashing.py
Lumorti: Could be in this file, or a new file like `test_hashing_qtensor.py` etc.

src/qilisdk/utils/hashing.py
```diff
@@ -0,0 +1,140 @@
...
+
+def _length_prefix(payload: bytes) -> bytes:
+    return len(payload).to_bytes(_CHUNK_SIZE_BYTES, byteorder="big", signed=False)
+
+
+def _encode_scalar(value: object) -> bytes:
+    if value is None:
+        return b"none"
+    if isinstance(value, str):
+        return b"str:" + value.encode("utf-8")
```
Lumorti, line 43: Out of interest, would the qili_hash of "mystringreal:3" give the same hash as ("mystring", 3)? Or would the length_prefix separate them?

Lumorti (changes requested):
Very nice! Only comment would be to add a few extra tests.

src/qilisdk/core/variables.py
amirazzam2000: the reason is that this extra information could be modified. So if they get modified that would change the hash of the object and that should never happen. The hash of the object should be unique and reproducible. The compromise here is that the name of the variable is their unique identifier so two variables with the same name will be considered the same the variable.

src/qilisdk/utils/hashing.py
amirazzam2000: the actually wouldn't, not only all hashes will be separated with length_prefix but also the tuple will be translated into this string before being hashed: `b'sequence:\x00\x00\x00\x00\x00\x00\x00\x0cstr:mystring\x00\x00\x00\x00\x00\x00\x00\x08real:3/1'`

which is different than hashing a string directly.

Lumorti (approved):
Perfect!!

Issue #156, opened 2026-02-12
Documentation formatting issue


draentropia:
In https://qilimanjaro-tech.github.io/qilisdk/main/examples/circuits.html in the OpenQASM section, the tittle is not well displayed in both themes.

draentropia:
That happens if you search a term and then click in the solution. The accent color for the search hides the content.

Lumorti:
Thanks a lot for finding and reporting it! Now fixed.

PR #160, opened 2026-02-23
[SDK-224] Docs Highlight Bug

Lumorti:
Fixed a bug in the documentation whereby titles could be hidden by a highlight after searching.

https://github.com/qilimanjaro-tech/qilisdk/issues/156

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -14,6 +14,11 @@
   src: url("PlusJakartaSans-Italic-VariableFont_wght.ttf") format("truetype-variations");
 }
 
+/* No search term highlighting of the titles */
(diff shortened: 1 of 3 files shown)


src/qilisdk/noise/readout_assignment.py
Lumorti, line 1: (This was just something that sphinx was complaining about, I tried fixing the other warnings but couldn't figure them out)

docs/_static/custom.css
Lumorti, line 1: Basically this now disables the background highlight on the title elements (but keeps the underline).

amirazzam2000 (approved):
thanks luke! 🙏

PR #161, opened 2026-02-24
[SDK-228] Convert SpeQtrum API key to `SecretStr`

fedonman:
Converted the Speqtrum API key type from `str` to Pydantic's `SecretStr` to better protect sensitive values.


src/qilisdk/speqtrum/speqtrum.py
```diff
@@ -306,7 +306,11 @@ def login(
         # Use provided parameters or fall back to environment variables via Settings()
         settings = get_settings()
         username = username or settings.speqtrum_username
-        apikey = apikey or settings.speqtrum_apikey
+        apikey = (
+            apikey or settings.speqtrum_apikey
+            if settings.speqtrum_apikey is None
+            else settings.speqtrum_apikey.get_secret_value()
+        )
```
Lumorti: Shouldn't it be the following? Or maybe I'm misunderstanding the intention.

```suggestion
        apikey = (
            apikey or settings.speqtrum_apikey
            if settings.speqtrum_apikey is not None
            else settings.speqtrum_apikey.get_secret_value()
        )
```

Lumorti (changes requested):
Nice! Just want to double-check that the if/else is correct.

src/qilisdk/speqtrum/speqtrum.py
fedonman: No, if `speqtrum_apikey` is **not** None, you want to call the `.get_secret_value()` on the object. Otherwise, return `speqtrum_apikey` (which is now None).

fedonman: Essentially it is this:

```
        apikey = (
            apikey or None
            if settings.speqtrum_apikey is None
            else settings.speqtrum_apikey.get_secret_value()
        )
 ```

Comment (Lumorti) [reply src/qilisdk/speqtrum/speqtrum.py]: Ah yes I see now, thanks!

**Review by Lumorti — APPROVED** [review]

File: src/qilisdk/speqtrum/speqtrum.py (hunk 1 above)
Comment (fedonman) [reply src/qilisdk/speqtrum/speqtrum.py]: I rewrote it for clarity as:

```
        apikey = (
            apikey or settings.speqtrum_apikey.get_secret_value()
            if settings.speqtrum_apikey is not None
            else None
        )
```

File: src/qilisdk/speqtrum/speqtrum.py [hunk 2]
```diff
@@ -306,7 +306,7 @@ def login(
         # Use provided parameters or fall back to environment variables via Settings()
         settings = get_settings()
         username = username or settings.speqtrum_username
-        apikey = apikey or settings.speqtrum_apikey
+        apikey = apikey or settings.speqtrum_apikey.get_secret_value() if settings.speqtrum_apikey is not None else None
```
Lumorti: Maybe you need parentheses?
```suggestion
        apikey = apikey or (settings.speqtrum_apikey.get_secret_value() if settings.speqtrum_apikey is not None else None)
```

fedonman: Yes, just fixed it. Thanks

PR #162, opened 2026-02-25
[SDK-229] Fixing OpenQasm 2.0's parsing 

amirazzam2000:
Fixed OpenQASM 2.0 import parsing by replacing parameter parsing with AST-based expression evaluation (supporting `pi`, arithmetic, and standard functions like `sin`/`cos`/`sqrt`), handling inline `//` comments, and adding stricter gate-parameter parsing with clear errors for invalid/deeply nested expressions; expanded unit tests to cover these cases and full supported gate round-trips.


src/qilisdk/utils/openqasm2.py
```diff
@@ -35,6 +37,133 @@
...
+    raise ValueError(f"Unsupported parameter expression: {original_expr}")
+
+
+def _parse_qasm2_gate_line(line: str) -> tuple[str, str | None, str] | None:
+    """Parse a QASM gate line with bounded parenthesis nesting.
+    Returns:
+        tuple[str, str | None, str] | None: the gate name, the string representing the parameter if available, and the rest.
+
+    Raises:
+        ValueError: if the parameter expression is nested for deeper than one level, or the expression is invalid.
```
Lumorti:
```suggestion
        ValueError: if the parameter expression is nested for deeper than the max depth allowed, or the expression is invalid.
```

src/qilisdk/utils/openqasm2.py
```diff
@@ -35,6 +37,133 @@
...
+
+    params_str = None
+    if rest.startswith("("):
+        depth = 0
+        closing_index = None
+        for index, char in enumerate(rest):
+            if char == "(":
+                depth += 1
+                if depth > _MAX_DEPTH:
+                    raise ValueError("Parameter expression nesting deeper than one level is not supported.")
```
Lumorti:
```suggestion
                    raise ValueError(f"Parameter expression nesting deeper than {_MAX_DEPTH} levels is not supported.")
```

src/qilisdk/utils/openqasm2.py
```diff
@@ -35,6 +37,133 @@
...
+_ALLOWED_QASM2_FUNCTIONS = {
+    "sin": math.sin,
+    "cos": math.cos,
+    "tan": math.tan,
+    "exp": math.exp,
+    "ln": math.log,
+    "sqrt": math.sqrt,
+}
+
+_MAX_DEPTH = 2
```
Lumorti, line 49: Why 2 and not like 3 or 4? What about general operations without functions, like "(((3 + 2) * 1.3) + 2*4.8)"?

Say the max depth was 4, I guess it just means that worst case it would scale as O(n^4) rather than O(n^2) as it is now. As long as there is a finite bound probably it's fine (i.e. making sure it's not O(n^m) where they can choose the m).

Lumorti (changes requested):
Good job! Just a few minor comments.

fedonman (changes requested):
Nice!! 💯 Add a changelog entry before merging!

src/qilisdk/utils/openqasm2.py
amirazzam2000: I assumed that we won't have these types of operations as part of a qasm file, the constant multiplications would all be evaluated beforehand. The reason i set it to two is because this will only be used in cases like ry(sin(2*pi)) or something like that.

Lumorti (approved):
💯

fedonman (approved):

PR #164, opened 2026-02-26
[SDK-232] Move Circuit Transpiler to QiliSDK

fedonman:
Added a composable digital circuit transpilation framework centered around `CircuitTranspiler` and concrete implemetations of `CircuitTranspilerPass`. The new transpiler is intentionally thin: it executes an ordered list of non-mutating passes, resets shared state for every `transpile()` call, and returns a `CircuitTranspilerResult` that exposes the transpiled circuit through `circuit`, the per-pass outputs through `intermediate_results`, the final user-facing logical-to-physical mapping through `layout`, and any pass-level diagnostics through `metrics`.

The default transpiler has two modes. Without topology information, `CircuitTranspiler.default()` builds a local-rewrite pipeline made of `DecomposeMultiControlledGatesPass`, `CancelIdentityPairsPass`, `DecomposeToCanonicalBasisPass`, and `FuseSingleQubitGatesPass`. This path is meant for basis normalization and cheap circuit cleanup without introducing device-specific constraints. When a topology is provided, the default pipeline becomes hardware-aware: it performs the same decomposition and simplification steps first, then applies layout and routing (`SabreLayoutPass` plus `SabreSwapPass` when no mapping is given, or `CustomLayoutPass` when a user mapping is supplied), and finally runs canonical decomposition and single-qubit fusion again so any routing artefacts are pushed back into the requested gate basis.

`DecomposeMultiControlledGatesPass` handles the structural lowering of multi-controlled single-qubit operations before any basis work is attempted. Gates such as `X(2).controlled(0, 1)` are recursively rewritten into equivalent sequences containing only single-control operations, square roots of the target gate, their adjoints, and auxiliary controlled Pauli-X operations. Single-control gates are left unchanged, while controlled versions of multi-qubit payloads are rejected explicitly, which keeps later passes focused on 1Q/2Q synthesis only.

`CancelIdentityPairsPass` performs fixed-point cancellation of gates whose product is the identity up to global phase, even when those gates are separated by operations on disjoint qubits. It removes explicit `I` gates immediately, recognizes self-inverse primitives such as `H`, `X`, `Y`, `Z`, `CNOT`, `CZ`, and `SWAP`, cancels parametric inverses such as `RX(theta)` with `RX(-theta)` and `U3(theta, phi, gamma)` with `U3(-theta, -gamma, -phi)`, understands `Adjoint(...)` wrappers and controlled inverse pairs, and falls back to matrix-based matching for custom gates that expose a unitary. Measurements and gates without a matrix act as barriers on the qubits they touch, so the pass stays conservative around non-unitary or opaque operations.

`DecomposeToCanonicalBasisPass` is responsible for basis normalization. It maps supported one-qubit gates into either a `U3` basis or an `RX`/`RY`/`RZ` basis, depending on `single_qubit_basis`, and rewrites two-qubit structure into either `CNOT`-based or `CZ`-based entangling skeletons, depending on `two_qubit_basis`. In practice this means `H`, `X`, `Y`, `Z`, `U1`, `U2`, arbitrary one-qubit `BasicGate` instances, `CNOT`, `CZ`, `SWAP`, single-control controlled one-qubit gates, `Adjoint(...)`, and one-qubit `Exponential(...)` gates can all be rewritten into a circuit containing only the chosen 1Q basis, the chosen 2Q entangler, and measurements. Multi-controlled gates are deliberately out of scope here and are expected to be lowered by `DecomposeMultiControlledGatesPass` first.

`FuseSingleQubitGatesPass` then compresses maximal adjacent runs of one-qubit unitary gates on each wire into a basis-respecting sequence. In `SingleQubitGateBasis.U3` mode, each fused run is emitted as a single `U3`. In `SingleQubitGateBasis.RxRyRz` mode, the pass emits compact axis-aligned rotations when the fused unitary is recognizable as pure `RX`, `RY`, or `RZ`, and otherwise falls back to a canonical `RZ-RY-RZ` realization. Multi-qubit gates, measurements, and any unsupported or non-unitary operations flush pending fusion on the affected qubits, so gate order and circuit semantics are preserved.

The topology-aware passes cover both automatic placement and user-directed placement. `SabreLayoutPass` is a layout-only pass that chooses a logical-to-physical mapping with a SABRE-style heuristic using a front layer, a bounded look-ahead set, randomized trials, and a light decay penalty to avoid thrashing. It retargets the entire circuit to physical qubits, can enlarge the returned circuit to the device size, and records the selected initial layout in the shared context. `SabreSwapPass` starts from either an explicit `initial_layout`, the layout already present in the context, or a default identity-style placement, and then inserts SWAPs so every emitted two-qubit gate acts on an edge of the topology while preserving the original gate order. It stores the routed final layout in `result.layout` and exposes `swap_count` through `result.metrics`. `CustomLayoutPass` is the deterministic alternative for users who already know where logical qubits should live: it validates that the mapping covers every logical qubit exactly once, is injective, and only references physical qubits in the topology; it then retargets the circuit to the requested physical qubits and, when necessary, inserts shortest-path SWAPs and immediately undoes them so later operations remain pinned to the user-requested mapping.

Changing the target basis is now a one-line configuration change on the default transpiler. The example below requests an `RX`/`RY`/`RZ` one-qubit basis and `CZ` as the only two-qubit entangler, so the final circuit contains only those basis gates plus measurements:

```python
from qilisdk.digital import CNOT, H, Circuit
from qilisdk.digital.circuit_transpiler import CircuitTranspiler
from qilisdk.digital.circuit_transpiler_passes import SingleQubitGateBasis, TwoQubitGateBasis

circuit = Circuit(2)
circuit.add(H(0))
circuit.add(CNOT(0, 1))

transpiler = CircuitTranspiler.default(
    single_qubit_basis=SingleQubitGateBasis.RxRyRz,
    two_qubit_basis=TwoQubitGateBasis.CZ,
)
result = transpiler.transpile(circuit)

assert all(gate.name in {"RX", "RY", "RZ", "CZ"} for gate in result.circuit.gates)
```

Supplying a topology activates layout and routing automatically. The topology can be provided either as a `rustworkx.PyGraph` or as a simple edge list, and the transpiler will return both the routed circuit and routing diagnostics:

```python
from qilisdk.digital import CZ, Circuit
from qilisdk.digital.circuit_transpiler import CircuitTranspiler

circuit = Circuit(3)
circuit.add(CZ(0, 2))

transpiler = CircuitTranspiler.default(topology=[(0, 1), (1, 2)])
result = transpiler.transpile(circuit)

print(result.layout)              # final logical -> physical mapping after routing
print(result.metrics["swap_count"])
```

If you want to pin logical qubits to specific physical locations, pass `qubit_mapping` together with the topology. In this mode the transpiler uses `CustomLayoutPass`, keeps the reported final layout equal to the user mapping, and routes non-adjacent two-qubit interactions with temporary SWAPs under the hood:

```python
from qilisdk.digital import CZ, RX, Circuit
from qilisdk.digital.circuit_transpiler import CircuitTranspiler

circuit = Circuit(2)
circuit.add(RX(0, theta=0.5))
circuit.add(CZ(0, 1))

transpiler = CircuitTranspiler.default(
    topology=[(0, 1), (1, 2)],
    qubit_mapping={0: 2, 1: 0},
)
result = transpiler.transpile(circuit)

assert result.layout == {0: 2, 1: 0}
assert result.circuit.nqubits == 3
```

The pipeline is also fully customizable, which makes it easy to build a transpiler that only performs the stages you want and then inspect the circuit produced after each stage. In the example below, the transpiler lowers multi-controlled gates, removes trivial inverse pairs, canonicalizes everything to a `CZ` + `U3` basis, and finally fuses one-qubit runs; the returned `intermediate_results` preserve the per-pass circuit snapshots in execution order:

```python
from qilisdk.digital import Controlled, X, Circuit
from qilisdk.digital.circuit_transpiler import CircuitTranspiler
from qilisdk.digital.circuit_transpiler_passes import (
    CancelIdentityPairsPass,
    DecomposeMultiControlledGatesPass,
    DecomposeToCanonicalBasisPass,
    FuseSingleQubitGatesPass,
    SingleQubitGateBasis,
    TwoQubitGateBasis,
)

circuit = Circuit(3)
circuit.add(Controlled(0, 1, basic_gate=X(2)))

transpiler = CircuitTranspiler(
    pipeline=[
        DecomposeMultiControlledGatesPass(),
        CancelIdentityPairsPass(),
        DecomposeToCanonicalBasisPass(
            single_qubit_basis=SingleQubitGateBasis.U3,
            two_qubit_basis=TwoQubitGateBasis.CZ,
        ),
        FuseSingleQubitGatesPass(single_qubit_basis=SingleQubitGateBasis.U3),
    ]
)
result = transpiler.transpile(circuit)

print([step.name for step in result.intermediate_results])
print(result.circuit)
```


src/qilisdk/digital/circuit_transpiler_passes/cancel_identity_pairs_pass.py
```diff
@@ -0,0 +1,333 @@
...
+
+            if not to_delete:
+                break
+
+            gates = [gate for i, gate in enumerate(gates) if i not in to_delete]
+
+        # Build new circuit (deepcopy to avoid Parameter sharing)
+        out = Circuit(circuit.nqubits)
+        for g in gates:
+            out.add(deepcopy(g))
```
Lumorti, line 201: Don't we usually want parameter sharing? If I transpile a circuit which contains some parameter, don't I usually then want my result to use that same parameter reference, just as if I had done new_circuit = old_circuit?

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
```diff
@@ -0,0 +1,342 @@
...
+
+        Args:
+            topology (PyGraph[int, int]): Undirected coupling graph whose node indices are physical qubits.
+            qubit_mapping (Mapping[int, int]): User-specified mapping ``logical -> physical``.
+
+        Raises:
+            TypeError: If ``topology`` is not a ``rustworkx.PyGraph`` instance.
+        """
+        if not isinstance(topology, PyGraph):
+            raise TypeError("CustomLayoutPass requires a rustworkx.PyGraph (undirected).")
```
Lumorti: Didn't I read earlier that we support giving the topology as a list of edges?

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
```diff
@@ -0,0 +1,342 @@
...
+            missing_logical_qubits = sorted(expected_logical_qubits - mapping_keys)
+            extraneous_logical_qubits = sorted(mapping_keys - expected_logical_qubits)
+            error_parts = []
+            if missing_logical_qubits:
+                error_parts.append(f"missing logical qubits {missing_logical_qubits}")
+            if extraneous_logical_qubits:
+                error_parts.append(f"extraneous logical keys {extraneous_logical_qubits}")
+            raise ValueError(
+                "User mapping must map *every* logical qubit in the circuit exactly once; " + "; ".join(error_parts)
+            )
```
Lumorti, line 120: Could we not safely assume that if the user's mapping doesn't mention a qubit, then that qubit doesn't change? Like maybe they have a huge system but they just want to set that qubit 0 <-> qubit 1 or something. But if you prefer to make them be more explicit then I'm also fine with that.

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
```diff
@@ -0,0 +1,342 @@
...
+
+            if len(gate_qubits) <= 1 or isinstance(gate, M):
+                mapped_qubits = tuple(logical_to_physical_layout[logical_qubit] for logical_qubit in gate_qubits)
+                output_circuit.add(self._retarget_gate(gate, mapped_qubits))
+                continue
+
+            if len(gate_qubits) != CustomLayoutPass._TWO_QUBIT_ARITY:
+                raise NotImplementedError(
+                    f"CustomLayoutPass currently supports routing for 1Q/2Q gates only; "
+                    f"received {type(gate).__name__} acting on {len(gate_qubits)} qubits."
```
Lumorti, line 166: Maybe a suggestion here to the user telling them to add a DecomposeMultiControlled pass if the reason its 3+ qubits is because of extra controls

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
```diff
@@ -0,0 +1,342 @@
...
+        if isinstance(gate, U3):
+            return U3(mapped_qubits[0], theta=gate.theta, phi=gate.phi, gamma=gate.gamma)
+
+        # 2-qubit basics
+        if isinstance(gate, CNOT):
+            return CNOT(mapped_qubits[0], mapped_qubits[1])
+        if isinstance(gate, CZ):
+            return CZ(mapped_qubits[0], mapped_qubits[1])
+        if isinstance(gate, SWAP):
+            return SWAP(mapped_qubits[0], mapped_qubits[1])
```
Lumorti, line 266: What about RX/RY/RZ/U3 with a single control? Or would that never reach here?

Are we making assumptions that they already used a decompose pass before using this one?

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
```diff
@@ -0,0 +1,342 @@
...
+        """
+        # 1-qubit basics
+        if isinstance(gate, RX):
+            return RX(mapped_qubits[0], theta=gate.theta)
+        if isinstance(gate, RY):
+            return RY(mapped_qubits[0], theta=gate.theta)
+        if isinstance(gate, RZ):
+            return RZ(mapped_qubits[0], phi=gate.phi)
+        if isinstance(gate, U3):
+            return U3(mapped_qubits[0], theta=gate.theta, phi=gate.phi, gamma=gate.gamma)
```
Lumorti, line 258: Why not support all single-qubit gates in the protocol? Since maybe the user just wants to do a CustomLayoutPass without doing any DecomposeToCanonicalBasis or anything else

src/qilisdk/digital/circuit_transpiler_passes/decompose_to_canonical_basis_pass.py
```diff
@@ -0,0 +1,661 @@
...
+        if isinstance(gate, I):
+            return []
+        if isinstance(gate, (H, X, Y, Z)):
+            return _normalize_single_qubit_sequence([_as_basis_1q(gate)], self._single_qubit_basis)
+
+        # param 1q
+        if isinstance(gate, U1):
+            return _normalize_single_qubit_sequence([_as_basis_1q(gate)], self._single_qubit_basis)
+        if isinstance(gate, U2):
+            return _normalize_single_qubit_sequence([_as_basis_1q(gate)], self._single_qubit_basis)
```
Lumorti, line 525: These could be merged
```suggestion
        if isinstance(gate, (H, X, Y, Z, U1, U2)):
            return _normalize_single_qubit_sequence([_as_basis_1q(gate)], self._single_qubit_basis)
```

Lumorti: In fact, could also be merged with the check for U3, RX, RY and RZ too, since as_basis_1q acting on those simply returns the same gate

src/qilisdk/digital/circuit_transpiler_passes/decompose_to_canonical_basis_pass.py
```diff
@@ -0,0 +1,661 @@
...
+            inverse_sequence: list[Gate] = []
+            for base_gate in reversed(rewritten_base_sequence):
+                inverse_sequence += _invert_basis_gate(base_gate)
+            return _normalize_single_qubit_sequence(inverse_sequence, self._single_qubit_basis)
+
+        # Exponential(1q)
+        if _is_exponential(gate):
+            basic_gate = gate.basic_gate
+            if basic_gate.nqubits != 1:
+                raise NotImplementedError("Exponential of multi-qubit gates not supported.")
```
Lumorti:
```suggestion
                raise NotImplementedError("Decomposing the exponential of a multi-qubit gate to a canonical basis is not supported.")
```
Just to make it clear that the NotImplementedError is specific to here, as exponentials of multi qubit gates might be supported elsewhere

src/qilisdk/digital/circuit_transpiler_passes/fuse_single_qubit_gates_pass.py
Lumorti, line 1: Shouldn't this check for multi-qubit gates and clear pending if found? Right now wouldn't this merge X(0) CNOT(0, 1) X(0) into I(0) CNOT(0, 1)? Maybe this should reuse the same logic as in CancelIdentityPairs for checking for blocking gates.

src/qilisdk/digital/circuit_transpiler_passes/sabre_swap_pass.py
```diff
@@ -0,0 +1,828 @@
...
+    return isinstance(gate, (Adjoint, Exponential))
+
+
+class SabreSwapPass(CircuitTranspilerPass):
+    """
+    SABRE routing (SWAP insertion) for 1Q/2Q circuits on an undirected coupling graph.
+
+    Inputs
+    ------
+    coupling : rustworkx.PyGraph
```
Lumorti: See previous comments

src/qilisdk/digital/circuit_transpiler_passes/sabre_layout_pass.py
```diff
@@ -0,0 +1,751 @@
...
+        * Extended look-ahead set of upcoming 2Q gates.
+        * Cost = sum distances(front) + beta * sum distances(lookahead),
+          with a light decay penalty on recently swapped physical qubits.
+    - Runs several randomized trials and keeps the best layout.
+
+    Parameters
+    ----------
+    coupling : PyGraph
+        Undirected coupling graph whose node indices represent *physical* qubits.
+        Node labels need not form a contiguous range. Edges indicate allowed 2Q interactions.
```
Lumorti: Could we allow a Python list of edges too? Why does the user need to use another external library just to express connectivity?

src/qilisdk/digital/circuit_transpiler_passes/sabre_swap_pass.py
Lumorti, line 1: I'm not so familiar with the SABRE stuff, but since these two passes both generate some new circuit, from my side that's fine, since it seems like it's a heuristic. I will leave the actually quality of said heuristic to be tested by our test suite and by the users.

src/qilisdk/digital/circuit_transpiler.py
```diff
@@ -11,36 +11,202 @@
...
+
+if TYPE_CHECKING:
+    from qilisdk.digital import Circuit
+
+
+def _is_topology_graph(topology: list[tuple[int, int]] | PyGraph[int, None]) -> TypeGuard[PyGraph[int, None]]:
+    return isinstance(topology, PyGraph) and all(isinstance(node, int) for node in topology.nodes())
+
+
+LayoutMap: TypeAlias = dict[int, int]
```
Lumorti: I think this wasn't used inside the passes, they used directly dict[int, int] themselves, maybe everything should use this name so that if later we want to change the map type (e.g. you want to move to rustworkx or something else) we can change it in one place. Same with the topology structure, maybe we should have something like:
`Topology: TypeAlias = PyGraph | list[tuple[int, int]]`
and then use that everywhere.

src/qilisdk/digital/circuit_transpiler.py
```diff
@@ -11,36 +11,202 @@
...
+        qubit_mapping: dict[int, int] | None = None,
+    ) -> Self:
+        if topology is None:
+            return cls(
+                [
+                    DecomposeMultiControlledGatesPass(),
+                    CancelIdentityPairsPass(),
+                    DecomposeToCanonicalBasisPass(
+                        single_qubit_basis=single_qubit_basis, two_qubit_basis=two_qubit_basis
+                    ),
```
Lumorti, line 158: There could be a case like U3 RZ being decomposed to the RX/RY/RZ basis so then we would have RX RY RZ RZ and maybe we could remove that last RZ (assuming the angles allow it).

On the other hand if these two passes were swapped, we could have U3 U3 which would decompose to RX RY RZ RX RY RZ and thus the identity wouldn't be detected, at least not until the fusing afterwards.

Or, what if we have Toffoli(0,1,2) Toffoli(0,1,2)? Right now that wouldn't be cancelled because they would get decomposed first.

So maybe the default stack should be:

CancelIdentityPairsPass (catches Toffoli(0,1,2) Toffoli(0,1,2))
DecomposeMultiControlledGatesPass
CancelIdentityPairsPass (catches X X)
DecomposeToCanonicalBasisPass
CancelIdentityPairsPass (catches RX RY RZ RZ)
FuseSingleQubitGatesPass
CancelIdentityPairsPass (catches U3 = I)

Or, more of a general idea, should we keep repeating all passes until no changes occur? Have like a "max_repeats" and then we run through the stack until self-consistent (or max_repeats reached). That way the user doesn't have to worry so much about the ordering of their passes.

Lumorti (changes requested):
Very nice, this is going to be super useful for users. And it will make it very easy if we have to create new passes later, I like the structure.

Just a few general comments, the only crucial ones I would say are the ones about allowing a list of edges and the one about the ordering of the passes.

src/qilisdk/digital/circuit_transpiler_passes/cancel_identity_pairs_pass.py
fedonman: I don't think we can do parameter sharing, since the circuit is transpiled. E.g. an RX(0, theta) might get transpiled into U3(theta, phi, gamma) or a CNOT might be trasnpiled into a U3-CZ-U3 introducing new parameters.

src/qilisdk/digital/circuit_transpiler.py
```diff
@@ -11,36 +11,202 @@
...
+    information, a richer hardware-aware pipeline can be built through :meth:`default`.
 
     Args:
-        pipeline (list[CircuitTranspilerPass] | None): Sequential list of passes to execute while transpiling.
+        pipeline (list[CircuitTranspilerPass]): Sequential list of passes to execute while transpiling.
     """
 
-    def __init__(self, pipeline: list[CircuitTranspilerPass] | None = None) -> None:
-        self._pipeline = pipeline or [DecomposeMultiControlledGatesPass()]
+    def __init__(self, pipeline: list[CircuitTranspilerPass]) -> None:
```
amirazzam2000, line 110: maybe we should allow topology and qubit_mapping parameters here instead of adding a Custom_layout_pass to be consistent with the interface of the default. and also it's much more intuitive.

src/qilisdk/digital/circuit_transpiler.py
```diff
@@ -11,36 +11,202 @@
...
+                    DecomposeToCanonicalBasisPass(
+                        single_qubit_basis=single_qubit_basis, two_qubit_basis=two_qubit_basis
+                    ),
+                    FuseSingleQubitGatesPass(single_qubit_basis=single_qubit_basis),
+                ]
+            )
 
-    def transpile(self, circuit: Circuit) -> Circuit:
-        """Run the configured pass pipeline over the provided circuit.
+        topology = CircuitTranspiler._build_topology_graph(topology)  # ty:ignore[invalid-argument-type]
```
amirazzam2000: where is this topology coming from?

src/qilisdk/digital/circuit_transpiler.py
```diff
@@ -11,36 +11,202 @@
...
+
+    @classmethod
+    def default(
+        cls,
+        single_qubit_basis: SingleQubitGateBasis = SingleQubitGateBasis.U3,
+        two_qubit_basis: TwoQubitGateBasis = TwoQubitGateBasis.CNOT,
+        topology: list[tuple[int, int]] | PyGraph[int, None] | None = None,
+        qubit_mapping: dict[int, int] | None = None,
+    ) -> Self:
+        if topology is None:
```
amirazzam2000, line 151: here we are not providing a path with topology, how is this parameter being stored or processed? because it seems like now it's not being used at all but the docs suggest that it is being used

src/qilisdk/digital/circuit_transpiler.py
fedonman: I agree, there is a lot of room for improvement and for adding new passes. For now, I would focus on the validity of the current implementation—making sure it does not transpile to something invalid—rather than on catching all cases. If there is a residual XX at the end of the compilation for a special input case, that is acceptable for the first version.

src/qilisdk/digital/circuit_transpiler.py
fedonman: It is used further down, if the user has provided it. I edited the docs to make it clearer.

src/qilisdk/digital/circuit_transpiler.py
fedonman: what do you mean? This describes the connected qubits, and is given as argument to the method.

src/qilisdk/digital/circuit_transpiler.py
fedonman: Changed the relevant passes to use `LayoutMap` instead.

src/qilisdk/digital/circuit_transpiler.py
fedonman: The idea is that user can choose the exact pipeline he likes, if we add other parameters and enforce specific ordering we will restrict this freedom. Also, other layout/routing passes might be added in the future.

src/qilisdk/digital/circuit_transpiler_passes/fuse_single_qubit_gates_pass.py
fedonman: Have you tested your case? It returns correct results for me.

src/qilisdk/digital/circuit_transpiler_passes/decompose_to_canonical_basis_pass.py
fedonman: Done.

src/qilisdk/digital/circuit_transpiler_passes/sabre_layout_pass.py
fedonman: Changed it.

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
fedonman: Changed it.

pyproject.toml
```diff
@@ -34,6 +34,7 @@ dependencies = [
     "loguru>=0.7.3",
     "matplotlib>=3.10.8",
     "mkl-service>=2.6.1; sys_platform != 'darwin'",
+    "rustworkx>=0.17.1",
```
DZPM, line 40: Should we mention somewhere that we're using qiskit's rustworkx?

src/qilisdk/digital/circuit_transpiler.py
amirazzam2000: but then what happens if the user provides more than one routing pass? should we make the routing pass separate from the pipeline?

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
fedonman: I would keep it explicit for now, and we can re-think it based on usage. For the user it's easy to create a default dict anyway `{q: q for q in range(nqubits)}` and change the elements he likes.

pyproject.toml
fedonman: I don't think it should get a special mention. The project started from Qiskit, but is open source now and independent. But I can mention it in the changelog.

src/qilisdk/digital/circuit_transpiler.py
fedonman: It cannot be separate, the routing is part of the pipeline. User might choose any position within the pipeline to place the routing. E.g. I might want to do `CancelIdentityGates -> CustomLayout` or `CustomLayout -> CancelIdentityGates`.

fedonman: We could add some validation checks in the future, to disallow pipelines that produce known errors. E.g. adding two or more routings passes.

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
Lumorti: Fair!

src/qilisdk/digital/circuit_transpiler_passes/custom_layout_pass.py
Lumorti: Thanks 🙌

src/qilisdk/digital/circuit_transpiler.py
Lumorti: Completely fair!

src/qilisdk/digital/circuit_transpiler.py
amirazzam2000: Ahhh makes sense! okay i still feel like we need to do some sort of check if the user applies two layout to raise a warning or an error. Because it does seem like it can be confusing.

Lumorti (approved):
Nice work!

amirazzam2000 (approved):
thanks vyron!! 🙌

DZPM (approved):
LGTM

PR #165, opened 2026-03-03
updating the qilisim interface

amirazzam2000:
Refactored the QiliSim backend configuration API around typed Pydantic models and aligned docs/tests with the new usage.

Main changes:

* Introduced structured QiliSim configuration models:
  * `AnalogMethod` (with `integrator`, `arnoldi`, and `direct` builders)
  * `DigitalMethod` (with `state_vector` builder)
  * `ExecutionConfig` (threading and RNG seed controls)
  * `MonteCarloConfig` (trajectory control for Monte Carlo mode)
* Updated `QiliSim` initialization to accept:
  * `analog_simulation_method`
  * `digital_simulation_method`
  * `execution_config`
  instead of many flat constructor parameters.
* Added `QiliSim.get_config()` and kept `solver_params` as a backward-compatible alias.
* Fixed solver configuration wiring so execution paths use the refactored internal config dictionary.
* Expanded docstrings for new and affected methods/classes, including:
  * per-field `Field(description=...)` metadata for Pydantic model fields
  * top-level model parameter documentation for better IDE hover help
  * classmethod argument documentation aligned with field descriptions.
* Updated backend documentation (`docs/fundamentals/backends.rst`) to describe the new configuration objects and provide an updated configuration example.
* Updated and expanded QiliSim unit/integration tests to validate:
  * new configuration builders and validation behavior
  * new backend construction patterns using `ExecutionConfig`/`AnalogMethod`/`DigitalMethod`/`MonteCarloConfig`
  * compatibility of shared backend integration suites with the new API.

### DIFF
--- a/docs/fundamentals/backends.rst
+++ b/docs/fundamentals/backends.rst
@@ -123,25 +123,24 @@ There is no need to install QiliSim separately, as it is included with the core 
 **Parameters**
 
 - ``noise_model`` (:class:`~qilisdk.noise.noise_model.NoiseModel`, optional): Noise model applied during simulation.
-- ``analog_simulation_method`` (:class:`~qilisdk.backends.qilisim.AnalogMethod`, optional): Analog simulation method and method-specific options.
-- ``digital_simulation_method`` (:class:`~qilisdk.backends.qilisim.DigitalMethod`, optional): Digital simulation options.
-- ``execution_config`` (:class:`~qilisdk.backends.qilisim.ExecutionConfig`, optional): Runtime execution options such as thread count and random seed.
+- ``analog_simulation_method`` (:class:`~qilisdk.backends.backend_config.AnalogMethod`, optional): Analog simulation method and method-specific options.
+- ``digital_simulation_method`` (:class:`~qilisdk.backends.backend_config.DigitalMethod`, optional): Digital simulation options.
+- ``execution_config`` (:class:`~qilisdk.backends.backend_config.ExecutionConfig`, optional): Runtime execution options such as thread count and random seed.
 
 **Configuration example**
 
 .. code-block:: python
 
     from qilisdk.backends import QiliSim
-    from qilisdk.backends.qilisim import AnalogMethod, DigitalMethod, ExecutionConfig, MonteCarloConfig
+    from qilisdk.backends.backend_config import AnalogMethod, DigitalMethod, ExecutionConfig, MonteCarloConfig
 
     backend = QiliSim(
         analog_simulation_method=AnalogMethod.arnoldi(
             dim=16,
-            num_substeps=2,
-            monte_carlo=MonteCarloConfig(trajectories=200),
+            num_substeps=2
         ),
         digital_simulation_method=DigitalMethod.state_vector(max_cache_size=2_000),
-        execution_config=ExecutionConfig(num_threads=4, seed=42),
+        execution_config=ExecutionConfig(num_threads=4, seed=42, monte_carlo=MonteCarloConfig(trajectories=200)),
     )
 
 **Example**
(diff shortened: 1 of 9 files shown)


docs/fundamentals/backends.rst
Lumorti, line 1: The monte carlo settings should probably be in execution settings. They are supported for all methods everywhere, and right now it makes it seem like you could have different monte carlo settings for different simulation types, like the user could set a different monte carlo config for sampling and for time evolution but the first of the these would be overwritten by the latter.

src/qilisdk/backends/qilisim.py
Lumorti, line 1: Could the various configs be moved to a different file? File could be called "backend_configs.py" or something.

Also maybe they shouldn't all inherit from something called QiliSimConfig, but maybe BackendConfig or SimulatorConfig, since it could be that we do something similar to this for CUDA/QuTiP.

src/qilisdk/backends/qilisim.py
```diff
@@ -31,120 +35,369 @@
...
+            DigitalMethod: Configured state-vector digital configuration.
+        """
+        return cls(max_cache_size=max_cache_size)
+
+
 class QiliSim(Backend):
     """
-    Backend based that runs both digital-circuit sampling and analog
+    Backend that runs both digital-circuit sampling and analog
     time-evolution experiments using a custom C++ simulator.
```
Lumorti: Maybe we could put an example here now, since it's a bit trickier to init

tests/integration/backends/test_all_backend_integration.py
```diff
@@ -59,14 +59,14 @@
 
 from qilisdk.backends.cuda_backend import CudaBackend
 
-backends = [QutipBackend(), QiliSim(seed=42, num_threads=1)]
+backends = [QutipBackend(), QiliSim(execution_config=ExecutionConfig(seed=42, num_threads=1))]
```
Lumorti, line 63: God it looks so ugly now. I accept it, but it's still ugly.

Did you say it's still possible to still use the old syntax? That pydantic would move seed and num_threads into the execution config if specified at the top level?

Alternatively, why can't the execution config options be at the top level, again? I understand the logic behind the methods having their own things, but for global options like seed I feel like having it be at the top level isn't confusing and just saves users having to make ExecutionConfigs all of the time.

tests/integration/backends/test_qilisim_specific_integration.py
```diff
@@ -30,9 +30,22 @@
 simulation_types = ["direct", "arnoldi", "integrate"]
 
 
+def _execution(seed: int | None) -> ExecutionConfig:
+    return ExecutionConfig(seed=seed, num_threads=1)
```
Lumorti: Ah see, you've already started abbreviating it huh

Lumorti (changes requested):
Thanks for handling this. I left a few comments.

I do overall think the execution config is too verbose, since adding an extra data object doesn't really serve any purpose other than making the user type more. The digital/analog method configs I totally get and agree with. But the execution config feels like it doesn't do anything. We can have the validation one layer up.

tests/integration/backends/test_all_backend_integration.py
amirazzam2000: well how often do users use these settings? if it's very frequent then yes we can move them but to me they seem like a specific operations that make sense to go inside a separate configuration class. Ideally the user wouldn't need to touch the execution settings unless they have a specific thing they want to do.

Lumorti (approved):
I still think execution_config=ExecutionConfig() is annoyingly long, but I accept it, since you're right in that it's basically only going to be us that uses it.

PR #167, opened 2026-03-03
[SDK-207] Matrix Free Methods

Lumorti:
Matrix-free methods have been added to QiliSim. This is now the default simulation method for circuit sampling, so without changing your workflow you should notice performance and memory improvements.


src/qilisdk/backends/backend_config.py
```diff
@@ -125,16 +125,20 @@ def get_config(self) -> SolverConfigDict:
         return d
 
     @classmethod
-    def integrator(cls, *, num_substeps: int = 2, monte_carlo: MonteCarloConfig | None = None) -> AnalogMethod:
+    def integrator(
+        cls, *, num_substeps: int = 2, matrix_free: bool = False, monte_carlo: MonteCarloConfig | None = None
```
amirazzam2000: could you remove the monte_carlo parameter from here, we must have forgotten to remove it in the previous pr

src/qilisdk/backends/backend_config.py
amirazzam2000: here as well

src/qilisdk/functionals/sampling_result.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, nshots: int, samples: dict[str, int]) -> None:
         """
         self._nshots = nshots
         self._samples = samples
+        self._state = None
```
amirazzam2000: make sure you remove this
```suggestion
```

src/qilisdk_cpp/backends/qilisim/analog/iterations.h
```diff
@@ -15,10 +15,12 @@
 
 #include <vector>
 #include "../libs/eigen.h"
+#include "../representations/matrix_free_hamiltonian.h"
 
-void arnoldi(const SparseMatrix& L, const SparseMatrix& v0, int m, std::vector<SparseMatrix>& V, SparseMatrix& H);
-void arnoldi_mat(const SparseMatrix& Hsys, const SparseMatrix& rho0, int m, std::vector<SparseMatrix>& V, SparseMatrix& Hk);
+void arnoldi(const SparseMatrix& L, const DenseMatrix& v0, int m, std::vector<DenseMatrix>& V, SparseMatrix& H);
```
amirazzam2000, line 20: why are we forcing all states to be dense now? I thought this would only be the case if matrix free is true no?

amirazzam2000 (changes requested):
great job! just some minor comments

src/qilisdk/backends/backend_config.py
```diff
@@ -252,21 +256,51 @@ class DigitalMethod(BaseSimulatorConfig):
...
         description="Maximum number of cached gate representations used by the digital simulator.",
     )
+    normalize_after_each_gate: bool = Field(
+        default=True,
+        description="Whether to normalize the statevector after each gate application to mitigate numerical errors at the cost of increased runtime.",
+    )
+    combine_single_qubit_gates: bool = Field(
+        default=True,
+        description="Whether to combine consecutive single-qubit gates into a single operation to reduce overhead at the cost of increased memory usage.",
+    )
```
fedonman, line 263: This is done by the transpiler if the user opts-in. I think the simulator should simulate the input circuit as-is without doing any changes.

src/qilisdk_cpp/backends/qilisim/digital/circuit_optimization.cpp
```diff
@@ -0,0 +1,93 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "circuit_optimizations.h"
+
+#include <iostream>  // TODO(luke) remove
+
+std::vector<Gate> combine_single_qubit_gates(const std::vector<Gate>& gates) {
```
fedonman, line 19: The previous comment is relevant here. I think we should leave optimizations to whatever is defined in the Transpiler, and QiliSim’s responsibility should be the simulation.

There is also a subtle bug introduced here: if the user provides a circuit with N consecutive single-qubit gates and a noise model with gate errors, QiliSim combines the gates into one, but the noise is also applied only once, whereas it should be applied N times.

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -75,30 +144,36 @@ void sampling(const std::vector<Gate>& gates,
...
-    bool monte_carlo = (!is_statevector && config.get_monte_carlo() && !has_noise);
+    bool monte_carlo = (!is_statevector && config.get_monte_carlo());
     if (monte_carlo) {
-        state = sample_from_density_matrix(state.sparseView(), config.get_num_monte_carlo_trajectories(), config.get_seed());
+        state = sample_from_density_matrix(state, config.get_num_monte_carlo_trajectories(), config.get_seed());
+    }
+
+    // Combine single-qubit gates for speed if not using noise models
+    std::vector<Gate> optimized_gates = gates;
+    if (!has_noise && config.get_combine_single_qubit_gates()) {
```
fedonman, line 169: Ah! Just saw you check this case. Nice. Rest of the comment holds though.

fedonman (changes requested):
Great addition to qilisim's features. Left a minor comment to discuss.

Something important: I just noticed we have zero tests for the C++ implementations. We should change this ASAP.

src/qilisdk_cpp/backends/qilisim/representations/matrix_free_hamiltonian.cpp
```diff
@@ -0,0 +1,198 @@
+// Copyright 2025 Qilimanjaro Quantum Tech
```
DZPM: 2026

```suggestion
// Copyright 2026 Qilimanjaro Quantum Tech
```

src/qilisdk_cpp/backends/qilisim/representations/matrix_free_hamiltonian.h
```diff
@@ -0,0 +1,38 @@
+// Copyright 2025 Qilimanjaro Quantum Tech
```
DZPM:
```suggestion
// Copyright 2026 Qilimanjaro Quantum Tech
```

src/qilisdk_cpp/backends/qilisim/representations/matrix_free_operator.h
```diff
@@ -0,0 +1,41 @@
+// Copyright 2025 Qilimanjaro Quantum Tech
```
DZPM:
```suggestion
// Copyright 2026 Qilimanjaro Quantum Tech
```

src/qilisdk/backends/backend_config.py
Lumorti: So originally I did this, then we removed it, then in some daily we said probably we should have it, since it's only some lines of code and there is no reason not to have it if the user doesn't use noise models. If the user enables noise models, this is disabled. It's purely a simulation implementation feature, maybe inside the CUDA/QuTiP backends they do something similar.

src/qilisdk_cpp/backends/qilisim/analog/iterations.h
Lumorti: In general for time evolution the states become dense very quickly, so starting dense speeds things up.

Lumorti:
> Something important: I just noticed we have zero tests for the C++ implementations. We should change this ASAP.

Yeah it's a good point, it's on my list of things to do, I'll make a linear task for it now.

src/qilisdk/backends/backend_config.py
amirazzam2000: I agree with luke here, this is not necessarily a transpilation step, it's a part of the simulation. because he is just combining single qubit gates into an equivalent gate to speed up the application of these gates. Which is just a simulation improvement in case there are no noise models and I don't see why wouldn't this be enabled by default.

amirazzam2000 (approved):
it looks good from my side! Thanks Luke 🙌

src/qilisdk/backends/backend_config.py
fedonman: OK. Makes sense.

fedonman (approved):
Great feature! 💯 💯

PR #168, opened 2026-03-04
[SDK-238] QTensor To C++

Lumorti:
QTensor has been moved to C++, giving increased performance. 

The interface remains almost exactly the same, the only change being that now `QTensor * QTensor` does element-wise multiplication, for matrix multiplication please change to using `QTensor @ QTensor`.


src/qilisdk/core/qtensor.py
```diff
@@ -81,48 +56,35 @@ class QTensor:
...
+        elif isinstance(other, np.ndarray):
+            if other.dtype != np.complex128:
+                other = other.astype(np.complex128)
+            self._qtensor_cpp = QTensorCpp(other)
         else:
-            self._nqubits = int(np.log2(r))
+            self._qtensor_cpp = QTensorCpp(other)
+
+    @property
+    def qtensor_cpp(self) -> QTensorCpp:
```
amirazzam2000: I don't know what is the benefit of having this. The python wrapper is meant to manage this object, accessing the object itself won't add any benefit and will be confusing for some users i feel.

src/qilisdk/core/qtensor.py
```diff
@@ -189,247 +188,223 @@ def adjoint(self) -> QTensor:
...
+        Compute the transpose of the QTensor.
+
+        Returns:
+            QTensor: A new QTensor that is the transpose of this object.
+        """
+        return QTensor(self._qtensor_cpp.transpose())
+
+    def dagger(self) -> QTensor:
+        """
+        Overload for adjoint() to provide a more familiar method name for quantum objects.
```
amirazzam2000: maybe say alias instead of overload because you don't modify the functionality.

src/qilisdk/core/qtensor.py
```diff
@@ -189,247 +188,223 @@ def adjoint(self) -> QTensor:
...
-
-        raise ValueError("The QTensor is not a valid state or operator for ptrace().")
-
-    def norm(self, order: int | Literal["fro", "tr"] = 1) -> float:
+            ValueError: If the indices in 'keep' are out of range.
+
+        Returns:
+            QTensor: A new QTensor representing the reduced density matrix.
+        """
+        return QTensor(self._qtensor_cpp.partial_trace_python(list(keep)))
```
amirazzam2000: I would confirm that the passed object is a set if having duplicate indices breaks your c++ code, because in python this is not checked automatically. So like type hinting is just for the looks not enforced in code at all.

src/qilisdk/core/qtensor.py
```diff
@@ -189,247 +188,223 @@ def adjoint(self) -> QTensor:
...
+            order (Literal["auto", "frobenius", "trace", "l1", "l2", "inf", "nuclear"], optional): The order of the norm. Defaults to "auto".
+
+        Returns:
+            float: The computed norm of the QTensor.
 
         Raises:
-            ValueError: If the QTensor is not a valid density matrix or state vector,
+            ValueError: If an unsupported norm order is specified.
+        """
+        if isinstance(order, int):
```
amirazzam2000, line 328: maybe check that the int is within the accepted range?

amirazzam2000: also in this case you can throw an error if they pass an int if you are expecting a literal

src/qilisdk/core/qtensor.py
```diff
@@ -189,247 +188,223 @@ def adjoint(self) -> QTensor:
...
+            return self._qtensor_cpp.norm("l" + str(order))
+        if order == "tr":
+            return self._qtensor_cpp.norm("nuclear")
+        return self._qtensor_cpp.norm(order)
+
+    def compute_eigendecomposition(self) -> None:
```
amirazzam2000: why are we returning None here? maybe we should return the eigen values and vectors?

src/qilisdk/core/qtensor.py
```diff
@@ -438,350 +413,416 @@ def to_density_matrix(self, max_relative_correction: float = _MAX_STATE_CORRECTI
...
-            return False
-        diff = self._data - self._data.getH()
-        if diff.nnz == 0:
-            return True
-        return float(scipy_norm(diff, ord="fro")) <= tol
+        return QTensor(QTensorCpp.bra_python(state))
 
-    # ----------- Basic Arithmetic Operators ------------
+    @classmethod
+    def zero(cls, rows: int, cols: int) -> QTensor:
```
amirazzam2000: I wouldn't specify rows and cols, i would specify nqubits and type (ket, bra, density_matrix) because otherwise this could imply that you can have a 2x3 matrix or something like that

src/qilisdk/core/qtensor.py
```diff
@@ -438,350 +413,416 @@ def to_density_matrix(self, max_relative_correction: float = _MAX_STATE_CORRECTI
...
+            QTensor: A new QTensor representing the zero matrix.
+        """
+        return QTensor(QTensorCpp.zero(rows, cols))
 
-    def __sub__(self, other: QTensor) -> QTensor:
-        if isinstance(other, QTensor):
-            return QTensor(self._data - other._data)
-        return NotImplemented
+    @classmethod
+    def identity(cls, dim: int) -> QTensor:
```
amirazzam2000: similarly here i wouldn't specify dim i would specify nqubits and always generate a 2**nqubits matrix.

src/qilisdk/core/qtensor.py
```diff
@@ -438,350 +413,416 @@ def to_density_matrix(self, max_relative_correction: float = _MAX_STATE_CORRECTI
...
+        Returns:
+            list[QTensor]: A list of QTensor objects representing the eigenvectors of the QTensor.
+        """
+        return [QTensor(vec) for vec in self._qtensor_cpp.get_eigenvectors_python()]
+
+    def expectation_value(self, other: QTensor, nshots: int = 0) -> complex:
+        """
+        Compute the expectation value of another QTensor with respect to this QTensor.
+
+        For state vectors, this corresponds to ⟨psi|O|psi⟩. For operators, it corresponds to Tr(rho O).
```
amirazzam2000, line 738:
```suggestion
        For state vectors, this corresponds to ⟨psi|O|psi⟩. For density matrices, it corresponds to Tr(rho O).
```

src/qilisdk/core/qtensor.py
```diff
@@ -438,350 +413,416 @@ def to_density_matrix(self, max_relative_correction: float = _MAX_STATE_CORRECTI
...
+        QTensor: A QTensor representing the \|n⟩'th basis state on a N-size Hilbert space (N=2**num_qubits).
+    """
+    if not (0 <= n < N):
+        raise ValueError(f"n must be in [0, {N - 1}]")
+    # one nonzero at (row=n, col=0), value=1.0
+    mat = csr_matrix(([1.0], ([n], [0])), shape=(N, 1))
+    return QTensor(mat)
+
+
+def identity(dim: int) -> QTensor:
```
amirazzam2000: same as above

amirazzam2000 (changes requested):

CMakeLists.txt
```diff
@@ -6,79 +9,87 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE INTERNAL "")
...
+    set(EIGEN_BUILD_TESTING OFF)
+    set(EIGEN_BUILD_PKGCONFIG OFF)
+    set(EIGEN_BUILD_DOC OFF)
+    include(FetchContent)
+    FetchContent_Declare(
+        eigen
+        GIT_REPOSITORY https://gitlab.com/libeigen/eigen.git
+        GIT_TAG 3.4.0
+        GIT_SHALLOW TRUE
+    )
```
fedonman, line 39: Not exclusive to this PR, but why do we choose the `3.4.0` version which is 5 years old, and not a later one? Even if we don't want to update to the `5.X` line because of breaking changes, there is the `3.4.1` release with lots of bug fixes. https://gitlab.com/libeigen/eigen/-/releases

Lumorti: Very good spot! I was always using libeigen3 because that's the one that's included in apt. But I will try 5 and make sure it doesn't break anything, maybe we can even get a speed-up

Lumorti: Upgrade to 5.0.1, no difference in the speed of the circuit simulation, from what I read online it's mostly just new things, the interface is still the same and the code compiles without problems.

docs/fundamentals/noise_models.rst
```diff
@@ -89,17 +89,17 @@ Noise Config
...
 .. code-block:: python
 
     from qilisdk.noise import NoiseModel, NoiseConfig
     from qilisdk.digital import X
 
-    # Create a noise configuration, setting the X gate duration to 20 ns
+    # Create a noise configuration, setting the X gate time to 20 ns
     conf = NoiseConfig()
-    conf.set_gate_duration[X] = 20e-9
+    conf.set_gate_time(X, 20e-9)
```
fedonman, line 102: Did we rename this in this PR or is it just a documentation correction? I prefer the `duration` because it is consistent with QProgram.

src/qilisdk/core/qtensor.py
```diff
@@ -81,48 +57,19 @@ class QTensor:
             density = ket * ket.adjoint()
     """
 
-    _MAX_STATE_CORRECTION = 1e-2
-
-    def __init__(self, data: np.ndarray | sparray | spmatrix) -> None:
+    def __init__(self, other: np.ndarray | sparray | spmatrix | list[list[Number]] | QTensor | QTensorCpp) -> None:
```
fedonman: I would remove the extra accepted types, because they should not be part of the public interface.

* `list[list[Number]]`: does anyone actually use this format?
* `QTensor`: it feels odd, and arguably un-Pythonic, for a class constructor to accept an instance of the same class. If the user needs a copy, they can call `copy()` or `deepcopy()`.
* `QTensorCpp`: I understand that this is used extensively internally, but will the user ever need it? They should not work with the C++ object directly. We could turn this into a private class method instead. For example:

```python
@classmethod
def _from_cpp(cls, qtensor_cpp: QTensorCpp) -> Self:
    obj = cls.__new__(cls)
    obj._qtensor_cpp = qtensor_cpp
    return obj
```

Then, instead of:

```python
return QTensor(QTensorCpp.identity(nqubits))
```

we would do:

```python
return cls._from_cpp(QTensorCpp.identity(nqubits))
```

src/qilisdk/core/qtensor.py
```diff
@@ -438,350 +402,420 @@ def to_density_matrix(self, max_relative_correction: float = _MAX_STATE_CORRECTI
...
+
+    def __sub__(self, other: QTensor) -> QTensor:
+        return QTensor(self._qtensor_cpp.sub_python(other))
 
-        tr = float(np.real(rho.trace()))
-        if abs(tr) < get_settings().atol:
-            raise ValueError("Cannot normalize density matrix with zero trace.")
-        return QTensor(rho.data / tr)  # keep it sparse
+    def __mul__(self, other: QTensor | Number) -> QTensor:
+        return QTensor(self._qtensor_cpp.mul_python(other))
```
fedonman, line 463: AI review:

This changes QTensor * QTensor from matrix multiplication to element-wise multiplication, which is a breaking API regression. For example, ket(0) * ket(0).adjoint() now returns a (1, 2) tensor instead of a (2, 2) density matrix. Do we want this or is a bug?

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -0,0 +1,2017 @@
...
+                        int col = int(it.col());
+                        std::complex<double> val = eval_inv * it.value();
+                        triplets.emplace_back(row, col, val);
+                    }
+                }
+            }
+        }
+        result.setFromTriplets(triplets.begin(), triplets.end());
+        return QTensorCpp(result);
+    }
```
fedonman: AI Review:

The cached eigendecomposition branch in inverse() is not reconstructing `V D^{-1} V^{-1}`. It only scales the eigenvectors and writes them into the output, which produces an incorrect inverse once eig() has been called. I verified this with diag(2, 3), which returns [[0.5, 0], [0.333..., 0]] instead of diag(0.5, 1/3).

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -0,0 +1,2017 @@
...
+        if (std::abs(val) > 0.0) {
+            triplets.emplace_back(idx.first, idx.second, val);
+        }
+    }
+    Eigen::SparseMatrix<std::complex<double>> result(keep_dim, keep_dim);
+    result.setFromTriplets(triplets.begin(), triplets.end());
+    return QTensorCpp(result);
+}
+
+double QTensorCpp::norm(const std::string& norm_type) {
```
fedonman, line 626: AI review:

The norm implementation has changed semantics. "trace" now returns the literal trace, and "nuclear" is computed from absolute eigenvalues rather than singular values. Both are incorrect for general non-Hermitian operators. For example, [[0, 1], [0, 0]] should have trace norm 1, not 0.

tests/unit/core/test_qtensor.py
```diff
@@ -605,27 +595,13 @@ def test_reset_qubits_requires_density_matrix():
...
 
-def test_large_trace_norm():
-    qket = ket(*([0 for _ in range(21)]))
-    rho = qket.to_density_matrix()
-    rho._data[0, 0] = 2.0
-    with pytest.raises(ValueError, match="Trace norm for large"):
-        rho.norm(order="tr")
-
-
 def test_non_hermitian_trace_norm():
```
fedonman, line 599: AI review:

This test appears to codify the incorrect norm behavior introduced in the C++ port. For [[0, 1], [0, 0]], the trace norm is not the sum of absolute eigenvalues; it is the sum of singular values, which is 1. I do not think we should merge a test that locks in the wrong definition.

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -0,0 +1,2017 @@
...
+    } else {
+        throw py::value_error("Data object must be a QTensor, a list of lists, a scipy sparse matrix, or a numpy array, got: " + std::string(py::str(data)));
+    }
+    if (_data.rows() == 0 || _data.cols() == 0) {
+        throw py::value_error("A QTensor should be initialized with a non-empty 2D array; got shape (" + std::to_string(_data.rows()) + ", " + std::to_string(_data.cols()) + ")");
+    }
+    if (((_data.rows() & (_data.rows() - 1)) != 0) || ((_data.cols() & (_data.cols() - 1)) != 0)) {
+        throw py::value_error("A QTensor should have dimensions that are powers of 2; got shape (" + std::to_string(_data.rows()) + ", " + std::to_string(_data.cols()) + ")");
+    }
+}
```
fedonman, line 165: The constructor validation is weaker than the previous Python implementation. It now accepts shapes like (2, 4) and (4, 2), which are neither operators, kets, bras, nor scalars, and the list path can return a (0, 0) tensor. That breaks the documented QTensor invariants and should be rejected at construction time.

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -0,0 +1,2017 @@
...
+    if ((is_ket() && other.is_ket()) || (is_bra() && other.is_bra())) {
+        return std::pow(std::abs(dot(other.conjugate())), 2);
+    } else if (is_operator() && other.is_operator()) {
+        QTensorCpp sqrt_self = sqrt();
+        std::cout << "sqrt_self:\n" << sqrt_self << std::endl;
+        QTensorCpp product = sqrt_self * other * sqrt_self;
+        std::cout << "product:\n" << product << std::endl;
+        QTensorCpp sqrt_product = product.sqrt();
+        std::cout << "sqrt_product:\n" << sqrt_product << std::endl;
+        return std::pow(sqrt_product.trace().real(), 2);
```
fedonman, line 1749: Remove prints.

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -0,0 +1,2017 @@
...
+    if (py::isinstance<QTensorCpp>(other)) {
+        return dot(other.cast<QTensorCpp>());
+    } else if (py::hasattr(other, "_qtensor_cpp")) {
+        return dot(other.attr("_qtensor_cpp").cast<QTensorCpp>());
+    } else {
+        throw py::value_error("Other object must be a QTensor");
+    }
+}
+
+#include <iostream>  // TODO remove
```
fedonman: Remove this.

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -0,0 +1,2017 @@
...
+    Args:
+        other (QTensorCpp): The other QTensor to compute the fidelity with respect to this QTensor.
+
+    Returns:
+        double: The fidelity between this QTensor and the other QTensor.
+
+    Raises:
+        py::value_error: If the two QTensors are not of the same type (both kets, both bras, or both operators).
+    */
+    if ((is_ket() && other.is_ket()) || (is_bra() && other.is_bra())) {
```
fedonman, line 1743: AI review:

The ket/bra fidelity path is incorrect for complex states because it conjugates other before calling dot(), even though dot() already conjugates self. Two orthogonal states such as ([1, i] / sqrt(2)) and ([1, -i] / sqrt(2)) currently report fidelity ~1 instead of 0.

fedonman (changes requested):
Great work! Left some comments from me, and also some from AI. Ignore any that is false positive.

src/qilisdk/core/qtensor.py
amirazzam2000: for the list[list[Number]] I have to say that it's very useful. I keep using it by default then having to translate that into a numpy array to pass it which is very annoying.

docs/fundamentals/noise_models.rst
Lumorti: It's a documentation correction, I was making a Jupyter notebook for Ameer and noticed I got an error running the code from the current docs, I also agree I prefer gate duration

src/qilisdk/core/qtensor.py
Lumorti: I think the list[list[Number]] is useful, but it's true I can't really see a use for QTensor or QTensorCpp, I mostly just put them because I made the C++ side quite general, will remove

src/qilisdk/core/qtensor.py
Lumorti: Was QTensor * QTensor not element-wise before? I thought in Python generally people use the QTensor @ QTensor for matrix multiplication

src/qilisdk_cpp/core/qtensor.cpp
Lumorti: Added checks for this

src/qilisdk_cpp/core/qtensor.cpp
Lumorti: So I wanted to distinguish between the trace and trace norm (also called the nuclear norm). Because sometimes (in the tests and in general) we want to normalize a matrix by its trace (quick to calculate) and not its trace/nuclear norm (requires a full eigendecomposition). So it's true that norm("trace") should probably mean the nuclear norm, whilst normalized("trace") should probably imply normalizing by dividing by the trace. My solution was to exclusively call the trace norm the nuclear norm everywhere, and have trace be reserved always for the literal trace of the matrix.

The part about the nuclear norm being computed wrong is correct, however, and is now fixed and tested.

tests/unit/core/test_qtensor.py
Lumorti: This is my bad, I misread the definition of the trace norm and assumed there was a mistake in the old test. Fixed!

src/qilisdk_cpp/core/qtensor.cpp
Lumorti: This is a nice find, I'll fix it and add a test

src/qilisdk/core/qtensor.py
amirazzam2000: yes but i think we didn't implement the @ operator we only had the *

src/qilisdk_cpp/core/qtensor.cpp
Lumorti: Good catch! Fixed, and added a lot more tests for all of the matrix arithmetic functions

src/qilisdk/core/qtensor.py
Lumorti: So what should * do? I'm fine with defaulting to matrix multiplication if you want

fedonman: For me, it's ok to keep * for element-wise and @ for matmul.

amirazzam2000: yeah I am also okay with adding @ operator just make sure to update the instances in the library where we multiply two QTensors directly (there shouldn't be many)

Lumorti: I tried searching but couldn't find any, and all of the tests pass, can you think of any places in the library where we might use matrix multiplication between QTensors?

amirazzam2000: my main guess is the reservoirs, as we do manual evolution for the circuits.

amirazzam2000: also if you can add it as a note in the changelog so we can communicate it when the release is made for people to update their code.

Lumorti: Done!

amirazzam2000: actually i just checked again, and the behavior with the old code was that * -> element wise and @ -> matrix product: 

```
    def __mul__(self, other: QTensor | Complex) -> QTensor:
        if isinstance(other, (int, float, complex)):
            return QTensor(self._data * other)
        if isinstance(other, QTensor):
            return QTensor(self._data * other._data)
        return NotImplemented

    def __matmul__(self, other: QTensor) -> QTensor:
        if isinstance(other, QTensor):
            return QTensor(self._data @ other._data)
        return NotImplemented
 ```
 
 so the changes luke did should be aligned with what was there before.

amirazzam2000 (approved):
looks good to me 👍 
Thanks luke!!

fedonman (approved):
💯 💯

PR #169, opened 2026-03-09
[SDK-226] Removing the exposed Parameter objects in Gate, Hamiltonian, and Interpolator. 

amirazzam2000:
Refactored parameter handling across digital and analog objects to use the shared `Parameterizable` interface, including explicit `get_*`/`set_*` parameter access patterns and synchronized tests/docs updates for the new structure.

### DIFF
--- a/docs/fundamentals/core.rst
+++ b/docs/fundamentals/core.rst
@@ -198,6 +198,15 @@ such as trainable parameters.
     trainable_values = circuit.get_parameter_values(where=lambda p: p.is_trainable)
     circuit.set_parameter_values([0.3], where=lambda p: p.is_trainable)
 
+.. note::
+
+    Parameterizable objects expose parameter state through
+    :meth:`~qilisdk.core.parameterizable.Parameterizable.get_parameters`,
+    :meth:`~qilisdk.core.parameterizable.Parameterizable.get_parameter_names`,
+    :meth:`~qilisdk.core.parameterizable.Parameterizable.get_parameter_values`,
+    and the corresponding ``set_*`` methods. Accessing ``.parameters`` directly
+    is not part of the public API.
+
 
 Mathematical Maps (sin, cos)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/src/qilisdk/core/parameterizable.py
+++ b/src/qilisdk/core/parameterizable.py
@@ -37,13 +37,13 @@ class Parameterizable(ABC):
         self._parameter_constraints: list[ComparisonTerm] = []
         self._prefix = ""
 
-    @property
-    def nparameters(self) -> int:
-        """Number of tunable parameters defined by the object."""
-        return len(dict(self._iter_parameter_items()))
+    # Private Methods
 
     def _iter_parameter_children(self) -> Iterable[Parameterizable]:  # noqa: PLR6301
-        """Yield parameterizable children to compose this object's parameter interface.
+        """Yield child objects that compose this object's parameter interface.
+
+        Override this method in subclasses that expose nested
+        :class:`Parameterizable` objects.
 
         Returns:
             Iterable[Parameterizable]: Child objects that contribute parameters to this instance.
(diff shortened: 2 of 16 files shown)


src/qilisdk/core/parameterizable.py
```diff
@@ -83,6 +85,48 @@ def get_prefix(self) -> str:
...
+
+    def _add_parameter_from(self, parameter_label: str, other: Parameterizable, new_label: str | None = None) -> None:
+        if new_label:
+            self._parameters[self._prefix + new_label] = other._parameters[parameter_label]
+        else:
+            self._parameters[self._prefix + parameter_label] = other._parameters[parameter_label]
+
+    @staticmethod
+    def _query_parameter_original_name(parameterizable: Parameterizable, label: str) -> str:
+        return parameterizable._parameters[label].label
```
fedonman: This is used only once in Circuit. I wouldn't create a method for it, it's better to just use the underlying code there directly.

src/qilisdk/core/parameterizable.py
fedonman, line 1: For the methods of this file, I would do the following for clarity:
1. Add docstrings to all methods public or private. Especially for the private ones, It should be explained if the developer should override the method or not and how (e.g. `_iter_parameter_children` is supposed to be overridden by subclasses)
2. For the methods that should not be overridden add the `@final` decorator from `typing` module.
3. Move all private / public methods together. (now they are mixed)

src/qilisdk/digital/gates.py
```diff
@@ -49,6 +49,9 @@ class Gate(Parameterizable, ABC):
 
     PARAMETER_NAMES: ClassVar[list[str]] = []
 
+    def __init__(self) -> None:
+        super().__init__()
```
fedonman, line 53: This is not needed.

fedonman (changes requested):
Very nice refactor! Made minor comments. Also CodeCov is reporting missing lines.

src/qilisdk/core/parameterizable.py
amirazzam2000: The reason it's here and not in circuit is because otherwise i will have to access a private member of Gate inside circuit. They are both a Parameterizable so i do it here for ty not to complain. 

I couldn't find any other way of doing it without accessing the private member of Gate (and not making that private member public cuz that would defeat the purpose of the PR)

src/qilisdk/digital/gates.py
amirazzam2000: it is needed, because otherwise the self._parameters/self._parameter_constraints are not created by default, which are both accessed by default by methods of parameterizable. 


I know it's enough to have it on all the children, but i think we should keep it for safety.

src/qilisdk/core/parameterizable.py
fedonman: Why not use the public `.get_parameters()`?

amirazzam2000: because we need to access the actual parameter information. here is the current structure: 

Parameterizable stores:
- String: containing a name it has assigned to the parameter.
- parameter: a reference to the actual parameter. Which contains:
                               - String: a label the user has assigned to the parameter (or we gave by default)
                               
now in all cases the parameter string and the parameterizable string are equal. Except for that one case inside Circuit. This is because in the gates and circuits we complicate things a bit with the pre-generated/automatic labels. so in here we have this method to query the original parameter label not the parameterizable label because in gates, the label is alway theta, regardless of what label the user provided. but in circuit the label the user provided becomes relevant. 

There is some breaking of the parameterizable interface in both gate and circuit. which is why they are the exception to the parameterizable.

docs/fundamentals/core.rst
```diff
@@ -198,6 +198,15 @@ such as trainable parameters.
     trainable_values = circuit.get_parameter_values(where=lambda p: p.is_trainable)
     circuit.set_parameter_values([0.3], where=lambda p: p.is_trainable)
 
+.. note::
+
+    Parameterizable objects expose parameter state through
```
Lumorti, line 203:
```suggestion
    Parameterizable objects expose parameter information through
```

Lumorti (approved):
I like the changes, nice to have everything using the same methods!

src/qilisdk/core/parameterizable.py
fedonman: Got it. Ok let's leave it as it is and we can think the special case of circuit separately.

fedonman (approved):
💯 💯

PR #171, opened 2026-03-11
[SDK-192] Add Repr To Everything

Lumorti:
Almost all QiliSDK objects can now be printed (i.e. print(thing)) to show details about the object.

### DIFF
--- a/src/qilisdk/analog/schedule.py
+++ b/src/qilisdk/analog/schedule.py
@@ -420,3 +420,19 @@ class Schedule(Parameterizable):
             renderer.save(filepath)
         else:
             renderer.show()
+
+    def __repr__(self) -> str:
+        schedule_str = "Schedule(\n"
+        schedule_str += "  hamiltonians={\n"
+        for label, ham in self._hamiltonians.items():
+            schedule_str += f"    '{label}': {ham},\n"
+        schedule_str += "  },\n"
+        schedule_str += "  coefficients={\n"
+        for label, coeff in self.coefficients_dict.items():
+            schedule_str += f"    '{label}': {coeff},\n"
+        schedule_str += "  },\n"
+        schedule_str += f"  dt={self.dt},\n"
+        schedule_str += f"  total_time={self._max_time},\n"
+        schedule_str += f"  interpolation={self._interpolation}\n"
+        schedule_str += ")"
+        return schedule_str
--- a/src/qilisdk/analog/schedule.py
+++ b/src/qilisdk/analog/schedule.py
@@ -422,17 +422,17 @@ class Schedule(Parameterizable):
             renderer.show()
 
     def __repr__(self) -> str:
-        schedule_str = "Schedule(\n"
-        schedule_str += "  hamiltonians={\n"
-        for label, ham in self._hamiltonians.items():
-            schedule_str += f"    '{label}': {ham},\n"
-        schedule_str += "  },\n"
-        schedule_str += "  coefficients={\n"
-        for label, coeff in self.coefficients_dict.items():
-            schedule_str += f"    '{label}': {coeff},\n"
-        schedule_str += "  },\n"
-        schedule_str += f"  dt={self.dt},\n"
-        schedule_str += f"  total_time={self._max_time},\n"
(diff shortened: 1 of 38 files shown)


src/qilisdk/analog/schedule.py
```diff
@@ -420,3 +420,19 @@ def draw(self, style: ScheduleStyle | None = None, filepath: str | None = None)
...
+        schedule_str += "  },\n"
+        schedule_str += "  coefficients={\n"
+        for label, coeff in self.coefficients_dict.items():
+            schedule_str += f"    '{label}': {coeff},\n"
+        schedule_str += "  },\n"
+        schedule_str += f"  dt={self.dt},\n"
+        schedule_str += f"  total_time={self._max_time},\n"
+        schedule_str += f"  interpolation={self._interpolation}\n"
+        schedule_str += ")"
+        return schedule_str
```
fedonman: Don't use the `+=` for string building, especially if loops are concerned, because the concatenation builds a new string every time. Instead, we can pre-calculate the loops with comprehensions, use `.join()` and add them with f-strings.

E.g.

```python
def __repr__(self) -> str:
    hamiltonians = ",\n".join(
        f"    {label!r}: {ham!r}" for label, ham in self._hamiltonians.items()
    )
    coefficients = ",\n".join(
        f"    {label!r}: {coeff!r}" for label, coeff in self.coefficients_dict.items()
    )

    return (
        "Schedule(\n"
        "  hamiltonians={\n"
        f"{hamiltonians}\n"
        "  },\n"
        "  coefficients={\n"
        f"{coefficients}\n"
        "  },\n"
        f"  dt={self.dt!r},\n"
        f"  total_time={self._max_time!r},\n"
        f"  interpolation={self._interpolation!r}\n"
        ")"
    )
```
 
or:

```python
def __repr__(self) -> str:
    lines = [
        "Schedule(",
        "  hamiltonians={",
        *(f"    {label!r}: {ham!r}," for label, ham in self._hamiltonians.items()),
        "  },",
        "  coefficients={",
        *(f"    {label!r}: {coeff!r}," for label, coeff in self.coefficients_dict.items()),
        "  },",
        f"  dt={self.dt!r},",
        f"  total_time={self._max_time!r},",
        f"  interpolation={self._interpolation!r},",
        ")",
    ]
    return "\n".join(lines)
```

fedonman: The `!r` calls the `__repr__`, which is what we want in case an object has both `__str__` and `__repr__`.

fedonman: Another solution that also enables unified styling is to use the `pprint` library. E.g.

```python
from pprint import pformat
from textwrap import indent

def __repr__(self) -> str:
    fields = {
        "hamiltonians": self._hamiltonians,
        "coefficients": self.coefficients_dict,
        "dt": self.dt,
        "total_time": self._max_time,
        "interpolation": self._interpolation,
    }
    body = indent(pformat(fields, sort_dicts=False, width=120), "  ")
    return f"Schedule(\n{body}\n)"
```

The output is not 100% same, but is consistent and does wrapping based on line width as well. I am sure we can further tweak it too.

fedonman (changes requested):
Very useful! 💯 I am requesting to remove the `+=` operator, I commented some alternatives.

src/qilisdk/noise/noise_model.py
```diff
@@ -257,3 +257,34 @@ def _add_parameter_perturbation(
             return
         self._check_scope_allowed(noise, AttachmentScope.PER_GATE_TYPE)
         self.per_gate_perturbations[gate, parameter].append(noise)
+
+    def __repr__(self) -> str:
+        SO_SONARQUBE_DOESNT_COMPLAIN = "  },"
```
fedonman, line 262: 😆 😆 You know, I can mark a SonarQube issue as false positive or acceptable, no need to satisfy it if it doesn't make sense.

fedonman (approved):
Great work 💯 💯

src/qilisdk/analog/schedule.py
Lumorti: Done!

PR #172, opened 2026-03-12
[SDK-245] QiliSim SWAP Matrix Free

Lumorti:
Added support for SWAP, Toffoli and generic controlled single-qubit gates to the matrix free sampler in QiliSim.


Lumorti:
(This was needed because some of the Shor benchmarks that Devanshi wants me to run use controlled U1s and Toffolis)

amirazzam2000 (approved):
thanks luke!

tests/integration/backends/test_all_backend_integration.py
```diff
@@ -546,3 +546,43 @@ def test_execute_quantum_reservoir_qutip(monkeypatch):
...
+
+@pytest.mark.parametrize("backend", backends)
+def test_toffoli(backend):
+    circuit = Circuit(nqubits=3)
+    circuit.add(X(0))
+    circuit.add(X(1))
+    circuit.add(X(2).controlled(0, 1))
+    result = backend.execute(Sampling(circuit=circuit, nshots=10))
+    assert isinstance(result, SamplingResult)
+    assert result.samples == {"111": 10}
```
fedonman, line 588: These worked before this PR though, right? Do we check the actual implementation?

Lumorti: These tests are new, added in this PR. Similar ones existed a while ago, but now are in the unit tests as more of a "does CUDA/QiliSim/QuTiP" _handle_ the gate correctly, but there was never an integration test for these gates since we split the integration/unit tests.

fedonman (approved):
💯 💯

PR #173, opened 2026-03-16
[SDK-243] Add Cpp Tests

Lumorti:
A test suite for C++ has been added, with 100% coverage, increasing the reliability of QiliSim and QTensor. In the process, several very minor bugs have been identified and fixed.


.github/workflows/tests.yml
Lumorti, line 1: Note for reviewers: this took a while to figure out, but basically we need to reactivate the environment each time rather than doing uv run, because uv run does an implicit uv sync and that messes with some of the coverage stuff

src/qilisdk_cpp/backends/qilisim/analog/iterations.h
Lumorti, line 1: Note for reviewers: throughout you're going to see GCOV_EXCL_BR_START and GCOV_EXCL_BR_STOP, these tell the coverage tool to only collect line-based coverage and not branch-based coverage. The branch-based coverage is done at a much lower level, so even simple statements like "var = value" have like 3 different branches, so is more or less nonsensical in the coverage report.

src/qilisdk_cpp/backends/qilisim/digital/gate.cpp
Lumorti, line 1: Note for reviewers: a general code change I have made is that now gate/operation names are standardized in the C++ side, if we could have type of controlled X, no matter what it was called, the gate type is now "X" and the controls and targets are stored. This reduces me having to checking for name == "Toffoli", name == "CX" etc. everywhere. Similarly the base matrix of a CX/CNOT is now just the matrix for the X.

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -343,6 +345,16 @@ void sampling_matrix_free(const std::vector<Gate>& gates, const std::vector<bool
...
+    // If it's a density matrix, check if it's pure
+    if (!is_statevector) {
+        DenseMatrix state_squared = state.adjoint().cwiseProduct(state);
+        double trace_squared = state_squared.trace().real();
+        if (std::abs(trace_squared - 1.0) < config.get_atol()) {
+            state = get_vector_from_density_matrix(state);
+            is_statevector = true;
+        }
+    }
+
```
Lumorti, line 357: Note for reviewers: copied from the non-matrix-free code

src/qilisdk_cpp/backends/qilisim/representations/matrix_free_hamiltonian.cpp
```diff
@@ -132,6 +153,13 @@ MatrixFreeHamiltonian& MatrixFreeHamiltonian::operator+=(const MatrixFreeHamilto
         A reference to the resulting Hamiltonian after addition.
     */
     std::unordered_map<std::string, int> op_map;
+    for (size_t i = 0; i < operators.size(); ++i) {
+        std::string id;
+        for (const auto& op : operators[i].second) {
+            id += op.get_id() + "|";
+        }
+        op_map[id] = int(i);
+    }
```
Lumorti, line 162: Note for reviewers: fix for a bug found by tests, basically the op_map didn't include terms in the base Hamiltonian, e.g. (X + Y) += X would give X + X + Y rather than 2 X + Y, wouldn't have affected results but just might not have been as compressed as it could have been

src/qilisdk_cpp/backends/qilisim/representations/matrix_free_hamiltonian.cpp
```diff
@@ -195,4 +223,48 @@ void MatrixFreeHamiltonian::add(const std::complex<double>& coeff, const std::ve
...
+            return false;
+        }
+        for (size_t j = 0; j < ops1.size(); ++j) {
+            if (!(ops1[j] == ops2[j])) {
+                return false;
+            }
+        }
+    }
+    return true;
+}
```
Lumorti, line 268: Note for reviewers: added equality, this sorts the two lists and then compares each one, has been tested

src/qilisdk_cpp/backends/qilisim/representations/matrix_free_operator.cpp
Lumorti, line 1: Note for reviewers: okay, this one you might just have to trust the tests, basically there were some issues with left/right application of Y operator, but this was never used in the circuit sampling nor the time evolution. For each section of this code there are four branches: applied to statevector, left multiplication with density matrix (i.e. O rho), right multiplication with a density matrix (i.e. rho O ^ dag) and left/right multiplication (i.e. O rho O ^ dag). Only the first and last of these was ever used, the other two are just in there for completeness, but they contained bugs for some cases since they were never tested until now.

src/qilisdk_cpp/backends/qilisim/utils/parsers.cpp
```diff
@@ -483,7 +482,23 @@ std::vector<Gate> parse_gates(const py::object& circuit, double atol, const py::
...
+                if (key_tuple.size() == 2) {
+                    py::handle gate_class = key_tuple[0];
+                    py::handle param_name = key_tuple[1];
+                    std::string gate_class_name = py::str(gate_class.attr("__name__"));
+                    std::string param_name_str = py::str(param_name);
+                    py::tuple new_key = py::make_tuple(gate_class_name, param_name_str);
+                    new_gate_noise_map[new_key] = item.second;
+                }
+            }
+            gate_noise_map = new_gate_noise_map;
```
Lumorti, line 501: Note for reviewers: this is mostly useful for unit testing, where rather than having to make a class that inherits from RX in order to mock it, I can just make a new class called RX and that will be treated as the same as the true RX class. Also in general it means dealing with less pythonic stuff within the C++.

src/qilisdk_cpp/backends/qilisim/qilisim.cpp
```diff
@@ -205,11 +203,6 @@ py::object QiliSimCpp::execute_time_evolution(const py::object& functional, cons
         if (hamiltonians.size() != parameters_list.size()) {
             throw py::value_error("Number of Hamiltonians does not match number of parameter lists");
         }
-        for (size_t h_ind = 0; h_ind < hamiltonians.size(); ++h_ind) {
-            if (parameters_list[h_ind].size() != step_list.size()) {
-                throw py::value_error("Number of parameters for Hamiltonian " + std::to_string(h_ind) + " does not match number of time steps");
-            }
-        }
```
Lumorti, line 212: Note for reviewers: removed because it was unreachable, this is implied by the parameters_list being initialized with the size of the step list earlier

src/qilisdk_cpp/libs/pybind.cpp
Lumorti, line 1: Note for reviewers: there were a bunch of problems caused by the previous method of initializing these python class types as constant references in the header, now we don't init the python stuff until explicitly told to. We can also clear the references, mostly useful in the test suite where we sometimes create/destroy new test environments.

tests/unit_cpp/core/test_qtensor.cpp
Lumorti, line 1: Note for reviewers: Finally we get to the tests. I would say either use AI to look through these or just pick a few randomly to assure yourself that the tests make sense, because otherwise it's a lot. Some of these files were done entirely by me, meanwhile I think the QTensor one is mostly Claude just because it's a lot of very simple tests.

Relevant things to know about the google test syntax:

- TEST vs TEST_F, very similar, except TEST_F uses the fixture (i.e. class) given by the first argument in the test init, e.g. TEST_F(myclass, mytestname) would create an instance of myclass as a testing environment
- EXPECT_THROW, EXPECT_ANY_THROW, EXPECT_EQ etc. do as you would imagine, they are the same as Python assert. The only difference is that if you see `EXPECT_EQ(something, something else) << "some text"`, the "some text" is the message output if the test fails.

scripts/checks.sh
Lumorti, line 1: Note for reviewers: Added a bunch of bash scripts in the scripts/ folder, because at this point some tasks like coverage are getting a bit complicated because of the C++/Python merging. So now for stuff like coverage you can just do bash scripts/cov.sh to generate a html coverage report, or do bash scripts/checks.sh to do all of the ruff/ty/clang-format/tests in one script

CMakeLists.txt
Lumorti, line 1: Note for reviewers: quite a lot of very specific changes, but main changes are that we now build each component as a shared library rather than a specifically python library, such that they can then be accessed by the C++ testing suite directly in C++. If the flag is given, everything will be build with coverage on, which generates gcno (generated on compilation, contains info about the source) and gcna (generated on running the code, contains info about which lines actually triggered). If the testing flag is given we build the test suite, which takes a looooot longer (like 50 files rather than 24)

scripts/checks.sh
fedonman: Nice idea!

fedonman (approved):
Seems great! Thanks a lot Luke! 💯

PR #174, opened 2026-03-17
[MISC] Update to CUDAQ 0.14

fedonman:
Updated CUDAQ to v0.14.


src/qilisdk/backends/cuda_backend.py
```diff
@@ -553,6 +553,18 @@ def _qtensor_observable_to_hamiltonian(observable: QTensor, nqubits: int) -> Ham
         except ValueError as exc:
             raise ValueError("QTensor observables in the CUDA backend must be Hermitian operators.") from exc
 
+    @staticmethod
+    def _qtensor_initial_state_to_cuda(initial_state: QTensor) -> State:
+        normalized_state = initial_state.unit()
+        if normalized_state.is_bra():
+            normalized_state = normalized_state.adjoint()
+
+        cuda_state_data = np.array(normalized_state.dense(), dtype=np.complex128)
```
amirazzam2000: we should probably use the QiliSDK settings complex datatype here? (double check but in the previous versions of cuda sometimes when you use complex64 it didn't work, but maybe that is fixed now?)

amirazzam2000 (changes requested):
thanks Vyron!! 🙌

src/qilisdk/backends/cuda_backend.py
Lumorti: If it was very specific about the data type before, then I don't trust that they'll be more general now

Because it was only on some specific cases where it caused problems, no? Like we didn't notice it for a while

changes/174.misc.md
Lumorti, line 1: I assume this didn't add any new support or anything new for the user, but in case it did, we should mention it here

Lumorti (approved):
Nice! Thanks for doing the upgrade 🚀

src/qilisdk/backends/cuda_backend.py
fedonman: For simulation, we set an environment variable for precision.

amirazzam2000: Exactly, this is why maybe we should modify this line to use the environment variable. But yeah we should be mindful because as luke mentioned in some cases it was causing problems. This is why i was asking if we can test to check if they fixed this issue (or at least they started to return a more helpful error message)

fedonman: I don't have a test case for this. Can you check it from you side?

What I meant, is that cudaq sets the precision when setting the backend, and for now it works only for the `nvidia` backend:
```python
                float_precision = "fp32" if get_settings().complex_precision == Precision.COMPLEX_64 else "fp64"
                cudaq.set_target("nvidia", option=float_precision)
                logger.debug("GPU detected, using cudaq's 'nvidia' backend")
```

fedonman: Current tests are passing, so I updated it.

amirazzam2000 (approved):
Thanks!! 🙌

PR #175, opened 2026-03-17
[SDK-237] Restructuring The Functionals

amirazzam2000:
Restructured the functionals and results system with a decoupled, type-safe readout architecture.

**Breaking Changes**

- **`Sampling` renamed to `DigitalPropagation`**: The functional now takes only a `circuit` argument. Shot count is specified via readout.
- **`TimeEvolution` renamed to `AnalogEvolution`**: The functional now takes `schedule`, `initial_state`, and `store_intermediate_results`. Observables and shot count are specified via readout.
- **`SamplingResult` and `TimeEvolutionResult` removed**: All primitive functionals now return a unified `FunctionalResult`.
- **`backend.execute()` now requires a `readout` parameter**: A `Readout` specification is passed at execution time instead of being bundled into the functional.

**New Readout System**

A new `qilisdk.readout` module provides the `Readout` builder and three readout types.
Build a specification by chaining `with_*` methods and pass it to `backend.execute()`:

```python
from qilisdk.analog import Z

result = backend.execute(functional, readout=Readout().with_sampling(nshots=1000))
result = backend.execute(functional, readout=Readout().with_expectation(observables=[Z(0)]))
result = backend.execute(functional, readout=Readout().with_state_tomography())

# Multiple readout types can be combined in a single execution
result = backend.execute(
    functional,
    readout=Readout().with_sampling(nshots=500).with_expectation(observables=[Z(0)]),
)
```

The three readout types are:

- `Readout().with_sampling(nshots)`: sample in the computational basis and collect bitstring counts.
- `Readout().with_expectation(observables, nshots=0)`: compute `⟨ψ|O|ψ⟩` for each observable (`nshots=0` uses the exact state-vector inner product).
- `Readout().with_state_tomography()`: return the full quantum state vector.

**Type-Safe Results**

`Readout` carries generic type parameters `S`, `E`, `T` that track which readout slots are populated.
`backend.execute()` propagates these to the returned `FunctionalResult`, so the type checker knows
which result fields are present without runtime guards:

```python
from qilisdk.analog import Z

# result.sampling is SamplingReadoutResult — type checker knows this, no Optional guard needed
result = backend.execute(functional, readout=Readout().with_sampling(nshots=1000))
top2 = result.sampling.get_probabilities(n=2)

# result.expectation is ExpectationReadoutResult — type checker knows this
result = backend.execute(functional, readout=Readout().with_expectation(observables=[Z(0)]))
evs = result.expectation.expectation_values
```

**Migration Guide**

**Before (digital circuit sampling):**

```python
from qilisdk.functionals import Sampling

sampling = Sampling(circuit=circuit, nshots=500)
result = backend.execute(sampling)
print(result.samples)
```

**After:**

```python

functional = DigitalPropagation(circuit)
result = backend.execute(functional, readout=Readout().with_sampling(nshots=500))
print(result.samples)
```

**Before (analog time evolution):**

```python
from qilisdk.functionals import TimeEvolution

te = TimeEvolution(schedule=schedule, initial_state=psi0, observables=[Z(0)], nshots=100)
result = backend.execute(te)
print(result.expected_values)
```

**After:**

```python
from qilisdk.analog import Z

evolution = AnalogEvolution(schedule=schedule, initial_state=psi0)
result = backend.execute(evolution, readout=Readout().with_expectation(observables=[Z(0)]).with_state_tomography())
print(result.expectation_values)
```

**Before (variational program):**

```python
vp = VariationalProgram(functional=Sampling(ansatz), optimizer=opt, cost_function=cost_fn)
result = backend.execute(vp)
```

**After:**

```python

vp = VariationalProgram(functional=DigitalPropagation(ansatz), optimizer=opt, cost_function=cost_fn)
result = backend.execute(vp, readout=Readout().with_sampling(nshots=1000))
```

**Result Properties**

The unified `FunctionalResult` provides convenience shortcuts:

- `get_samples`: shot counts (`with_sampling`)
- `get_probabilities`: measurement probabilities (`with_sampling` or `with_state_tomography`)
- `get_state`: full quantum state (`with_state_tomography`)
- `get_expectation_values`: list of expectation values (`with_expectation`)
- `intermediate_samples`, `intermediate_probabilities`, `intermediate_states`, `intermediate_expectation_values`: per-step lists when `store_intermediate_results=True`

Typed forwarding properties (`result.sampling`, `result.expectation`, `result.state_tomography`) return the raw result objects and are preferred for production code.

**Other Changes**

- Added a `Readout` documentation page covering the builder pattern, all three readout types, result access, intermediate results, and a complete end-to-end example.
- Added QiliSim C++ backend support for `DigitalPropagation`, `AnalogEvolution`, and `QuantumReservoir`.
- Added CUDA and QuTiP backend support for the new readout system.
- Updated SpeQtrum cloud submission (`SpeQtrum.submit`) to accept `Readout`.
- Updated all documentation to reflect the new structure.


changes/175.feature.md
```diff
@@ -0,0 +1,92 @@
...
+
+### Result Properties
+
+The unified `FunctionalResult` provides:
+
+- `final_samples` — Shot counts from `SamplingReadout`
+- `final_probabilities` — Measurement probabilities
+- `final_state` — Quantum state from `StateTomographyReadout`
+- `final_expected_values` — Expectation values from `ExpectationReadout`
+- `samples`, `probabilities`, `states`, `expected_values` — Lists of intermediate results when `store_intermediate_results=True`
```
fedonman: I know it's not ready for review, just a minor comment.

I would rename the `final_XYZ` to just `XYZ` since they are the properties the user will interact the most, and rename the intermediate ones to `intermediate_XYZ`.

E.g. `samples` and `intermediate_samples`.

It's more natural to the user and closer to what we have now.

amirazzam2000: ohhh good point, I will update them.

changes/175.feature.md
```diff
@@ -0,0 +1,92 @@
+Restructured the functionals and results system with a decoupled readout architecture.
+
+### Breaking Changes
+
+- **`Sampling` renamed to `DigitalPropagation`**: The functional now takes only a `circuit` argument. Shot count is specified via readout.
+- **`TimeEvolution` renamed to `AnalogEvolution`**: The functional now takes `schedule`, `initial_state`, and `store_intermediate_results`. Observables and shot count are specified via readout.
```
Lumorti, line 6: Should it be DigitalPropogation and AnalogPropogation? I feel like with these two being similar but not the same users (and me) are gonna accidentally write DigitalEvolution and AnalogPropogation sometimes.

Actually, maybe DigitalEvolution isn't so bad, I know you're worried it might seem like Trotterization, but I think that is the most natural name for it

Actually, maybe it's not so bad, I'm getting used to it as I'm reading the rest of the PR

changes/175.feature.md
```diff
@@ -0,0 +1,92 @@
...
+- **`SamplingResult` and `TimeEvolutionResult` removed**: All primitive functionals now return a unified `FunctionalResult`.
+- **`backend.execute()` now requires a `readout` parameter**: Readout methods are decoupled from functionals and passed at execution time.
+
+### New Readout System
+
+A new `qilisdk.readout` module provides three readout methods that are passed to `backend.execute()`:
+
+- `SamplingReadout(nshots)` — Sample the circuit a given number of times.
+- `ExpectationReadout(observables)` — Compute expectation values of observables.
+- `StateTomographyReadout()` — Return the full quantum state.
```
Lumorti: Em dashes, could it be that perhaps Claude wrote this? 👀

docs/fundamentals/digital.rst
Lumorti, line 1: Very nice renaming, TrotterizedTimeEvolution did look so much like a functional

docs/conf.py
```diff
@@ -128,7 +128,7 @@ def _safe_collect_pages(app):  # noqa: ANN001, ANN201
     env = app.builder.env
     if not hasattr(env, "_viewcode_modules"):
         return
-    if not _viewcode.is_supported_builder(app.builder):
+    if not _viewcode.is_supported_builder(app.builder, app.config.viewcode_enable_epub):
```
Lumorti, line 131: What does this do?

src/qilisdk/backends/backend.py
```diff
@@ -113,36 +231,36 @@ def _execute_quantum_reservoir(self, functional: QuantumReservoir) -> QuantumRes
...
+        Runs the classical optimizer loop: at each iteration the inner
+        functional is executed with the given ``readout`` methods and the
+        cost function is evaluated.
+
         Args:
-            functional (VariationalProgram): Variational program to optimize.
+            functional (VariationalProgram): Variational
+                program to optimize.
+            readout (list[ReadoutMethod]): Readout specifications forwarded
+                to each inner functional execution.
```
Lumorti: Does this render correctly in the API docs? I know sphinx can be quite picky about spaces and newlines and lists etc.

src/qilisdk/backends/backend_config.py
```diff
@@ -287,10 +311,19 @@ def statevector(
...
+            normalize_after_each_gate (bool): Whether to normalize the
+                statevector after each gate application to mitigate
+                numerical errors at the cost of increased runtime.
+                Defaults to ``False``.
+            matrix_free (bool): Whether to use the matrix-free
+                implementation for statevector simulation. Defaults to
+                ``True``.
+            combine_single_qubit_gates (bool): Whether to combine
+                consecutive single-qubit gates into a single operation.
+                Defaults to ``True``.
```
Lumorti, line 321: It seems something (Claude or Ruff or who knows what) really likes short lines, maybe this is a bit too short though

src/qilisdk/backends/qilisim.py
```diff
@@ -25,31 +25,40 @@
...
-
-    backend = QiliSim(
-        analog_simulation_method=AnalogMethod.arnoldi(
-            dim=16,
-            num_substeps=2
-        ),
-        digital_simulation_method=DigitalMethod.statevector(max_cache_size=2_000),
-        execution_config=ExecutionConfig(num_threads=4, seed=42, monte_carlo=MonteCarloConfig(trajectories=200),),
-    )
+    """Backend that runs digital-circuit and analog time-evolution experiments using a custom C++ simulator.
```
Lumorti, line 35: But, weirdly enough, this was moved from being two lines to one

src/qilisdk/core/qtensor.py
```diff
@@ -589,6 +590,15 @@ def basis_state(cls, n: int, N: int) -> QTensor:
         mat = csr_matrix(([1.0], ([n], [0])), shape=(N, 1))
         return QTensor(mat)
 
+    def scale_qtensor(self, nqubits: int) -> QTensor:
```
Lumorti: Maybe a docstring for this

Lumorti: Also maybe it should just be called like `expand(nqubits)`? Since when I first saw scale I thought it was just gonna be an overload for multiplication/division by a scalar

src/qilisdk/functionals/__init__.py
```diff
@@ -12,24 +12,28 @@
...
 # limitations under the License.
 
+"""Public API for quantum functionals and their results.
+
+This package exposes the core functional abstractions used to describe
+executable quantum routines (digital propagation, analog evolution,
+quantum reservoirs, variational programs) together with their result
+containers.
+"""
+
```
Lumorti: Why add this here? Does it appear in the API docs somewhere? If so, should we put them in all of the __init__ files?

src/qilisdk/functionals/functional_result.py
```diff
@@ -11,8 +11,237 @@
...
+            intermediate_results (list[list[ReadoutResult]] | None): Optional
+                per-step intermediate readout results. Defaults to None.
+
+        Raises:
+            ValueError: If ``readout_results`` contains duplicate readout types.
+        """
+
+        if len({ro.__class__ for ro in readout_results}) != len(readout_results):
+            raise ValueError(
+                f"Each type of readout is allowed to be specified once.\nprovided a list with the following types {readout_results}"
```
Lumorti: Wait, I thought you wanted to be able to do readout=[SamplingReadout(nshots=10), SamplingReadout(nshots=100000)] or something like that?

src/qilisdk/functionals/functional_result.py
```diff
@@ -11,8 +11,237 @@
...
+        """
+        if self._intermediate_results:
+            if self.has_probabilities():
+                results = []
+                for res in self:
+                    results.append(res.probabilities)
+                return results
+            raise ValueError(
+                "Can't find probabilities in results, because no Sampling/State Tomography readout was provided."
+            )
```
Lumorti:
```suggestion
            raise ValueError("Can't find probabilities in results, because no Sampling/State Tomography readout was provided.")
```

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+                exact (state-vector-based) evaluation.
+
+        Returns:
+            ExpectationReadout: A configured expectation-value readout instance.
+
+        Raises:
+            AttributeError: If called on a subclass instead of
+                :class:`ReadoutMethod`.
+        """
+        cls._require_factory_called_on_base("expectation_values")
```
Lumorti: What happens if it doesn't have it? Does this throw a nice error? Or should we add some sort of check here?

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+            expected = ", ".join(positional_names)
+            raise TypeError(
+                f"{class_name} expects at most {len(positional_names)} positional argument(s): ({expected})."
+            )
+        merged = dict(keyword_args)
+        for name, value in zip(positional_names, positional_args):
+            if name in merged:
+                raise TypeError(f"{class_name} got multiple values for '{name}' (positional and keyword).")
+            merged[name] = value
+        return merged
```
Lumorti: Why is this needed? I'm sure there's a good reason, but I don't immediately see why we can't just use normal Python positional args

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+
+    _scaled_nqubits: int | None = PrivateAttr(default=None)
+
+    @model_validator(mode="after")
+    def set_qtensor_observables(self) -> ExpectationReadout:
+        self.qtensor_observables = [(o if isinstance(o, QTensor) else o.to_qtensor()) for o in self.observables]
+        self._scaled_nqubits = None
+        return self
+
+    def scale_observables(self, nqubits: int) -> None:
```
Lumorti: I made a comment before on a similar function, but I do think expand_observables is clearer than scale_observables, scale sounds like you're doing observables *= float or something

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+    :meth:`ReadoutMethod.state_tomography`.
+
+    Args:
+        state_tomography_method (Literal["exact"]): Tomography method
+            identifier.  Currently only ``"exact"`` is supported.  Accepted
+            as a keyword argument or as the first positional argument.
+        compute_probabilities (bool): If ``True`` (the default), the
+            resulting :class:`~qilisdk.readout.readout_result.StateTomographyReadoutResult`
+            will also derive computational-basis probabilities from the
+            reconstructed state.
```
Lumorti: Should this be here? It isn't mentioned in the rest of the class. Also, remember that now we have QTensor.probabilities()  if they want to calculate this

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+            ValueError: If no suitable result is present.
+        """
+        for ro in self._readout_results:
+            if isinstance(ro, SamplingReadoutResult):
+                return ro.probabilities
+            if isinstance(ro, StateTomographyReadoutResult) and ro.readout.compute_probabilities:
+                return ro.probabilities  # ty:ignore[invalid-return-type]
+        raise ValueError(
+            "Can't find probabilities in results, because no Sampling/State Tomography readout was provided."
+        )
```
Lumorti: I do think that having probabilities here could give some unintended results. Like if I do [Sampling, StateTomography] I would get a different result that [StateTomography, Sampling] which I think the user wouldn't expect.

Also, I think getting probabilities from StateTomography is fair, because you hope that it will be close to the actual state, but with sampling I think it might be a stretch. Like say I do SamplingReadout(nshots=10) and I get {"0": 3, "1": 7}. Can I really say that my probabilities are [30%, 70%]? Because doing .probabilities seems to imply that you get the underlying probability distribution, which you should get something close with state tomog, but sampling maybe not

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+            if not all(len(bitstring) == nqubits for bitstring in bitstrings):
+                raise ValueError("Not all bitstring keys have the same length.")
+
+            # Calculate probabilities
+            self._probabilities = {
+                bitstring: (counts / self._readout.nshots if self._readout.nshots and self._readout.nshots > 0 else 0.0)
+                for bitstring, counts in self._samples.items()
+            }
+        else:
+            self._probabilities = probabilities
```
Lumorti: If we make a SamplingResult from probabilities, should we then simulate samples? I personally think we shouldn't have probabilities anywhere, but if we are then maybe we should also simulate some samples from the probabilities, since the user has requested a sampling result, not a probability result

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+            n = len(self._probabilities)
+        return heapq.nlargest(n, self._probabilities.items(), key=operator.itemgetter(1))
+
+    def __repr__(self) -> str:
+        return "State Tomography Results: (\n" + (f"\tfinal_state={pformat(self.state)}\n") + ")\n\n"
+
+    __str__ = __repr__
+
+
+def _probabilities_from_state(state: QTensor, *, check_normalization: bool = False) -> dict[str, float]:
```
Lumorti: QTensor.probabilities() already exists!

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+                f"Probabilities not normalized: sum={s}. "
+                "If expected (e.g. unnormalized state), set check_normalization=False."
+            )
+    n_qubits = state.nqubits
+    return {format(idx, f"0{n_qubits}b"): float(p) for idx, p in enumerate(probs)}
+
+
+def _samples_from_state(state: QTensor, nshots: int = 100, seed: int | None = None) -> dict[str, int]:
+    probabilities = _probabilities_from_state(state)
+    return _samples_from_probabilities(probabilities=probabilities, nshots=nshots, seed=seed)
```
Lumorti, line 461: We could also add a QTensor.sample(nshots) method to QTensor, if you wanted? Since we already functions like that in the C++, could even take a noise model as an optional parameter

src/qilisdk_cpp/backends/qilisim/utils/sample.cpp
```diff
@@ -0,0 +1,149 @@
...
+        counts = apply_readout_error(counts, noise_model_cpp, n_qubits);
+    }
+
+    // Only keep measured qubits in the counts
+    return  filter_counts(counts, qubits_to_measure);
+    // py::dict samples;
+    // for (const auto& pair : counts) {
+    //     samples[py::cast(pair.first)] = py::cast(pair.second);
+    // }
+    // return samples;
```
Lumorti:
```suggestion
 
```

src/qilisdk_cpp/backends/qilisim/qilisim.cpp
```diff
@@ -22,16 +28,69 @@
...
+
+        } else if (py::isinstance(ro, SamplingReadout)) {
+            int n_shots = ro.attr("nshots").cast<int>();
+            std::map<std::string, int> counts = construct_samples(state_dense, n_qubits, n_shots, noise_model_cpp, config, qubits_to_measure);
+            py::dict samples;
+            for (const auto& pair : counts) {
+                samples[py::cast(pair.first)] = py::cast(pair.second);
+            }
+            results.append(SamplingReadoutResult(
+                "readout"_a = py::module_::import("copy").attr("copy")(ro),
```
Lumorti: Do we need the copy?

src/qilisdk_cpp/backends/qilisim/qilisim.cpp
```diff
@@ -91,26 +146,30 @@ py::object QiliSimCpp::execute_sampling(const py::object& functional, const py::
...
-    py::dict samples;
-    for (const auto& pair : counts) {
-        samples[py::cast(pair.first)] = py::cast(pair.second);
-    }
+    // py::dict samples;
+    // for (const auto& pair : counts) {
+    //     samples[py::cast(pair.first)] = py::cast(pair.second);
+    // }
+
+
```
Lumorti:
```suggestion
```

Lumorti (changes requested):
Very nice changes overall! Just a few comments, notably that I'm not sure about having SamplingReadoutResult.probabilities, also QTensor.probabilities() already exists, and maybe we should make QTensor.expand() and QTensor.sample() since they are general useful

src/qilisdk/readout/readout_result.py
fedonman: I believe it is the other way around: we should keep calculating the probabilities from the samples (the current behavior, and what all frameworks do) and remove that calculation from `StateTomography`. If the user requests the state, they can easily get the probabilities by calling `QTensor.probabilities`.

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+    from .readout import ExpectationReadout, ReadoutMethod, SamplingReadout, StateTomographyReadout
+
+
+NORMALIZATION_TOLERANCE = 1e-8
+
+
+def _real_if_close(number: Number) -> Number:
+    if isinstance(number, complex) and number.imag < get_settings().atol:
+        return number.real
+    return number
```
fedonman, line 48: Should take the absolute right?


```suggestion
def _real_if_close(number: Number) -> Number:
    if isinstance(number, complex) and abs(number.imag) < get_settings().atol:
        return number.real
    return number
```

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+    def _init_from_samples(
+        self, readout: SamplingReadout, samples: dict[str, int] | None, probabilities: dict[str, float] | None = None
+    ) -> None:
+        self._readout: SamplingReadout = readout
+        if samples:
+            self._samples: dict[str, int] = samples
+        elif probabilities:
+            self._samples = _samples_from_probabilities(probabilities)
+        else:
+            raise ValueError("can't initialize Sampling Results if both samples and probabilities are not provided.")
```
fedonman: `if samples:` evaluates to `False` if an empty dictionary is provided `samples = {}`, do we want this? Same question applies to other places where we do `if dictionary:` instead of `if dictionary is not None:`

src/qilisdk/readout/readout.py
fedonman: I agree. The point of the refactor was for each readout to provide its relevant result. `StateTomography` should return the state. Probabilities can be calculated easily by calling `QTensor.probabilities()`.

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+        readout.scale_observables(nqubits=state.nqubits)
+        self._expected_values = [_real_if_close((expect_val(o, state))) for o in readout.qtensor_observables]
+
+    def _init_from_expected_values(
+        self,
+        readout: ExpectationReadout,
+        expected_values: list[Number],
+    ) -> None:
+        self._readout: ExpectationReadout = readout
+        self._expected_values = expected_values
```
fedonman: Add annotations.
```suggestion
        self._expected_values: list[Number] = expected_values
```

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+            self._init_from_state(readout, state)
+        else:
+            raise ValueError(
+                "Can't initialize Expectation Readout if the expected values and the state are not provided."
+            )
+
+    def _init_from_state(self, readout: ExpectationReadout, state: QTensor) -> None:
+        self._readout = readout
+        readout.scale_observables(nqubits=state.nqubits)
+        self._expected_values = [_real_if_close((expect_val(o, state))) for o in readout.qtensor_observables]
```
fedonman: Add annotations. Also, why do we mutate the input?


```suggestion
        self._readout: ExpectationReadout = readout
        readout.scale_observables(nqubits=state.nqubits)
        self._expected_values: list[Number] = [_real_if_close((expect_val(o, state))) for o in readout.qtensor_observables]
```

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+            dict[str, float]: Mapping of bitstring to probability.
+
+        Raises:
+            ValueError: If no suitable result is present.
+        """
+        for ro in self._readout_results:
+            if isinstance(ro, SamplingReadoutResult):
+                return ro.probabilities
+            if isinstance(ro, StateTomographyReadoutResult) and ro.readout.compute_probabilities:
+                return ro.probabilities  # ty:ignore[invalid-return-type]
```
fedonman: Why do we ignore the return type mismatch?

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+    probabilities = _probabilities_from_state(state)
+    return _samples_from_probabilities(probabilities=probabilities, nshots=nshots, seed=seed)
+
+
+def _samples_from_probabilities(
+    probabilities: dict[str, float], nshots: int = 100, seed: int | None = None
+) -> dict[str, int]:
+    states = np.array(list(probabilities.keys()))
+    probs = np.array(list(probabilities.values()), dtype=np.float64)
+    probs /= probs.sum()
```
fedonman: This silently renormalizes the input probabilities. If someone passes probabilities that don't sum to 1 due to a real bug upstream, this hides it. Consider at least logging a warning when the sum deviates beyond a threshold, similar to how `_probabilities_from_state` has an explicit `check_normalization` flag.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+        Raises:
+            AttributeError: if it's called on a child rather than the base class.
+        """
+        if cls is not ReadoutMethod:
+            raise AttributeError(
+                f"{factory_name}() is only available on ReadoutMethod; use ReadoutMethod.{factory_name}(...)"
+            )
+
+    @classmethod
+    def sample(cls, nshots: int = 100) -> ReadoutMethod:
```
fedonman:
```suggestion
    def sample(cls, nshots: int = 100) -> SamplingReadout:
```

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+            SamplingReadout: A configured sampling readout instance.
+
+        Raises:
+            AttributeError: If called on a subclass instead of
+                :class:`ReadoutMethod`.
+        """
+        cls._require_factory_called_on_base("sample")
+        return SamplingReadout(nshots=nshots)
+
+    def is_sample(self) -> bool:
```
fedonman: Rename to the class name that is being checked: SamplingReadout. Same with the other two, `is_expectation_readout` and `is_state_tomography_readout`.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+            ExpectationReadout: A configured expectation-value readout instance.
+
+        Raises:
+            AttributeError: If called on a subclass instead of
+                :class:`ReadoutMethod`.
+        """
+        cls._require_factory_called_on_base("expectation_values")
+        return ExpectationReadout(observables=observables, nshots=nshots)
+
+    def is_expectation_values(self) -> bool:
```
fedonman: Rename to the class name that is being checked: ExpectationResult. Also, consider returning a TypeGuard so it's more useful to the user.

```suggestion
    def is_expectation_readout(self) -> TypeGuard[ExpectationReadout]:
```

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+    def is_state_tomography(self) -> bool:
+        """Check whether this readout is a :class:`StateTomographyReadout`.
+
+        Returns:
+            bool: ``True`` if this instance is a
+            :class:`StateTomographyReadout`, ``False`` otherwise.
+        """
+        return isinstance(self, StateTomographyReadout)
+
+    def is_valid(self) -> bool:
```
fedonman: This isn't used anywhere. Is it useful for the user to be included in public API? Same for the three `is_XYZ` methods that are called.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+            manually.
+
+    Examples:
+        >>> ExpectationReadout(observables=[hamiltonian], nshots=0)
+        >>> ExpectationReadout([hamiltonian], 0)
+    """
+
+    nshots: int = Field(default=0, ge=0)
+    observables: list[Hamiltonian | QTensor]
+    qtensor_observables: list[QTensor] = Field(default=[], init=False)
```
fedonman: Use `default_factory` for mutable defaults.
```suggestion
    qtensor_observables: list[QTensor] = Field(default_factory=list, init=False)
```

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+
+        Args:
+            nqubits (int): Target qubit count to scale the observables to.
+        """
+        if self._scaled_nqubits == nqubits:
+            return
+        self.qtensor_observables = [
+            (o.scale_qtensor(nqubits) if isinstance(o, QTensor) else o.to_qtensor(nqubits)) for o in self.observables
+        ]
+        self._scaled_nqubits = nqubits
```
fedonman, line 165: I don't think we should mutate the `qtensor_observables` here. Why not return a new instance?

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+
+    @classmethod
+    def _require_factory_called_on_base(cls, factory_name: str) -> None:
+        """Ensure factory helpers are only called on ``ReadoutMethod``.
+
+        Raises:
+            AttributeError: if it's called on a child rather than the base class.
+        """
+        if cls is not ReadoutMethod:
+            raise AttributeError(
```
fedonman: `AttributeError` is semantically wrong here — the attribute exists, it just shouldn't be called on a subclass. `TypeError` is the standard choice for "you called this wrong" (same as Python raises for wrong argument counts).

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+        Args:
+            nshots (int | None): Number of shots (>= 0), by keyword or first positional arg.
+        """
+        payload = self._merge_positional_args(
+            class_name=type(self).__name__,
+            positional_args=args,
+            positional_names=("nshots",),
+            keyword_args=data,
+        )
+        super().__init__(**payload)
```
fedonman: This completely destroys IDE autocompletion and type checker visibility. A user writing `SamplingReadout(nshots=...)` gets no help from their editor.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+            observables (list[Hamiltonian | QTensor] | None): Observables to measure.
+            nshots (int | None): Number of shots (>= 0), by keyword or second positional arg.
+        """
+        payload = self._merge_positional_args(
+            class_name=type(self).__name__,
+            positional_args=args,
+            positional_names=("observables", "nshots"),
+            keyword_args=data,
+        )
+        super().__init__(**payload)
```
fedonman: Same here.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+        Args:
+            state_tomography_method (Literal["exact"] | None): Method by keyword or first positional arg.
+        """
+        payload = self._merge_positional_args(
+            class_name=type(self).__name__,
+            positional_args=args,
+            positional_names=("state_tomography_method",),
+            keyword_args=data,
+        )
+        super().__init__(**payload)
```
fedonman: Same here.

src/qilisdk/readout/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+    :class:`ReadoutResult` -- abstract single-result base.
+    :class:`SamplingReadoutResult` -- bitstring counts and probabilities.
+    :class:`ExpectationReadoutResult` -- expectation values.
+    :class:`StateTomographyReadoutResult` -- quantum state and probabilities.
+"""
+
+from .readout import ExpectationReadout, ReadoutMethod, SamplingReadout, StateTomographyReadout
+from .readout_result import ExpectationReadoutResult, ReadoutResult, SamplingReadoutResult, StateTomographyReadoutResult
+
+__all__ = [
```
fedonman, line 59: `ReadoutCompositeResults` should be included here?

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+
+    The container is iterable and supports ``len()``.
+
+    Args:
+        readout_results (list[ReadoutResult]): Individual readout results
+            from the execution.
+    """
+
+    def __init__(self, readout_results: list[ReadoutResult]) -> None:
+        self._readout_results = readout_results
```
fedonman: Right now `ReadoutCompositeResults ` is a `list[ReadoutResult]` with linear isinstance-scanning for every accessor. This has two consequences:

Every property can fail at runtime (`ValueError` if the right type isn't in the list), and there's no way for a caller to know at compile time whether `,samples` will work.

The `has_*` + accessor pattern is a manual tagged union — you're reimplementing what the type system should give you.

A struct with named optional fields would be statically checkable. E.g.

```
@dataclass(frozen=True)
class ReadoutCompositeResults(Result):
    sampling: SamplingReadoutResult | None = None
    expectation: ExpectationReadoutResult | None = None
    state_tomography: StateTomographyReadoutResult | None = None
```

Now callers do results.sampling.samples and the type checker sees immediately that sampling might be None — no ValueError surprise. The `has_*` methods become unnecessary (just check is not None). And if we add a fourth readout type, the type checker forces to handle it everywhere.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,320 @@
...
+
+from typing import Literal
+
+from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_validator
+
+from qilisdk.analog import Hamiltonian  # noqa: TC001
+from qilisdk.core import QTensor
+
+
+class ReadoutMethod(BaseModel):
```
fedonman: My main issue with the new Readout architecture is that we will lose lots of static typing that we currently have.

There's a 1:1 relationship between each `ReadoutMethod` subclass and its `ReadoutResult` subclass, but the type system knows nothing about it. Nothing prevents this:

```python
# Totally wrong, but no type error
ExpectationReadoutResult(readout=some_sampling_readout, expected_values=[...])
```

This could be encoded with generics so the compiler enforces the relationship:

```python
R = TypeVar("R", bound="ReadoutResult")

class ReadoutMethod(BaseModel, Generic[R]):
    ...

class SamplingReadout(ReadoutMethod["SamplingReadoutResult"]):
    ...
```

And the inverse:

```python
C = TypeVar("C", bound="ReadoutMethod[Any]")

class ReadoutResult(Result, Generic[C]):
    @property
    @abstractmethod
    def readout(self) -> C: ...

class SamplingReadoutResult(ReadoutResult[SamplingReadout]):
    ...
```

This way any readout/result mismatch becomes a static type error rather than a silent runtime issue.

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+    ) -> None: ...
+
+    def __init__(
+        self,
+        readout: SamplingReadout,
+        *,
+        samples: dict[str, int] | None = None,
+        probabilities: dict[str, float] | None = None,
+        state: QTensor | None = None,
+    ) -> None:
```
fedonman: Despite the two overloads, the actual `__init__` accepts all four parameters as `None` simultaneously and relies on runtime `if/elif/else` to reject bad combos. The overloads suggest mutual exclusivity but don't enforce it: a caller can write `SamplingReadoutResult(r, samples={...}, state=tensor)` and the type checker won't complain.

Consider separate constructors instead:

```
class SamplingReadoutResult(ReadoutResult):
    @classmethod
    def from_samples(cls, readout, samples, probabilities=None) -> Self: ...

    @classmethod
    def from_state(cls, readout, state) -> Self: ...

```

Each factory has an unambiguous signature. No optional-everything `__init__`, no runtime validation of which combination was passed. The type checker now fully understands what's going on.

This same pattern applies to `ExpectationReadoutResult`.

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,622 @@
...
+
+    @overload
+    def __init__(self, readout: ExpectationReadout, *, state: QTensor) -> None: ...
+
+    @overload
+    def __init__(
+        self,
+        readout: ExpectationReadout,
+        *,
+        expected_values: list[Number],
```
fedonman: Can expectation values be complex?

src/qilisdk/functionals/quantum_reservoirs_result.py
```diff
@@ -12,19 +12,11 @@
...
-def _complex_dtype() -> np.dtype:
-    """Return the configured complex dtype for reservoir result arrays."""
-    return get_settings().complex_precision.dtype
+from .functional_result import FunctionalResult
 
 
 @yaml.register_class
-class QuantumReservoirResult(TimeEvolutionResult):
-    """Container for expectation values and states produced by a reservoir simulation."""
+class QuantumReservoirResult(FunctionalResult):
```
fedonman: Why do we keep this class but delete `SamplingResult` and `TimeEvolutionResult`?

src/qilisdk/functionals/functional_result.py
```diff
@@ -11,8 +11,237 @@
...
+from collections.abc import Iterator
+
+from qilisdk.core import QTensor
 from qilisdk.core.result import Result
+from qilisdk.core.types import Number
+from qilisdk.readout import ReadoutResult
+from qilisdk.readout.readout_result import ReadoutCompositeResults
 
 
 class FunctionalResult(Result):
```
fedonman: Does `FunctionalResult` still pull its weight here? Every property (`.samples`, `.probabilities`, `.state`, `.expected_values`) and every `has_*` method is a 1:1 delegation to `ReadoutCompositeResults`. The only unique value-add is the intermediate time-step sequencing (`intermediate_results`, `intermediate_*` properties), which is only used by 3 backend files.

From my understanding, the purpose of the new readout result classes was to *replace* `FunctionalResult` as the result container — not to sit behind it as a delegate. In their current form, these two classes expose nearly identical APIs, which means the same type-safety gaps are duplicated across two surfaces.

src/qilisdk/functionals/variational_program_result.py
```diff
@@ -11,50 +11,46 @@
...
 from qilisdk.functionals.functional_result import FunctionalResult
 from qilisdk.optimizers.optimizer_result import OptimizerIntermediateResult, OptimizerResult
 from qilisdk.yaml import yaml
 
-TResult_co = TypeVar("TResult_co", bound=FunctionalResult, covariant=True)
-
 
 @yaml.register_class
-class VariationalProgramResult(FunctionalResult, Generic[TResult_co]):
+class VariationalProgramResult(Result):
```
fedonman, line 23: Why `VariationalProgramResult` stopped being a `FunctionalResult`? `VariationalProgram` is still a `Functional`.

fedonman (changes requested):
Thanks a lot for the PR Ameer!

I've left several inline comments on specific issues (typing bugs, naming, dead code, etc.), but I want to flag the broader architectural concern here.

I like the core idea of separating readout configuration (`ReadoutMethod` → `SamplingReadout`, `ExpectationReadout`, `StateTomographyReadout`) from result containers (`ReadoutResult` → `SamplingReadoutResult`, etc.). That's a clean decomposition.

However, the current implementation removes the strong type checking that existed before this PR. The previous codebase used `TypeVar` + `Generic` to create compile-time guarantees about which functional produces which result:

```python
# BEFORE: functional.py (main branch)
TResult_co = TypeVar("TResult_co", bound=FunctionalResult, covariant=True)

class PrimitiveFunctional(Parameterizable, Functional, ABC, Generic[TResult_co]):
    ...

class Sampling(PrimitiveFunctional[SamplingResult]):
    result_type: ClassVar[type[SamplingResult]] = SamplingResult

# BEFORE: backend.py (main branch) — the type checker KNOWS the return type
@overload
def execute(self, functional: Sampling) -> SamplingResult: ...

@overload
def execute(self, functional: TimeEvolution) -> TimeEvolutionResult: ...

@overload
def execute(self, functional: VariationalProgram[Sampling]) -> VariationalProgramResult[SamplingResult]: ...

# Generic propagation — TResult flows through VariationalProgram to its result
def _execute_variational_program(
    self, functional: VariationalProgram[PrimitiveFunctional[TResult]]
) -> VariationalProgramResult[TResult]: ...
```

In the new code, none of this type-level machinery exists. ReadoutMethod is not generic, ReadoutResult doesn't bind back to its config, and the backend overloads have been flattened. Here are some concrete examples of what breaks:

Mismatched readout/result pairing — no type error:
```python
sampling_readout = SamplingReadout(nshots=1000)

# WRONG: passing a SamplingReadout where ExpectationReadout is expected
# The type checker accepts this because the constructor takes ReadoutMethod (the base)
result = ExpectationReadoutResult(readout=sampling_readout, expected_values=[0.5])
```

Backend return type is opaque — callers must downcast:
```python
# BEFORE: type checker inferred SamplingResult
result = backend.execute(Sampling(nshots=1000))
result.samples  # ✅ type checker knows this exists

# AFTER: type checker only sees FunctionalResult
result = backend.execute(DigitalPropagation(circuit), readout=ReadoutMethod.sample(1000))
result.samples  # ⚠️ works at runtime, but only because FunctionalResult delegates to
                # ReadoutCompositeResults which does an isinstance scan — could raise ValueError
```

ReadoutCompositeResults accessors can fail at runtime with no type-level warning:
```python
results = ReadoutCompositeResults([
    StateTomographyReadoutResult(readout=tomo_readout, state=state)
])

# Type signature says this returns dict[str, int] — but it raises ValueError at runtime
# because there's no SamplingReadoutResult in the list
samples = results.samples  # 💥 ValueError

# The has_* guard doesn't help the type checker either — it's not a TypeGuard
if results.has_samples():
    samples = results.samples  # type checker still can't narrow this
```

More such examples exist, including the `SpeQtrum` module that I intentionally left without review.

I think the readout abstraction is the right direction, but it needs to somehow carry the same generic type safety that the functional layer had. `ReadoutMethod` should be `Generic[R]` binding each config to its result type, and the backend should recover the overload-based return type guarantees that were lost.

changes/175.feature.md
amirazzam2000: hahahahah most the documentation is generated by claud yes.

changes/175.feature.md
amirazzam2000: hahahahahah I will consider this comment resolved then.

docs/conf.py
amirazzam2000: this was a weird thing. When I was building the sphinx i kept getting an error that this method is missing a mandatory field. it could be related to some sphinx version update that might have been pushed recently.

src/qilisdk/backends/backend.py
amirazzam2000: I will remove the white spaces. this doc-string was claud generated (or rather updated) so it added these white spaces.

src/qilisdk/core/qtensor.py
amirazzam2000: I like that I will do the updates!

src/qilisdk/functionals/__init__.py
amirazzam2000: claud added it and i missed it. thanks for noticing.

src/qilisdk/readout/readout.py
amirazzam2000: This is added because by default the PyDantic data object requires the user to explicitly write the name of every parameter passed to the object and will fail with a not so clear message. so i added this to allow the users to provide the parameters by positions instead of explicitly specifying the name.

src/qilisdk/readout/readout_result.py
amirazzam2000: mainly we have probabilities here to be consistent with the old interface. The previous Sampling functional returned the samples and the probabilities. Also as Vyron mentioned many other libraries return these two simultaneously. Now in regards to how we should simulate that, this is a different question that we can discuss.

src/qilisdk_cpp/backends/qilisim/qilisim.cpp
amirazzam2000: python objects are usually passed by reference, this guarantees that the readout method that corresponds to this results object is unchanged. however, i do agree that this is adding lots of overhead that might not be needed.

src/qilisdk/readout/readout_result.py
amirazzam2000: I think luke is referring here to the correctness of the results. For example let's say i run a functional and my samples return: {0: 10} that doesn't mean that the probabilities are {0: 100%} but it could be something like: {0: 99%, 1: 1%} which is something that potentially you could probe with more samples but it's incorrect if we return {0: 100%}.

one thing i could do here is to always compute the probabilities from the state (we always have access to the state in simulation) and then i can store that in SamplingReadoutResults and we can remove the probabilities from the stateTomographyReadoutResults.

src/qilisdk/readout/readout_result.py
amirazzam2000: yes in this case we want it to raise an error if you provide an empty samples list because that doesn't make sense. like you need to have sampled something to build a sampling results.

src/qilisdk/readout/readout_result.py
amirazzam2000: you are right this should be a float.

src/qilisdk/readout/readout_result.py
amirazzam2000: we are mutating the object for caching purposes, this increases how fast we can execute things like quantum reservoirs given that you don't need to expand the objects. and usually people create a new readout object per-experiment.

src/qilisdk/readout/readout.py
amirazzam2000: this _qtensor_observables attribute is basically an internal cache and it needs to be updated if we want to have any performance improvement from using it.

src/qilisdk/readout/readout_result.py
fedonman: > we always have access to the state in simulation

1. Functionals are executed in hardware too.
2. Even in simulation this statement is not true, your circuit might have lots of qubits e.g. > 50 and you cannot store the state due to memory limits, but you can run sampling, e.g. using tensor networks.
3. All frameworks calculate probabilities from samples, this is what we also currently do. It will be very confusing if we break this pattern.

> For example let's say i run a functional and my samples return: {0: 10} that doesn't mean that the probabilities are {0: 100%}

The probabilities based on your sampling are exactly {0: 100%}, if you need more precision you should increase the nshots. (that's the essence of the probabilistic nature of quantum computing)

If you want to cheat and get the _exact, ideal_ probabilities, you should calculate the state (if you can) and get probabilities from there. But this is simulator-only, small hilbert space-only.

src/qilisdk/readout/readout.py
amirazzam2000: if you run it on one of the children you will get this error message: 



```

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], [line 3](vscode-notebook-cell:?execution_count=1&line=3)
      1 from qilisdk.readout import ReadoutMethod, ExpectationReadout
----> [3](vscode-notebook-cell:?execution_count=1&line=3) ExpectationReadout.sample(100)

File ~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:76, in ReadoutMethod.sample(cls, nshots)
     59 @classmethod
     60 def sample(cls, nshots: int = 100) -> SamplingReadout:
     61     """Create a :class:`SamplingReadout`.
     62 
     63     This factory must be called on :class:`ReadoutMethod` itself, not on a
   (...)     74             :class:`ReadoutMethod`.
     75     """
---> [76](https://file+.vscode-resource.vscode-cdn.net/home/ameer/Documents/Qili/qilisdk/.tmp/~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:76)     cls._require_factory_called_on_base("sample")
     77     return SamplingReadout(nshots=nshots)

File ~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:55, in ReadoutMethod._require_factory_called_on_base(cls, factory_name)
     49 """Ensure factory helpers are only called on ``ReadoutMethod``.
     50 
     51 Raises:
     52     TypeError: if it's called on a child rather than the base class.
     53 """
     54 if cls is not ReadoutMethod:
---> [55](https://file+.vscode-resource.vscode-cdn.net/home/ameer/Documents/Qili/qilisdk/.tmp/~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:55)     raise TypeError(
     56         f"{factory_name}() is only available on ReadoutMethod; use ReadoutMethod.{factory_name}(...)"
     57     )

TypeError: sample() is only available on ReadoutMethod; use ReadoutMethod.sample(...)---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], [line 3](vscode-notebook-cell:?execution_count=1&line=3)
      1 from qilisdk.readout import ReadoutMethod, ExpectationReadout
----> [3](vscode-notebook-cell:?execution_count=1&line=3) ExpectationReadout.sample(100)

File ~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:76, in ReadoutMethod.sample(cls, nshots)
     59 @classmethod
     60 def sample(cls, nshots: int = 100) -> SamplingReadout:
     61     """Create a :class:`SamplingReadout`.
     62 
     63     This factory must be called on :class:`ReadoutMethod` itself, not on a
   (...)     74             :class:`ReadoutMethod`.
     75     """
---> [76](https://file+.vscode-resource.vscode-cdn.net/home/ameer/Documents/Qili/qilisdk/.tmp/~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:76)     cls._require_factory_called_on_base("sample")
     77     return SamplingReadout(nshots=nshots)

File ~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:55, in ReadoutMethod._require_factory_called_on_base(cls, factory_name)
     49 """Ensure factory helpers are only called on ``ReadoutMethod``.
     50 
     51 Raises:
     52     TypeError: if it's called on a child rather than the base class.
     53 """
     54 if cls is not ReadoutMethod:
---> [55](https://file+.vscode-resource.vscode-cdn.net/home/ameer/Documents/Qili/qilisdk/.tmp/~/Documents/Qili/qilisdk/src/qilisdk/readout/readout.py:55)     raise TypeError(
     56         f"{factory_name}() is only available on ReadoutMethod; use ReadoutMethod.{factory_name}(...)"
     57     )

TypeError: sample() is only available on ReadoutMethod; use ReadoutMethod.sample(...)

```

src/qilisdk/readout/readout.py
amirazzam2000: I kept the property in the functional results but started using the qtensor's probabilities method. the reason i kept is because the qtensor's probabilities have a different format to the one we use in the functional results. so for a consistent user experience i kept it there (basically calling the qtensor one and then reformatting it to the functional results format).

src/qilisdk/readout/readout_result.py
amirazzam2000: So the current behaviour is that we have two methods of constructing the sampling results. from state and from samples. 

- if you construct the samples from the state, the probabilities are computed from the qtensor and then sampled. 

- if you construct the samples from samples, then the probabilities are computed from the provided samples. (this is the default behavior for QiliSim and CudaBackend [when using noise models] as they provide the samples. But for QuTip we don't compute samples anymore.)

src/qilisdk/readout/readout_result.py
amirazzam2000: I don't know if it's a good idea to mix QTensor and the noise model. I think it's best to keep the interactions with the noise model only through the backend. But yeah if you think this will provide value we can add it as well.

src/qilisdk/functionals/variational_program_result.py
amirazzam2000: Variational Program is a super functional not a primitive one. these results are only for primitive funcitonals.

src/qilisdk/functionals/functional_result.py
Lumorti: I do agree that it seems somewhat redundant right now, but I could also imagine that in future we might want to include results other than just the readout (e.g. maybe we have different types of post-processing or variational-specific or reservoir-specific quantities).

src/qilisdk/functionals/functional_result.py
```diff
@@ -66,23 +64,53 @@ def intermediate_results(self) -> list[ReadoutCompositeResults]:
...
+        """
+        Outcome probabilities from the final execution step.
+
+        Raises:
+            ValueError: if sampling or state tomography readouts are not specified.
+        """
+        if has_sampling(self._readout_results):
+            return self._readout_results.sampling.probabilities
+        if has_state_tomography(self._readout_results):
+            return self._readout_results.state_tomography.probabilities
```
Lumorti, line 167: Urr technically 🤓☝️ we _could_ calculate the probabilities from the expectation values too

But that seems like such a specific case that would not be useful

src/qilisdk/functionals/functional_result.py
amirazzam2000: The idea also here is that the type of final result and the intermediate results is the same. and then we need a class to manage both, which is what this class is doing. Like it's a way to combine all the results and allow the user to access them in an easier way.

src/qilisdk/functionals/functional_result.py
amirazzam2000: I am not sure if you can do it always. Like you need the user to have measured a specific set of observables to be able to recreate the probabilities. but for an arbitrary set of observables and without accessing the state vector it doesn't seem doable to me. :thinking:

src/qilisdk_cpp/backends/qilisim/utils/sample.cpp
```diff
@@ -0,0 +1,145 @@
...
+                filtered_bitstring += bitstring[q];
+            }
+        }
+        filtered_counts[filtered_bitstring] += count;
+    }
+    return filtered_counts;
+}
+
+std::map<std::string, int> construct_samples(const DenseMatrix& state, int n_qubits, int n_shots, NoiseModelCpp& noise_model_cpp, const QiliSimConfig& config, const std::vector<bool>& qubits_to_measure){
+
```
Lumorti, line 113: Missing a doc string

src/qilisdk/readout/readout.py
Lumorti: Don't forget about this

src/qilisdk/functionals/digital_propagation.py
```diff
@@ -0,0 +1,96 @@
...
+        super().__init__()
+        self.circuit = circuit
+
+    def _iter_parameter_children(self) -> Iterator[Parameterizable]:
+        """Yield the circuit as the sole parameterizable child.
+
+        Yields:
+            Iterator[Parameterizable]: The underlying ``Circuit``.
+        """
+        yield self.circuit
```
Lumorti, line 62: General question, since every parameterizable has its children as an iterable, why do we need to specifically reimplement set_parameter_values etc. each time? Can't be just assume that, given a list of new parameters, in the base class we iterate through the children and set all parameter recursively? And that way each new parameterizable only needs to implement the iter_children and that's it

src/qilisdk/readout/readout.py
Lumorti: Agreed, this should be is_sampling_readout

Lumorti (changes requested):
Looking really good! A few tiny things still

src/qilisdk/functionals/digital_propagation.py
amirazzam2000: The issue was that some of the children need to reset some caches, if we don't overwrite the set_parameter method to reset these caches then it gets skipped. This was happening with interpolators. Here is the structure: 

analogEvolution -> schedule -> [interpolator, hamiltonian]. 

when I call get parameters on analogEvolution it goes through the children and collects all the parameters. and if i call set parameter on analogEvolution it does the same it goes through all the children and updates all the parameters. However, Interpolator uses lots of caching to make its operation faster. therefor everytime the parameters are updated we need to clear these caches. so now I overwrote the set_paprameter methods of AnalogEvolution and Schedule to be able to call delete_cache from schedule on the interpolators. then the actual update of the parameters happens through super().set_parameters which is the method specified in the parameterizable interface.

src/qilisdk/readout/readout.py
amirazzam2000: you can't do TypeGuard on self, you have to have take an input. so I added external helper methods that check this.

src/qilisdk/functionals/digital_propagation.py
Lumorti: I see I see, thanks for the explanation

src/qilisdk/readout/readout.py
Lumorti: I more meant the renaming, "is_expectation_values" is strange, it should either be "has_expectation_values" or "is_expectation_readout"

src/qilisdk/readout/readout.py
amirazzam2000: having it in both files is hard because of circular import. I added strong type checking to avoid this problem in runtime. However, the linters will still not flag this as an issue.

src/qilisdk/functionals/quantum_reservoirs_result.py
amirazzam2000: I missed this i am removing it now!

src/qilisdk/readout/readout.py
amirazzam2000: yes i changed them to: 
`is_sampling_readout`
`is_expectation_readout`
`is_state_tomography_readout`

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,463 @@
...
+
+class _HasStateTomography(Protocol):
+    state_tomography: StateTomographyReadoutResult
+
+
+def has_state_tomography(obj: ReadoutCompositeResults) -> TypeGuard[_HasStateTomography]:
+    return obj.state_tomography is not None
+
+
+@dataclass(frozen=True)
```
fedonman: Why is this class a `dataclass`?

src/qilisdk/backends/backend.py
```diff
@@ -184,9 +299,118 @@ def evaluate_sample(parameters: list[float]) -> float:
...
+        Args:
+            final_state (QTensor): The final quantum state to sample from.
+            readout (SamplingReadout): The sampling readout configuration.
+            seed (int | None): Optional random seed for reproducible sampling. Defaults to ``None``.
+            **kwarg (Any): Additional keyword arguments forwarded to then result constructor.
+
+        Returns:
+            SamplingReadoutResult: The constructed sampling result.
+        """
+        return SamplingReadoutResult.from_state(readout=copy(readout), state=final_state)
```
fedonman: 1. Remove `kwargs`, it is not used.
2. Why do we need to define class method for a one-liner? Can't you call the line directly in `_construct_results_list`? (it is only used there)

src/qilisdk/backends/backend.py
```diff
@@ -184,9 +299,118 @@ def evaluate_sample(parameters: list[float]) -> float:
...
+                configuration.
+            **kwarg (Any): Additional keyword arguments forwarded to the
+                result constructor.
+
+        Returns:
+            ExpectationReadoutResult: The constructed expectation result.
+        """
+        # Pre-scale on the original so the copy inherits cached qtensor_observables
+        readout.expand_observables(nqubits=final_state.nqubits)
+        return ExpectationReadoutResult.from_state(readout=copy(readout), state=final_state)
```
fedonman: Same here:

1. Remove `kwargs`, it is not used.
2. Why do we need to define class method for a one-liner? Can't you call the line directly in `_construct_results_list`? (it is only used there)

src/qilisdk/backends/backend.py
```diff
@@ -184,9 +299,118 @@ def evaluate_sample(parameters: list[float]) -> float:
...
+            readout (StateTomographyReadout): The state-tomography readout
+                configuration.
+            **kwarg (Any): Additional keyword arguments forwarded to the
+                result constructor.
+
+        Returns:
+            StateTomographyReadoutResult: The constructed state-tomography
+                result.
+        """
+        return StateTomographyReadoutResult(readout=copy(readout), state=final_state)
```
fedonman: Same here:

1. Remove `kwargs`, it is not used.
2. Why do we need to define class method for a one-liner? Can't you call the line directly in `_construct_results_list`? (it is only used there)

src/qilisdk/backends/backend.py
```diff
@@ -184,9 +299,118 @@ def evaluate_sample(parameters: list[float]) -> float:
...
+
+    @classmethod
+    def _construct_results_list(
+        cls, final_state: QTensor, readout: list[ReadoutMethod], seed: int | None = None, **kwarg: Any
```
fedonman: Remove `kwargs`, it is not used.

Lumorti (approved):
Perfect, thanks for the changes! This is a really nice PR overall, and might finally be the thing that gets Qili people using QiliSDK!

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,463 @@
...
+        samples: dict[str, int] = _samples_from_probabilities(probabilities, nshots=readout.nshots)
+        return cls(readout=readout, samples=samples, probabilities=probabilities)
+
+    def __init__(
+        self, readout: SamplingReadout, *, samples: dict[str, int] | None, probabilities: dict[str, float] | None = None
+    ) -> None:
+        if not isinstance(readout, SamplingReadout):
+            raise TypeError(f"Expected SamplingReadout, got {type(readout).__name__}")
+        if samples is None:
+            raise ValueError("Can't construct the Sampling results if samples are not provided.")
```
fedonman: It doesn't make sense to allow `None` for `samples` and then immediately raise error if None. Remove the `| None` possibility for the `samples` parameter, and also the `if samples is None` check.

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,463 @@
...
+        self._samples: dict[str, int] = samples or {}
+        self._probabilities: dict[str, int | float] = probabilities or {}
```
fedonman:
```suggestion
        self._probabilities: dict[str, float] = probabilities or {}
```

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,463 @@
...
+            expected_values: list[int | float] = [
+                _assert_real((expect_val(o, state))) for o in readout.qtensor_observables
+            ]
+        except ValueError:
+            raise ValueError(
+                "Encountered an imaginary expected value while computing the expectation values, try reducing the total tolerance or improving simulation precision."
+            )
+        return cls(readout=readout, expected_values=expected_values)
+
+    def __init__(self, readout: ExpectationReadout, expected_values: list[float]) -> None:
```
fedonman: Should it be `expected_values` or `expectation_values`? The documentation and `__repr__` method say expectation values.

src/qilisdk/readout/readout.py
fedonman: Forgot this.

src/qilisdk/readout/readout_result.py
```diff
@@ -0,0 +1,463 @@
...
+        if not samples and not probabilities:
+            raise ValueError("can't initialize Sampling Results if both samples and probabilities are not provided.")
+        if not probabilities:
+            bitstrings = list(samples.keys())
+            nqubits = len(bitstrings[0])
+            if not all(len(bitstring) == nqubits for bitstring in bitstrings):
+                raise ValueError("Not all bitstring keys have the same length.")
+
+            # Calculate probabilities
+            probabilities: dict[str, int | float] = {
```
fedonman:
```suggestion
            probabilities: dict[str, float] = {
```

src/qilisdk/readout/readout_result.py
fedonman:
```suggestion
        self._probabilities: dict[str, float] = probabilities or {}
```

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,260 @@
...
+        if len(positional_args) > len(positional_names):
+            expected = ", ".join(positional_names)
+            raise TypeError(
+                f"{class_name} expects at most {len(positional_names)} positional argument(s): ({expected})."
+            )
+        merged = dict(keyword_args)
+        for name, value in zip(positional_names, positional_args):
+            if name in merged:
+                raise TypeError(f"{class_name} got multiple values for '{name}' (positional and keyword).")
+            merged[name] = value
```
fedonman: This is not used anywhere. Remove it.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,260 @@
...
+    def _require_factory_called_on_base(cls, factory_name: str) -> None:
+        """Ensure factory helpers are only called on ``ReadoutMethod``.
+
+        Raises:
+            TypeError: if it's called on a child rather than the base class.
+        """
+        if cls is not ReadoutMethod:
+            raise TypeError(
+                f"{factory_name}() is only available on ReadoutMethod; use ReadoutMethod.{factory_name}(...)"
+            )
```
fedonman: `ReadoutMethod.sampling(...)`, `ReadoutMethod.expectation(...)`, etc. use `_require_factory_called_on_base` to raise `TypeError` at runtime if called on a subclass. This is an anti-pattern:

  - IDE autocompletion shows `SamplingReadout.sampling()` as valid
  - The user can't discover at write-time that it's forbidden.

 Alternative: Make them module-level functions or use `__init_subclass__` to delete factory methods from children.
  
Or simply drop the factories -> `SamplingReadout(nshots=100)` is already clean enough; the factories add no value.

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,260 @@
...
+
+from typing import Literal
+
+from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_validator
+
+from qilisdk.analog import Hamiltonian  # noqa: TC001
+from qilisdk.core import QTensor
+
+
+class ReadoutMethod(BaseModel):
```
fedonman: Why this class hierarchy is a Pydantic model?

src/qilisdk/readout/readout.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Args:
+            nshots (int): Number of measurement shots.  Must be >= 0.
+
+        Returns:
+            SamplingReadout: A configured sampling readout instance.
+
+        Raises:
+            AttributeError: If called on a subclass instead of
+                :class:`ReadoutMethod`.
```
fedonman: General comment for all docstrings: Try and correct this <80 chars new line formatting we got.

fedonman (changes requested):
Left some comments for the files I had time to check, will continue.

src/qilisdk/readout/readout.py
amirazzam2000: they are just data storage classes

src/qilisdk/readout/readout_result.py
amirazzam2000: it only stores data and has no logic

src/qilisdk/backends/backend.py
amirazzam2000: these methods are there because initially CudaBackend was modifying how the output is being written for a specific readout. And I thought it would be useful in general to allow the backends more flexibility to modify how they write the results. Now it's not being used if you think we should remove them i can do it no problem. what do you think?

src/qilisdk/readout/readout_result.py
amirazzam2000: so it's expectation readout but the attribute is expected values. I feel like expectation values is weird, i am more used to seeing it as expected values.

amirazzam2000: but i will change it to expectation because actually now it's confusing

src/qilisdk/backends/backend.py
fedonman: I would remove them, they are all 1-liners that are used once, no need for a method.

src/qilisdk/readout/readout_result.py
fedonman: It feels weird to use 1 dataclass amongst many normal python classes in the same hierarchy. The same goes with the Pydantic Model. It is usually a "all-or-none" scenario. It also might affect serialization (the format of serialized string) because pydantic models and dataclasses behave differently than normal classes.

It's not critical, but I would change both to normal python class.

src/qilisdk/backends/backend.py
```diff
@@ -123,82 +115,29 @@ def execute(self, functional: Functional, readout: ReadoutMethod | list[ReadoutM
...
+            FunctionalResult: the final quantum reservoir execution results.
 
         Raises:
-            ValueError: If the quantum state cannot be repaired to a valid
-                density matrix between layers.
-
-        Notes:
-            Noise models are not supported with quantum reservoirs and will
-            be silently ignored if one is set on the backend.
+            ValueError:if throughout the execution the state becomes invalid due to numerical instabilities.
```
Lumorti, line 140: Should this value error be moved to the raises section?

src/qilisdk/readout/readout_result.py
```diff
@@ -441,6 +310,151 @@ def __repr__(self) -> str:
...
+
+    def __repr__(self) -> str:
+        out = ""
+        if self.sampling:
+            out += str(self.sampling)
+        if self.expectation_values:
+            out += str(self.expectation_values)
+        if self.state_tomography:
+            out += str(self.state_tomography)
+        return out or f"{type(self).__name__}(empty)"
```
Lumorti, line 450: Does this look nice when printed? Because I notice it doesn't have any ReadoutResults(...) or newlines like a lot of our other __repr__s

src/qilisdk/readout/readout_spec.py
```diff
@@ -0,0 +1,187 @@
...
+    """
+
+    def __init__(self: ReadoutSpec[None, None, None]) -> None:
+        self._sampling: SamplingReadout | None = None
+        self._expectation: ExpectationReadout | None = None
+        self._state_tomography: StateTomographyReadout | None = None
+
+    # -- builder methods -- each one transforms one type parameter --------
+
+    def with_sampling(self, nshots: int = 100) -> ReadoutSpec[SamplingReadoutResult, E, T]:
```
Lumorti: Is 100 the default everywhere for shots? Would 1000 be more reasonable default?

generic-functional-result-design.md
Lumorti, line 1: Should this still be in the base of the main repo? Wasn't this just for internal use?

Maybe we should make a little folder in the repo for internal docs?

Lumorti (approved):
Approved, but have a look at the minor comments, just in case.

Good job on the big PR!!

src/qilisdk/readout/readout_result.py
amirazzam2000: it looks like this: 

```
Sampling Results: (
	nshots=70,
	samples={'0': 30, '1': 40}
)

Expectation Value Results: (
	expectation_values=[0, 1, 2],
)
```

the new lines come from the the other readout results prints.

generic-functional-result-design.md
amirazzam2000: I removed it!

src/qilisdk/readout/readout_spec.py
amirazzam2000: changed it :+1:

src/qilisdk/backends/backend.py
amirazzam2000: it is in a Raises section, no?

fedonman (approved):
Amazing work!! 💯 💯 This refactoring will indeed move the library forward. Thanks a lot @amirazzam2000 🚀

src/qilisdk/backends/backend.py
Lumorti: Ah oops nevermind

PR #177, opened 2026-03-19
[SDK-248] Cache Matrix of Gates

fedonman:
`Gate.matrix` is now a cached property instead of a stored instance attribute. This considerably reduces serialization size by excluding the matrix from YAML output.

### DIFF
--- a/changes/177.feature.md
+++ b/changes/177.feature.md
@@ -1,1 +1,1 @@
-Gate matrix is now a cached property instead of a stored instance attribute. This considerably reduces serialization size by excluding the matrix from YAML output.
+`Gate.matrix` is now a cached property instead of a stored instance attribute. This considerably reduces serialization size by excluding the matrix from YAML output.
--- a/changes/177.feature.md
+++ b/changes/177.feature.md
@@ -1,0 +1,1 @@
+Gate matrix is now a cached property instead of a stored instance attribute. This considerably reduces serialization size by excluding the matrix from YAML output.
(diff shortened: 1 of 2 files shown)


changes/177.feature.md
```diff
@@ -0,0 +1 @@
+Gate matrix is now a cached property instead of a stored instance attribute. This considerably reduces serialization size by excluding the matrix from YAML output.
```
Lumorti: Maybe put the dot to show that we mean matrix of a gate, not that it's a gate matrix being cached in some other class etc.
```suggestion
Gate.matrix is now a cached property instead of a stored instance attribute. This considerably reduces serialization size by excluding the matrix from YAML output.
```

src/qilisdk/digital/gates.py
```diff
@@ -288,29 +288,32 @@ def __init__(
...
-        return self._matrix
+        return self._generate_matrix()
 
     @property
     def target_qubits(self) -> tuple[int, ...]:
         return self._target_qubits
 
+    def _invalidate_matrix(self) -> None:
+        self.__dict__.pop("matrix", None)
+
```
Lumorti, line 301: Probably we should add some tests for this

Lumorti (changes requested):
Very nice simplification! We should use this cached_property more in future.

Left a few minor comments.

src/qilisdk/digital/gates.py
fedonman: It's already been tested in `test_gates.py` through the set/get methods.

changes/177.feature.md
fedonman: Done.

amirazzam2000 (approved):
this is a very cool implementation of it! Thanks vyron!

Lumorti (approved):
Nice job!

PR #178, opened 2026-03-24
[SDK-247] OpenQASM 3.0 Support

Lumorti:
Support has been added for OpenQASM 3.0, allowing users to import/export from strings/files using the OpenQASM 3.0 standard. Usage is the same as with the OpenQASM 2.0 interface:

```
from qilisdk.utils.openqasm3 import from_qasm3_file, from_qasm3, to_qasm3, to_qasm3_file

circuit = from_qasm3_file("input.qasm")
to_qasm3_file(circuit, "output.qasm")
as_string = to_qasm3(circuit)
from_string = from_qasm3(as_string)
```


src/qilisdk/digital/gates.py
```diff
@@ -438,7 +451,13 @@ def is_modified_from(self, gate_type: type[TBasicGate]) -> bool:
 
 @yaml.register_class
 class Controlled(Modified[TBasicGate]):
-    def __init__(self, *control_qubits: int, basic_gate: TBasicGate) -> None:
+    def __init__(self, *control_qubits: int, basic_gate: Gate) -> None:
+
+        # If doing Controlled of another Controlled, combine into one with all control qubits.
+        if isinstance(basic_gate, Controlled) and isinstance(basic_gate.basic_gate, Gate):
```
amirazzam2000, line 457: shouldn't you check that both gates are the same? like what if I have a CZ controlling a CX then things get weird.

src/qilisdk/digital/gates.py
```diff
@@ -482,7 +501,7 @@ class Adjoint(Modified[TBasicGate]):
     Represents the adjoint (conjugate transpose) of a unitary gate.
     """
 
-    def __init__(self, basic_gate: TBasicGate) -> None:
+    def __init__(self, basic_gate: Gate) -> None:
```
amirazzam2000, line 504: why did we do this change?

src/qilisdk/digital/gates.py
```diff
@@ -438,7 +451,13 @@ def is_modified_from(self, gate_type: type[TBasicGate]) -> bool:
 
 @yaml.register_class
 class Controlled(Modified[TBasicGate]):
-    def __init__(self, *control_qubits: int, basic_gate: TBasicGate) -> None:
+    def __init__(self, *control_qubits: int, basic_gate: Gate) -> None:
```
amirazzam2000, line 454: I wouldn't allow for the basic_gate to be a generic Gate otherwise you need to add a bunch of checks and i am not sure if our transpilers can handle that.

src/qilisdk/utils/openqasm3.py
```diff
@@ -0,0 +1,1221 @@
...
+     - `from qilisdk.utils.openqasm3 import to_qasm3`
+     - `from qilisdk.utils.openqasm3 import from_qasm3`
+
+    Understanding the following table:
+    - ✅: The feature is fully supported
+    - 🟡: The feature is partially supported, see the note for explanation
+    - ❌: The feature is not supported at all
+
+    Feature Table
+    OpenQASM 3 Feature     Qiskit SDK       IBM Qiskit Runtime      QiliSDK       Notes
```
amirazzam2000: remember to update this table and remove Qiskit

amirazzam2000 (changes requested):
thanks luke!!

src/qilisdk/digital/gates.py
Lumorti: It's to allow controlled controlled gates (which just get merged into a single controlled gate).

If someone wants to do controlled(controlled(adjoint(gate))), as they can in OpenQASM 3.0, we should allow that. The generate_matrix still works as expected, since all they do is add controls or change the base matrix.

src/qilisdk/digital/gates.py
Lumorti: How would that happen? This is for handling Controlled(0, basic_gate=Controlled(1, basic_gate=X(2))).

Or as it would more commonly happen:
```
for i in [0,1,2]:
    gate = gate.controlled(i)
```

src/qilisdk/digital/gates.py
Lumorti: See below comment

src/qilisdk/utils/openqasm3.py
Lumorti: Done!

src/qilisdk/digital/gates.py
amirazzam2000: Hmmmm I see yes it's fine.

src/qilisdk/digital/gates.py
amirazzam2000: perfect! thanks luke!

amirazzam2000 (approved):
Thank you Luke! looks good from my side

PR #179, opened 2026-04-01
[SDK-250] Add QTensor Draw Method

Lumorti:
Added a method to plot a single-qubit QTensor state on the Bloch sphere:
```
state = QTensor.ket(0)
state.draw()
```


tests/unit_python/core/test_qtensor.py
```diff
@@ -125,11 +126,11 @@ def test_ptrace_valid():
     reduced_ket_qubit_1 = qket.ptrace(keep=[1])
     reduced_ket_qubit_0 = qket.ptrace(keep=[0])
 
-    qket = ket(0, 1, 0, *[0 for _ in range(20)])
+    qket = ket(0, 1, 0, *[0 for _ in range(5)])
     reduced_ket_qubit_1_big = qket.ptrace(keep=[1])
     reduced_ket_qubit_0_big = qket.ptrace(keep=[0])
 
-    qket = bra(0, 1, 0, *[0 for _ in range(20)])
+    qket = bra(0, 1, 0, *[0 for _ in range(5)])
```
Lumorti, line 132: Note for reviewers: These tests were taking like 10 seconds each, I guess because it was a 23 qubit state, to be looked at in future when we improve QTensor. But for now just reduced it to 8 qubits so it's fast again.

Lumorti:
Note for reviewers: here's what it looks like, in the live version you can rotate it with the mouse

image

src/qilisdk/core/qtensor.py
```diff
@@ -764,6 +766,77 @@ def __setstate__(self, state: dict[str, csr_matrix]) -> None:
         """
         self.__init__(state["data"])
 
+    def draw(self) -> None:
```
fedonman: I would consider moving the implementation into a dedicated `QTensorRenderer`, similar to how `Circuit` and `Schedule` are drawn. E.g. for Circuit:

```
    def draw(self, style: CircuitStyle = CircuitStyle(), filepath: str | None = None) -> None:
        """
        Render this circuit with Matplotlib and optionally save it to a file.

        The circuit is rendered using the provided style configuration. If ``filepath`` is
        given, the resulting figure is saved to disk (the output format is inferred
        from the file extension, e.g. ``.png``, ``.pdf``, ``.svg``).

        Args:
            style (CircuitStyle): Visual style configuration applied to the plot.
                If not provided, the default :class:`CircuitStyle` is used.
            filepath (str | None): Destination file path for the rendered figure.
                If ``None``, the figure is not saved.
        """
        from qilisdk.utils.visualization.circuit_renderers import MatplotlibCircuitRenderer  # noqa: PLC0415

        renderer = MatplotlibCircuitRenderer(self, style=style)
        renderer.plot()
        if filepath:
            renderer.save(filepath)
```

This way, the user can change styling or save to file in a unified way.

amirazzam2000: I agree with this comment from vyron!

amirazzam2000 (changes requested):
Looks amazing Luke!! thank you 🙌

src/qilisdk/core/qtensor.py
Lumorti: Honestly I expected this comment, but my logic was is that this is small function without customization. Circuit drawing is very customizable etc., meanwhile if I made a QTensorRenderer it would just be this single function wrapped in a class

amirazzam2000: well but you can add themes, so like the other two drawers you can have dark and light themes that change the colors of the backgrounds and allow the user to select the colors of the lines etc. it adds more customizability to the plots

Lumorti: Moved it to a separate renderer, and added some style choices

amirazzam2000 (approved):
Amazing! thanks Luke. 

If you can add some documentation for how to use this in the docs (and maybe show examples of the usage of different styles)

Lumorti:
Added more docs, here's the render:

image

PR #182, opened 2026-04-08
HOTFIX - Fix Issues with the Multiversion Build

Lumorti:
Fixed a bug when building the docs. Versions of sphinx >= 9 cause issues with the newest (0.2.4) version of sphinx-multiversion, so now we limit the version of sphinx to be less than 9.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -99,12 +99,12 @@ docs = [
     "astroid<3.2",
     "nbsphinx>=0.9.7",
     "pypandoc>=1.15",
-    "sphinx>=8.2.3",
+    "sphinx==8.2.3",
     "sphinx-autoapi>=3.4.0",
     "sphinx-automodapi>=0.20.0",
     "sphinx-design>=0.7.0",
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -99,12 +99,12 @@ docs = [
     "astroid<3.2",
     "nbsphinx>=0.9.7",
     "pypandoc>=1.15",
-    "sphinx>=8.2.3",
+    "sphinx==8.2.3",
```
amirazzam2000, line 102: why do we need to fix the sphinx version?

amirazzam2000 (approved):
Thanks for finding the error luke! I will approve it but i left one small question.

pyproject.toml
Lumorti: It's sphinx that breaks it, specifically sphinx >= 9.0.0

sphinx-multiversion isn't being updated I think

Lumorti: image

We are now part of the list of people who have done the same fix haha

PR #183, opened 2026-04-08
[SDK-157] Docs Revamp

Lumorti:
Added a "Tutorials" section to the documentation, covering some quantum basics as well as a number of useful examples.
The rest of the documentation has also been proof-read and adjusted.


src/qilisdk/analog/schedule.py
Lumorti, line 1: Note for reviewers: All changes to the `src/` are to fix docs errors. 

In this case `PARAMETERIZED_NUMBER` was defined in multiple places so I moved it to a central location (it can't be moved to types because that would cause a circular import since it needs `Parameter`).

scripts/docs.sh
```diff
@@ -0,0 +1,101 @@
...
+FAIL=0
+SKIP=0
+
+# Save the current directory so we can return to it later
+ORIG_DIR=$(pwd)
+
+# Change to the root directory of the project
+cd "$(dirname "$(dirname "$(realpath "$0")")")"
+
+TMPFILE=$(mktemp /tmp/docs_test_XXXXXX.py)
```
DZPM, line 20: Why the `_XXXXXX`?

DZPM:
Thanks for the dark mode and for removing the "em dashes"

I'm worried about the README.md and the installation.rst slowly diverging over time, like:
image
image

docs/getting_started/installation.rst
```diff
@@ -39,13 +39,68 @@ You can combine extras:
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The above method will install the pre-built binaries of QiliSDK of the most recent release. 
+If you want to get the newest features (those that we're still working on), you can compile the library from source, 
+but be aware that this is not recommended for most users, and we make no promises that the latest code will be stable.
+
+Support for Windows is somewhat limited, so we recommend using WSL, which can be installed as per 
+`this guide <https://learn.microsoft.com/en-us/windows/wsl/install>`__.
+
+First, make sure you have a Python, pip, git, uv and build-essentials installed. For Ubuntu/Debian, you can run:
```
amirazzam2000: maybe we can write a note or something for mac users to check that they have c++ compiler installed in case it doesn't come built with macs out of the box

src/qilisdk/core/interpolator.py
```diff
@@ -18,20 +18,19 @@
...
 from copy import copy
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Mapping, TypeAlias
+from typing import TYPE_CHECKING, Any, Mapping
 
 import numpy as np
 
 from qilisdk.core.parameterizable import Parameterizable
-from qilisdk.core.variables import LEQ, BaseVariable, Parameter, Term
+from qilisdk.core.variables import LEQ, PARAMETERIZED_NUMBER, BaseVariable, Parameter, Term
```
amirazzam2000: I would move PARAMETERIZED_NUMBER to core/types.py so that it's properly shared

docs/tutorials/applications/application_annealing.rst
```diff
@@ -0,0 +1,151 @@
...
+The hope is that if we do this slowly enough, the system will remain in its ground state throughout the evolution, 
+and thus end up in the ground state of the final Hamiltonian, which corresponds
+to the optimal solution of our problem.
+
+Assuming we have reformulated our problem into a QUBO, we can construct the problem Hamiltonian for quantum annealing,
+which is given by:
+
+.. math:: 
+
+    H_{prob} = \sum_{i,j} c_{ij} Z(i) Z(j)
```
amirazzam2000, line 34: we should have some explanation on how to transform from classical binary variables [0, 1] to spin variables  [-1, 1]. In this case you end up with the same expression but not always. you need to do the change of variables: x = (1 - z) / 2

docs/tutorials/applications/application_annealing.rst
```diff
@@ -0,0 +1,151 @@
...
+    preferences = [[0, 1, 3, 4],
+                  [1, 0, 5, 2],
+                  [3, 5, 0, 6],
+                  [4, 2, 6, 0]]
+    model = QUBO("team_formation_example")
+    team_1 = sum(preferences[i][j] * vars[i] * vars[j] for i in range(num_people) for j in range(i+1, num_people))
+    team_0 = sum(preferences[i][j] * (1 - vars[i]) * (1 - vars[j]) for i in range(num_people) for j in range(i+1, num_people))
+    model.set_objective(team_0 + team_1, label="obj", sense=ObjectiveSense.MAXIMIZE)
+    model.add_constraint("team_size_constraint", EQ(sum(vars[i] for i in range(num_people)), 2), lagrange_multiplier=10)
+
```
amirazzam2000, line 62: can we add spaces in this code to make it more readable. 
```suggestion
    model = QUBO("team_formation_example")
    
    team_1 = sum(
        preferences[i][j] * vars[i] * vars[j] 
        for i in range(num_people) 
        for j in range(i+1, num_people)
    )
    
    team_0 = sum(
        preferences[i][j] * (1 - vars[i]) * (1 - vars[j]) 
        for i in range(num_people) 
        for j in range(i+1, num_people)
    )
    
    model.set_objective(
        team_0 + team_1, 
        label="obj", 
        sense=ObjectiveSense.MAXIMIZE
    )
    
    model.add_constraint(
        "team_size_constraint", 
        EQ(sum(vars[i] for i in range(num_people)), 2), 
        lagrange_multiplier=10
    )
```

docs/tutorials/applications/application_annealing.rst
```diff
@@ -0,0 +1,151 @@
...
+                        "problem": {(0, T): lambda t: t / T},
+                    },
+                    dt=0.01,
+                )
+
+We will start in the ground state of our mixing Hamiltonian, which is the equal superposition state over all possible team assignments:
+
+.. code-block:: python
+
+    initial_state = (QTensor.ket(1, 0, 0, 0) + QTensor.ket(0, 1, 0, 0) + QTensor.ket(0, 0, 1, 0) + QTensor.ket(0, 0, 0, 1)).normalized()
```
amirazzam2000: we can represent these states in a much simpler way: 
```
tensor_prod([ket(0) + ket(1) for _ in range(n_qubits]).unit()
```

docs/tutorials/applications/application_qaoa.rst
```diff
@@ -0,0 +1,169 @@
...
+    from qilisdk.core.model import QUBO, ObjectiveSense
+    from qilisdk.core.variables import BinaryVariable, EQ
+
+    num_people = 4
+    vars = [BinaryVariable(f"x{i}") for i in range(num_people)]
+    preferences = [[0, 1, 3, 4],
+                  [1, 0, 5, 2],
+                  [3, 5, 0, 6],
+                  [4, 2, 6, 0]]
+    model = QUBO("team_formation_example")
```
amirazzam2000: same comment here

docs/modules/noise_models/noise_models.rst
amirazzam2000: can we rename these to Digital Propagation and Analog Evolution?

docs/modules/speqtrum/speqtrum.rst
```diff
@@ -88,6 +92,7 @@ decoded strings or structured :class:`~qilisdk.speqtrum.speqtrum_models.ExecuteR
 When you wait on a :class:`~qilisdk.speqtrum.speqtrum_models.JobHandle`, the returned object is a
 :class:`~qilisdk.speqtrum.speqtrum_models.TypedJobDetail` that exposes a strongly typed :meth:`~qilisdk.speqtrum.speqtrum_models.TypedJobDetail.get_results` helper.
 
+.. SKIP
```
amirazzam2000, line 95: what are all these skips?

amirazzam2000 (changes requested):
Amazing work Luke!! This is a massive update to the documentation, thanks a lot ❤️

src/qilisdk/core/interpolator.py
Lumorti: That was my first thought, but that causes a dependency loop since we need Parameter

docs/modules/speqtrum/speqtrum.rst
Lumorti: These are for the doc checker script, it's a comment which says "don't worry that this block isn't executable". I put them if the code block uses something like "functional = DigitalPropogation(...)" which obviously wouldn't run but it's also not an error

src/qilisdk/core/interpolator.py
amirazzam2000: okay in that case i would leave PARAMETERIZED_NUMBER in interpolator and import it in schedule from interpolator. i don't think it needs to be in variable, also it's at the end of the file which is not ideal.

docs/modules/speqtrum/speqtrum.rst
amirazzam2000: we should keep them in mind whenever we update the interface to check that the code in the docs is still valid.

docs/getting_started/installation.rst
Lumorti: This is in build-essentials, along with cmake and make

docs/tutorials/applications/application_annealing.rst
Lumorti: added

docs/tutorials/applications/application_annealing.rst
Lumorti: Done!

docs/tutorials/applications/application_annealing.rst
Lumorti: Good catch!

scripts/docs.sh
Lumorti: It acts as template to put random numbers:

image

Lumorti:
> I'm worried about the README.md and the installation.rst slowly diverging over time

Yes I completely get that, I also don't like the duplication. Maybe the README should just be a big button that sends you to the docs, and then apart from that just has some developer-specific guides as we already have (like how to use ruff/ty/towncrier/tests).

src/qilisdk/core/interpolator.py
Lumorti: Done!

docs/modules/noise_models/noise_models.rst
Lumorti: Sure

NataliaSirera (approved):

NataliaSirera:
Great job! Thank you!!

docs/getting_started/installation.rst
amirazzam2000: yes but build essentials is not a MacOS package. So we should mention what they need to install to have a C++ compiler. I found this stack overflow that might be relevant:
https://stackoverflow.com/questions/38086451/how-to-install-build-essential-on-mac

README.md
```diff
@@ -437,7 +71,6 @@ This section covers how to set up a local development environment for qilisdk, r
 
 - Python **3.10+** (we test against multiple versions, but 3.10 is the minimum for local dev).
 - [Git](https://git-scm.com/) for version control.
-- [uv](https://pypi.org/project/uv/) for dependency management.
```
amirazzam2000, line 440: why did we remove this? UV is our official tool for installing the library/managing dependencies. I wouldn't remove it.

Lumorti: Have a read of that section. We install uv as part of the install instructions. This list is the list of things the user has to have installed before they can follow the steps, which does not include uv

docs/getting_started/installation.rst
Lumorti: I spoke to Natalia about it, and she said that apparently mac ships with a C++ compiler and she didn't need to install anything I think, it's just that she was getting this bug from before that should hopefully be fixed

amirazzam2000: okay amazing! thanks :pray:

README.md
amirazzam2000: but you still use the uv instructions later on in the section, but now the user is unaware what is uv

README.md
```diff
@@ -473,10 +106,6 @@ This section covers how to set up a local development environment for qilisdk, r
      ```bash
      uv sync --all-extras --all-groups
```
amirazzam2000: This command is wrong since we have cuda13 and 12 support, it should be: 

`uv sync --extra all-cu13 --all-groups` for cuda 13
`uv sync --extra all-cu12 --all-groups` for cuda 12

README.md
amirazzam2000: but i see what you mean i guess because it's mentioned above we can skip the repetition.

changes/143.feature.md
```diff
@@ -4,12 +4,13 @@ Example:
...
 ```python
 from qilisdk.analog import Schedule, Z
-from qilisdk.backends import QutipBackend
+from qilisdk.backends import QiliSim
 from qilisdk.core import ket
 from qilisdk.digital import Circuit, RX
+from qilisdk.readout import Readout
 
-theta = ReservoirInput("theta", 0.0)
```
amirazzam2000, line 12: why did you change the name from theta? this is not a gate and it should be fine to call it theta, theta_1 seems weird given that there is no theat_2

amirazzam2000 (approved):
thanks a lot luke! this is a massive change great job! 👏 🙏

changes/143.feature.md
Lumorti: It was giving an error, changing it to theta 1 fixed the error, I'll double check why now

amirazzam2000: ohhhhh i think i know why, it's because the theta is getting mixed up with the gate's theta parameter. okay let's rename it to res_in maybe? and we open an issue in linear

PR #184, opened 2026-04-08
[HOTFIX] Fix the Previous Hotfix Regarding Docs Building

Lumorti:
This hotfix fixes a previous error introduced by the last hotfix. Now that we're on the older version of sphinx we don't need the `app.config.viewcode_enable_epub` arg anymore.

I just ran this locally and it builds both the normal docs and the multiversion fine.


amirazzam2000 (approved):
good catch! Thanks luke

PR #185, opened 2026-04-16
[SDK-252] Mid-Circuit Measurements

Lumorti:
QiliSim now supports mid-circuit measurements. Additionally, partial samples are now returned as "11_0", where a "_" indicates that the qubit wasn't measured. Usage:

```python
from qilisdk.digital import Circuit, X, M

c = Circuit(2)
c.add(X(0))
c.add(M(0))
c.add(X(0))
c.add(M(0))
c.add(M(1))

results = QiliSim().execute(DigitalPropagation(c), Readout().with_sampling(1000))
print(results)
```


scripts/install_dev.sh
Lumorti, line 1: (clears the build cache, helps prevent some errors with this)

Lumorti:
(Codecov is saying there are lines not covered by tests, but then if you actually click on it to see which lines, you see everything is green and the overall coverage is INCREASED by this patch)

src/qilisdk/backends/qutip_backend.py
```diff
@@ -133,6 +136,16 @@ def _execute_digital_propagation(
         measurements_set = {q for g in functional.circuit.gates if isinstance(g, M) for q in g.qubits}
         measurements = sorted(measurements_set)
 
+        # Check for mid-circuit measurements
+        for i in range(len(functional.circuit.gates)):
+            gate = functional.circuit.gates[i]
+            if isinstance(gate, M):
+                for later_gate in functional.circuit.gates[i + 1 :]:
```
amirazzam2000: maybe for efficiency we can do something like: 
`
if any(not isinstance(later_gate, M) for later_gate in funcitional.circuit.gates[i+1:])
`
instead of an explicit loop. this can be a little more efficient.

src/qilisdk/readout/readout_result.py
```diff
@@ -100,32 +100,137 @@ class SamplingReadoutResult(ReadoutResult[SamplingReadout]):
...
+            if nqubits_samples >= nqubits_total:
+                return cls._filter_samples(samples_to_filter, qubits_to_measure)
+            # Assuming 3 qubits, if asked for 0 and 1 and we have xx, return xx_
+            if nqubits_samples < nqubits_total:
+                return cls._expand_samples(samples_to_filter, qubits_to_measure, nqubits_total)
+        return samples_to_filter
+
     @classmethod
-    def from_samples(cls, samples: dict[str, int]) -> Self:
+    def from_samples(
```
amirazzam2000, line 195: should we check the case where len(bitstring)  is less than nqubits and qubits to measure is not specified?

src/qilisdk/readout/readout_result.py
```diff
@@ -439,6 +544,64 @@ def from_dict(cls, data: dict) -> ReadoutCompositeResults:
 
         return cls(sampling=sampling, expectation_values=expectation_values, state_tomography=state_tomography)
 
+    def get_samples(self) -> dict[str, int]:
```
amirazzam2000, line 576: these operations are not type safe, can you take a look at how they are implemented in functional results and try to do something similar?

src/qilisdk/functionals/functional_result.py
```diff
@@ -203,67 +203,75 @@ def get_expectation_values(self) -> list[float]:
 
     # -- intermediate accessors (runtime-checked) -------------------------
 
-    @property
-    def intermediate_samples(self) -> list[dict[str, int]]:
-        """Measurement samples for every time-step (intermediate + final).
+    def get_intermediate_samples(self) -> list[dict[str, int]]:
```
amirazzam2000, line 210: can you implement these in a type safe way like the other get methods?

src/qilisdk/functionals/functional_result.py
```diff
@@ -203,67 +203,75 @@ def get_expectation_values(self) -> list[float]:
...
+            for res in self._intermediate_results:
                 if has_sampling(res):
                     results.append(res.sampling.samples)
             return results
         raise ValueError("Can't find intermediate samples because intermediate Results were not stored.")
 
-    @property
-    def intermediate_probabilities(self) -> list[dict[str, float]]:
-        """Outcome probabilities for every time-step (intermediate + final).
+    def get_intermediate_probabilities(self) -> list[dict[str, float]]:
```
amirazzam2000, line 234: same here

src/qilisdk/functionals/functional_result.py
```diff
@@ -203,67 +203,75 @@ def get_expectation_values(self) -> list[float]:
...
                     results.append(res.state_tomography.probabilities)
                 elif has_sampling(res):
                     results.append(res.sampling.probabilities)
             return results
         raise ValueError("Can't find intermediate probabilities because intermediate Results were not stored.")
 
-    @property
-    def intermediate_states(self) -> list[QTensor]:
-        """Quantum state vectors for every time-step (intermediate + final).
+    def get_intermediate_states(self) -> list[QTensor]:
```
amirazzam2000, line 258: same here

src/qilisdk/functionals/functional_result.py
```diff
@@ -203,67 +203,75 @@ def get_expectation_values(self) -> list[float]:
...
+            for res in self._intermediate_results:
                 if has_state_tomography(res):
                     results.append(res.state_tomography.state)
             return results
         raise ValueError("Can't find intermediate states because intermediate Results were not stored.")
 
-    @property
-    def intermediate_expectation_values(self) -> list[list[float]]:
-        """Expectation values for every time-step (intermediate + final).
+    def get_intermediate_expectation_values(self) -> list[list[float]]:
```
amirazzam2000, line 282: same here

amirazzam2000 (changes requested):
Great work!! Thanks Luke!

src/qilisdk/backends/qutip_backend.py
```diff
@@ -133,6 +136,16 @@ def _execute_digital_propagation(
...
+                    if not isinstance(later_gate, M):
+                        raise ValueError(
+                            "Mid-circuit measurements are not supported in the QutipBackend. All measurements must be at the end of the circuit."
+                        )
+
```
amirazzam2000: consider changing this block of code into: 

```python
gates = functional.circuit.gates

# Find the index of the first measurement gate
try:
    first_m = next(i for i, g in enumerate(gates) if isinstance(g, M))
except StopIteration:
    first_m = None  # No measurements at all

# If there's a measurement, check that nothing after it is a non-measurement
if first_m is not None and any(
    not isinstance(g, M) for g in gates[first_m + 1:]
):
    raise ValueError(
        "Mid-circuit measurements are not supported in the QutipBackend. "
        "All measurements must be at the end of the circuit."
    )
```

I am just repeating this comment because i think we need to be as efficient as possible with this because this might be something we have to do many times (especially with variational program), and python loops are notorious for being very inefficient and slow.

amirazzam2000 (approved):

PR #186, opened 2026-04-17
[SDK-256] Add QTensor Uniform Init Method

Lumorti:
Added a new QTensor constructor to allow users to easily construct a uniform superposition state of a certain number of qubits:
```python3
state = QTensor.uniform(2)
```


Lumorti:
Note, most of the files changes are from the clang-format, which wasn't run in the last few PRs I think.

The only files where something was actually added have qtensor in their name (i.e. qtensor.cpp, qtensor.py, test_qtensor.cpp,  test_qtensor.py).

Also, code-cov is lying, everything that this PR adds is covered.

amirazzam2000 (approved):
Thanks Luke!!

PR #187, opened 2026-04-17
[SDK-150] Docs Translations

Lumorti:
The QiliSDK docs have been translated into Catalan and Spanish (for now, more to come later).
You can change the language via the language switcher at the top of the page.


Lumorti:
Note for reviewers: each po files contains the extracted text from the English docs. The files contain string pairs like this:

```
#: ../../modules/analog/analog.rst:21
msgid "**Common constructors**"
msgstr "**Constructors comuns**" 
```

where the first line is the English and the second is the translated. You can build the docs by entering the docs folder and typing "make translations", which is probably how I would recommend reviewing this PR.

amirazzam2000 (approved):

PR #188, opened 2026-04-24
[SDK-260] Make Qilisim Faster

Lumorti:
QiliSim's default analog simulator - `integrator` - has been optimized and now offers around 4x performance for the same number of steps.

A new analog simulation method has also been added: `adaptive_integrator`, capable of speeding up when the problem is easy and slowing down when precision is needed. It does this using a Dormand-Prince-style RK45 method. For some cases this can offer a speedup, but the main advantage is that the user doesn't have to choose their `dt` carefully. Usage is as follows:

```python
backend = QiliSim(analog_simulation_method=AnalogMethod.adaptive_integrator(tol=0.01),)
```


src/qilisdk/backends/backend_config.py
```diff
@@ -89,19 +89,19 @@ class AnalogMethod(BaseSimulatorConfig):
 
     Args:
         evolution_method (str): Analog time-evolution method to use:
-            ``"direct"``, ``"arnoldi"``, ``"integrate"``, or
-            ``"integrate_matrix_free"``. Defaults to ``"integrate"``.
+            ``"direct"``, ``"arnoldi"``, ``"integrate_rk4"``, ``"integrate_rk45_matrix_free"``, or
```
amirazzam2000, line 92: i would use the names intergator and adaptive_integrator for consistency with the class methods and to avoid breaking changes for our pre-established interface.

src/qilisdk/backends/backend_config.py
```diff
@@ -125,27 +125,41 @@ def get_config(self) -> SolverConfigDict:
...
             "arnoldi_dim": self.arnoldi_dim,
             "num_arnoldi_substeps": self.num_arnoldi_substeps,
-            "num_integrate_substeps": self.num_integrate_substeps,
+            "adaptive_tol": self.adaptive_tol,
         }
 
         return d
 
     @classmethod
-    def integrator(cls, *, num_substeps: int = 2, matrix_free: bool = False) -> AnalogMethod:
```
amirazzam2000, line 134: why did we remove num_substeps? before increasing the number of steps actually saved some of my simulations.

src/qilisdk/backends/backend_config.py
Lumorti: These are just the internal names, the user never sees these, no? I think they're only used in the c++

src/qilisdk/backends/backend_config.py
Lumorti: Yes I was gonna comment about this, basicially I tested, and it's always better to just half the dt than doing two substeps each time.

It was mostly a relic brought over from the the methods that implement e^iH

But basically there is no need to do two steps at H(t), you might as well do a step at H(t) and then one at H(t+dt)

Lumorti: It's only useful for the direct/arnoldi methods since their costly part is calculating e^iH, so reusing that part for the substeps saves time and adds accuracy, but for the integrator we might as well just use more timesteps, it's the same number of evaluations but better coverage of the function

src/qilisdk/backends/backend_config.py
amirazzam2000: well a user can still construct it by doing `AnalogMethod(evolution_method="integrate_rk4", num_integrate_substeps=num_substeps)`

which would be inconsistent with the classmethod names

Lumorti: I was preparing in case later I add different integrators etc.

src/qilisdk/backends/backend_config.py
amirazzam2000: okay so for now the dt is the only way of controlling the number of integrator steps, is that right?

src/qilisdk/backends/backend_config.py
amirazzam2000: then maybe you would keep the integrator as it is but would add an extra field called integrator method. like we do with matrix free. For now i don't mind the adaptive integraor being a separate class method, but maybe we can do something similar to matrix free.

Lumorti: Eh, the settings are different, for the adaptive there is this "tol" parameter, and I thought the whole point of separating the classes was to make it so the user can't set unnecessary params. In fact I think the matrix free ones should probably also be different constructors, the only reason I didn't was because I hoped we could silently remove the non-matrix-free ones in a later patch

src/qilisdk/backends/backend_config.py
Lumorti: Yes, if you want to replicate the only behavior, just half dt, and that should give better performance than doing two substeps

amirazzam2000 (approved):
thank you luke!! 🙌

PR #189, opened 2026-04-24
feat(build): added cmake flags for py-spy/pyroscope profiling

JavierSab:


Lumorti:
Should I close this? Or did you still want to use it for something?

PR #191, opened 2026-04-28
[SDK-264] Add Schedule Eigenspectrum Plot

Lumorti:
You can now generate a plot of the eigenspectrum of a Schedule for small (i.e. less than 7 qubits) systems.
This can also be combined with simulation data to visualize how the state evolves.
Usage is as follows:

```python
from qilisdk.core import QTensor, Interpolation

nqubits = 5
T = 10.0
dt = 0.1
Hz = sum(Z(i)*Z((i+1) % nqubits) for i in range(nqubits))
schedule = Schedule(
    hamiltonians={"driver": Hx, "problem": Hz},
    coefficients={
        "driver": {(0.0, T): lambda t: 1 - t / T},
        "problem": {(0.0, T): lambda t: t / T},
    },
    dt=dt,
    interpolation=Interpolation.LINEAR,
)

backend = QiliSim(analog_simulation_method=AnalogMethod.integrator())
evolution = AnalogEvolution(schedule=schedule, initial_state=QTensor.uniform(nqubits), store_intermediate_results=True)
res = backend.execute(evolution, Readout().with_state_tomography())

schedule.draw_eigenvalues(levels=200, intermediate_states=res.get_intermediate_states(), show_overlaps=True)
```

which generates the following, where the percentages are the overlap with the evolved state versus each of the eigenstates:

fig


src/qilisdk/analog/schedule.py
```diff
@@ -420,6 +423,56 @@ def draw(self, style: ScheduleStyle | None = None, filepath: str | None = None)
...
+
+        Raises:
+            ValueError: If the number of qubits exceeds the supported limit for eigenvalue plotting.
+            ValueError: If show_overlaps is True but intermediate_states is not provided.
+
+        """
+        from qilisdk.utils.visualization.schedule_renderers import MatplotlibEigenvalueRenderer  # noqa: PLC0415
+
+        # Don't allow big systems
+        _MAX_QUBITS_FOR_EIGENVALUE_PLOTTING = 7
```
fedonman: Why do we restrict this? The plot might be useful to a user even if it is not very readable or takes time to draw. Ideally, I would remove it. If we do need to enforce a restriction, I would add it as a new field in the settings instead of hardcoding the value.

src/qilisdk/analog/schedule.py
```diff
@@ -420,6 +423,56 @@ def draw(self, style: ScheduleStyle | None = None, filepath: str | None = None)
...
+        if self.nqubits > _MAX_QUBITS_FOR_EIGENVALUE_PLOTTING:
+            raise ValueError(
+                f"Eigenvalue plotting is not supported for schedules with more than {_MAX_QUBITS_FOR_EIGENVALUE_PLOTTING} qubits due to computational complexity."
+            )
+
+        # If we try to show overlaps but haven't given intermediate states, raise an error
+        if show_overlaps and intermediate_states is None:
+            raise ValueError(
+                "Overlaps can't be shown without intermediate states. Please provide intermediate_states or set show_overlaps to False."
+            )
```
fedonman: Maybe instead of raising an error, log a warning and plot with show_overlaps=False? Technically the plot can be drawn, just not as expected. No hard opinion on this.

src/qilisdk/utils/visualization/schedule_renderers.py
```diff
@@ -173,3 +187,170 @@ def _make_axes(dpi: int, style: ScheduleStyle) -> plt.Axes:
...
+            full_eigenvalues.append([float(ev.real) for ev in vals[: self.levels]])
+            full_eigenstates.append(list(vecs[: self.levels]))
+
+            # Also plot the expectation value if we have intermediate states
+            if self.intermediate_states:
+                state = self.intermediate_states[i]
+                exp_val = state.expectation_value(as_qtensor)
+                actual_expectation_values.append(float(exp_val.real))
+
+        return full_eigenvalues, full_eigenstates, actual_expectation_values
```
fedonman: Is plotting the only place we need to calculate eigenvalues and eigenstates of a schedule? I feel like this is super useful to the user. Should we consider moving it to the `analog` or `utils` module?

fedonman (changes requested):
Awesome work! Left some comments for discussion.

src/qilisdk/utils/visualization/schedule_renderers.py
Lumorti: I'm not sure how generally useful it is, @amirazzam2000 is their any general use for this? It's quite an expensive operation

src/qilisdk/analog/schedule.py
Lumorti: Sure

src/qilisdk/analog/schedule.py
Lumorti: I can remove this, I only put it because a while back when I was considering adding Hamiltonian.eigenvalues() and Hamiltonian.ground_state(), Ameer's concern was that the user could do this on a large Hamiltonian without realizing and it would crash their computer

src/qilisdk/utils/visualization/schedule_renderers.py
Lumorti: Will move to schedule

src/qilisdk/analog/schedule.py
Lumorti: Will raise a warning instead

src/qilisdk/analog/schedule.py
```diff
@@ -530,6 +534,82 @@ def draw(self, style: ScheduleStyle | None = None, filepath: str | None = None)
         else:
             renderer.show()
 
+    def draw_eigenvalues(
+        self,
+        levels: int = 50,
+        style: ScheduleStyle | None = None,
+        filepath: str | None = None,
+        intermediate_states: list[QTensor] | None = None,
+        show_overlaps: bool = False,
```
fedonman, line 543: Minor: I would re-order the `style` and `filepath` to be the first and second arguments, and the rest follow afterwards, so it's consistent with the rest of the draw methods.

fedonman (approved):
LGTM. Great work 💯

src/qilisdk/analog/schedule.py
Lumorti: Will do!

PR #195, opened 2026-04-30
[SDK-141] constraint linearization

amirazzam2000:
`Model.to_qubo()` now automatically *linearizes* pseudo-Boolean objectives and constraints of degree greater than two. High-degree monomials are reduced pairwise to quadratic form by introducing fresh auxiliary binary variables `w = a * b` enforced via the Rosenberg penalty

```
P(a, b, w) = a * b - 2 * a * w - 2 * b * w + 3 * w
```

which is quadratic, non-negative, and zero iff `w = a * b`. Shared sub-products (e.g. `x*y*z` and `x*y*w` both reusing `x*y`) share a single auxiliary and a single penalty.

Two new kwargs control the behavior:

- `linearize` (default `True`): toggles the reduction. Set `linearize=False` to keep the previous strict behavior where exporting a model with terms of degree 3+ raises `ValueError`.
- `linearization_lagrange_multiplier` (default `100`): the Lagrange multiplier applied to each Rosenberg penalty constraint.

Usage:

```python
from qilisdk.core import BinaryVariable, EQ, Model, ObjectiveSense

x, y, z = BinaryVariable("x"), BinaryVariable("y"), BinaryVariable("z")

model = Model("cubic")
model.set_objective(x * y * z, sense=ObjectiveSense.MAXIMIZE)
model.add_constraint("forbid_triple", EQ(x * y * z, 0), lagrange_multiplier=10)

qubo = model.to_qubo(linearization_lagrange_multiplier=50)
ham = qubo.to_hamiltonian()
```


src/qilisdk/core/model.py
```diff
@@ -466,25 +469,212 @@ def to_qubo(
...
+        introducing auxiliary binary variables and corresponding Rosenberg penalty constraints.
+        See :meth:`QUBO.from_model` for details on the reduction scheme.
+
         Args:
-            lagrange_multiplier_dict (dict[str, float] | None, optional): A dictionary with lagrange multiplier values to scale the model's constraints. Defaults to None.
-            penalization (Literal[&quot;unbalanced&quot;, &quot;slack&quot;], optional): the penalization used to handel inequality constraints. Defaults to "slack".
-            parameters (list[float] | None, optional): the parameters used for the unbalanced penalization method. Defaults to None.
+            lagrange_multiplier_dict (dict[str, float] | None, optional): A dictionary with lagrange multiplier values
+                            to scale the model's constraints. Defaults to None.
+            penalization (Literal[&quot;unbalanced&quot;, &quot;slack&quot;], optional): the penalization used to handel
```
Lumorti:
```suggestion
            penalization (Literal[&quot;unbalanced&quot;, &quot;slack&quot;], optional): the penalization used to handle
```

src/qilisdk/core/model.py
```diff
@@ -466,25 +469,212 @@ def to_qubo(
...
+            self,
+            lagrange_multiplier_dict,
+            penalization,
+            parameters,
+            linearize=linearize,
+            linearization_lagrange_multiplier=linearization_lagrange_multiplier,
+        )
+
+
+class _Linearizer:
```
Lumorti, line 514: I don't think I've seen us do any _ClassName classes so far, like for instance all of the FunctionalResult And CompositeResult which I would say are more or less internal classes are still not with an underscore. But this isn't a blocker for me, just a comment

src/qilisdk/core/model.py
```diff
@@ -466,25 +469,212 @@ def to_qubo(
...
+           deterministic and independent of the monomial ordering in the source term.
+
+        Returns:
+            tuple[BaseVariable, BaseVariable]: the pair of factors to replace with an auxiliary.
+        """
+        for a, b in itertools.combinations(variables, 2):
+            a_sorted, b_sorted = sorted([a, b], key=lambda v: v.label)
+            if (a_sorted.label, b_sorted.label) in self._substitutions:
+                return a, b
+        return variables[0], variables[1]
```
Lumorti, line 677: I was gonna comment above about the way the pairs are chosen, but this is a nice consistent way to do it, and I like the caching, nice job!

src/qilisdk/core/model.py
```diff
@@ -536,38 +744,52 @@ def qubo_objective(self) -> Objective | None:
...
-        terms: list[tuple[Number, BaseVariable]] = []
 
-        if term.degree > 1:
-            raise ValueError(f'QUBO constraints only allow linear terms but received "{term}" of degree {term.degree}')
+        def to_real(num: Number) -> RealNumber:
+            if isinstance(num, RealNumber):
+                return num
+            if isinstance(num, complex) and abs(num.imag) < get_settings().atol:
+                return num.real
+            raise ValueError("Complex values encountered in the constraint.")
```
Lumorti, line 768: We do have a lot of mini functions like real_if_close and other type-checker functions, we should think about moving them all to utils or something (longer term I mean, not this PR)

src/qilisdk/core/model.py
```diff
@@ -767,18 +982,28 @@ def add_constraint(
...
             c = copy.copy(term)
 
-        if c.degree > 2:  # noqa: PLR2004
+        if linearize and self._linearizer is None:
+            self._linearizer = _Linearizer()
+
+        if self._linearizer is None and c.degree > 2:  # noqa: PLR2004
             raise ValueError(
-                f"QUBO models can not contain terms of order 2 or higher but received terms with degree {c.degree}."
+                f"QUBO constraints can not contain terms of order 2 or higher but received terms with degree {c.degree}."
```
Lumorti:
```suggestion
                f"QUBO constraints can not contain terms of order 2 or higher but received terms with degree {c.degree}. Set linearize=True to allow linearization."
```

src/qilisdk/core/model.py
```diff
@@ -799,17 +1024,31 @@ def add_constraint(
...
+        Raises:
+            ValueError: if the degree of the provided term is larger than 2.
+
         """
 
         self._check_variables(term)
 
+        if self._linearizer is None and term.degree > 2:  # noqa: PLR2004
+            raise ValueError(
+                f"QUBO objective can not contain terms of order higher than 2 but received terms with degree {term.degree}."
```
Lumorti:
```suggestion
                f"QUBO objective can not contain terms of order higher than 2 but received terms with degree {term.degree}. Set linearize=True to enable linearization."
```

tests/integration/core/__init__.py
Lumorti, line 1: We should add a check script to make sure all files have this, I'll add a linear task to do it

tests/integration/core/test_linearization.py
```diff
@@ -0,0 +1,147 @@
+# Copyright 2025 Qilimanjaro Quantum Tech
```
Lumorti: Actually, shouldn't this be 2026 since this a new file?
```suggestion
# Copyright 2026 Qilimanjaro Quantum Tech
```

tests/integration/core/test_model.py
Lumorti, line 1: Could maybe use a few more tests, I noticed codecov still had some missing lines, and I looked and they seem quite easily testable, but no worries if not

tests/unit_python/core/test_model.py
Lumorti, line 1: Why the linearization tests here and not in `test_linearization.py`?

Lumorti (approved):
Nice work, good implementation! I left a few comments, but most are just comments

src/qilisdk/core/model.py
amirazzam2000: I guess we can make it public, but i didn't think it will have any relevance outside of QUBO constraint linearization. The main task this serves is adding caching and avoiding having all this code inside the qubo and maybe in the future we will have different linearization algorithms so their codes will always be contained in classes.

src/qilisdk/core/model.py
amirazzam2000: good idea, maybe they are term utils

tests/integration/core/__init__.py
amirazzam2000: I think you already did this, no?

tests/unit_python/core/test_model.py
amirazzam2000: these are unit tests where the other was integration test. I will rename the integration test to test_model for consistency

tests/integration/core/__init__.py
amirazzam2000: luke : yes

Lumorti (approved):
💯️

PR #196, opened 2026-04-30
[Misc] Update Dependencies

fedonman:
Removed `pre-commit` since it is unused. Updated dependencies and fixed or ignored the new `ty` rules.


src/qilisdk/digital/gates.py
```diff
@@ -451,13 +451,7 @@ def is_modified_from(self, gate_type: type[TBasicGate]) -> bool:
 
 @yaml.register_class
 class Controlled(Modified[TBasicGate]):
-    def __init__(self, *control_qubits: int, basic_gate: Gate) -> None:
-
-        # Make sure it's an instance of the class and not the class itself
-        # i.e. Z(0).controlled(1) not Z.controlled(0, 1)
-        if not isinstance(basic_gate, Gate):
-            raise ValueError("basic_gate must be an instance of a Gate, not the class itself.")
```
Lumorti, line 459: I added this because at some point I had written Z.controlled(0, 1) (or maybe Claude wrote it) and the error message that gives was very hard to debug, since it doesn't error here, it only errors much later I think

fedonman: I guessed so, but this can happen with any method, it doesn't make sense to add checks everywhere.

Lumorti: Yeah fair enough

Lumorti (approved):
Nice fixes!

PR #197, opened 2026-05-06
[SDK-274] fixing set parameters in qr

amirazzam2000:
Fixed `ReservoirLayer.set_parameters` and `ReservoirLayer.set_parameter_bounds` so they dispatch through each sub-component (input encoding, evolution dynamics, output encoding) instead of relying on the inherited `Parameterizable` implementation. The previous behavior updated `Parameter` values directly and bypassed the encoding gates' setters, leaving their cached matrices stale after a parameter update on the layer. Added tests covering both the matrix-cache invalidation on encoding gates and the propagation of bounds to each child component.


Lumorti (approved):
Nice fix!

PR #198, opened 2026-05-06
Hotfix - fix docs translation build errors

Lumorti:
Fixed a few small issues with the new build process. Now it redirects root as expected and copies all the needed files.


docs/locale/es/LC_MESSAGES/modules/backends/backends_overview.po
Lumorti, line 1: (I noticed this last minute, that this was inconsistent between files)

.github/workflows/sphinx.yml
```diff
@@ -56,11 +56,16 @@ jobs:
             exit 1
           fi
           printf '<!doctype html><meta http-equiv="refresh" content="0; url=%s/">' "$latest_tag" > docs/_build/en/index.html
+          printf '<!doctype html><meta http-equiv="refresh" content="0; url=%s/">' "en/$latest_tag" > docs/_build/index.html
+          mkdir -p docs/_build/main
+          printf '<!doctype html><meta http-equiv="refresh" content="0; url=%s/">' "../en/main" > docs/_build/main/index.html
+          mkdir -p docs/_build/0.1.8
```
DZPM, line 62: `0.1.8` is hardcoded in 3 places. It'll get stale, it'll get more complicated the more versions we support.

Could be a list of versions in a config file (`VERSIONS`?), easier to maintain and iterate over the template

docs/locale/es/LC_MESSAGES/modules/backends/backends_overview.po
```diff
@@ -21,7 +21,7 @@ msgstr ""
 
 #: ../../modules/backends/backends_overview.rst:2
 msgid "Overview"
-msgstr "Descripción general"
+msgstr "Descripción General"
```
DZPM: Spanish doesn't capitalize all word. It's "Descripción general" rather than "Descripción General".

.github/workflows/sphinx.yml
Lumorti: This was just for this version so that anyone that has a bookmark to the old url syntax now redirects to the new link. Nobody will bookmark the /0.1.9 page because it doesn't exist, they will bookmark /en/0.1.9 so there won't be a problem in future, this is just here for backwards compatibility for this version change

docs/locale/es/LC_MESSAGES/modules/backends/backends_overview.po
Lumorti: Ah I didn't know! I'll change it here, then I'll get Claude to check the rest of the translations, but I'll do that in the translations-typos branch

DZPM (approved):
LGTM!

PR #199, opened 2026-05-06
[SDK-275] Fix Gate Matrix Regeneration

amirazzam2000:
Fixed `Circuit.set_parameters`, `Circuit.set_parameter_bounds`, and `Circuit.set_prefix` so they correctly handle a single `Parameter` shared across multiple gates. Previously `Circuit._parameters_link` stored a single `(label, gate)` tuple per parameter, so only the last-added gate received the update; earlier gates kept their cached matrices stale after the underlying `Parameter` was mutated. The link is now a list of every `(label, gate)` pair sharing a parameter, and all of them are updated. Added regression tests covering shared parameters across `set_parameters`, `set_parameter_bounds`, and the `set_prefix` + `set_parameters` flow.


Lumorti (approved):
Nice fix, good catch!

PR #202, opened 2026-05-11
[SDK-240] Fix Windows Build Issues

Lumorti:
Fixed build issues when trying to compile QiliSDK from source on Windows using VSCode.


.github/workflows/platforms.yml
```diff
@@ -0,0 +1,40 @@
...
+            **/uv.lock
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install qilisdk
+        run: uv -v sync --group dev --extra all-cu13 -Ccmake.build-type=Debug --reinstall
+
+      - name: Run Python tests
+        run: |
+          source .venv/bin/activate
+          pytest tests/unit_python/
```
fedonman: This runs all unit tests, I would keep this a smoke test that simply checks if we can import qilisdk aka the build was successful.
```suggestion
      - name: Smoke test import
        run: uv run python -c "import qilisdk; print('qilisdk imported successfully')"
```

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "rustworkx>=0.17.1",
     "openqasm3[parser]>=1.0.1",
     "psutil>=7.2.2",
+    "sphinx-tabs>=3.5.0",
```
fedonman: Why is this added in the core dependencies?

fedonman (commented):
Great 💯 It works for me locally. Made two comments.

pyproject.toml
Lumorti: Very true, it should just be in the docs group

fedonman (approved):
Awesome 💯 💯

PR #206, opened 2026-05-13
[SDK-280] Improve Experiment Plots

Lumorti:
Curve fitting has been added to the T1 experiment plot, to better indicate the T1 time.

Two Tones with simulated data:
image

T1 with simulated data:
image


src/qilisdk/experiments/experiment_result.py
```diff
@@ -225,6 +226,55 @@ class T1ExperimentResult(ExperimentResult):
...
+                a (float): Amplitude of the decay.
+                t1 (float): T1 relaxation time (in microseconds).
+                b (float): Baseline offset.
+
+            Returns:
+                np.ndarray: The modeled decay curve values at time t.
+            """
+            return a * np.exp(-t / t1) + b
+
+        initial_guess = [1, 10, -8]
```
NataliaSirera: make an argument with default None and then is calculated based on a formula?

src/qilisdk/experiments/experiment_result.py
```diff
@@ -225,6 +226,57 @@ class T1ExperimentResult(ExperimentResult):
     plot_title: ClassVar[str] = "T1"
     """Default title for T1 experiment plots."""
 
+    def plot(self, save_to: str | None = None, initial_guess: list[float] | None = None) -> None:
```
fedonman: Idea: Instead of overriding the `plot` method, which already works well for both 1D and 2D cases, we could define a private method responsible for calculating the fitting data and the corresponding plot calls. Then, each specific experiment would only need to override that method. What do you think?

fedonman (commented):
Great! Left a comment to discuss.

src/qilisdk/experiments/experiment_result.py
Lumorti: I'm not against refactoring it a bit, but what did you have in mind? That the shared plot function calls some "add_fit" which for now is just an empty function for all cases except for the T1?

fedonman: Yes, exactly. The idea is that the plotting of data is universal for all experiments, and the only thing that changes is the calculation and plotting of fitting.

fedonman: I am saying because otherwise we will need to override and essentially copy/paste the plotting for every experiment.

Lumorti: Makes sense! Will do

changes/206.misc.md
```diff
@@ -0,0 +1 @@
+The plots for T1 and Two-Tone experiments have been improved, such as by adding fitting to the T1.
```
fedonman: Two-Tone experiment hasn't been edited.

fedonman (approved):
LGTM 💯 Edit the changelog entry before merging.

PR #207, opened 2026-05-13
[SDK-278] Adding lp parser

amirazzam2000:
Added LP-format (CPLEX flavor) support to `qilisdk.utils.lp_parser`. A `Model` can now be read from or written to LP text and files, round-trippable through the parser and serializer:

* `from_lp(lp_str)` / `from_lp_file(path)` parse LP-format text into a `Model`.
* `to_lp(model)` / `to_lp_file(model, path)` serialize a `Model` back to LP-format text.

Supported features:

* Linear and quadratic objectives (`Minimize` / `Maximize`, including bilinear and squared terms).
* Linear constraints with senses `<`, `<=`, `=`, `>=`, `>` (and the LP synonyms `=<`, `=>`).
* `Bounds` section: two-sided ranges (`lo <= x <= hi`), one-sided bounds (`x >= lo`, `x <= hi`), fixed values (`x = c`), `free` declarations, and `±infinity` keywords.
* `General` (integer) and `Binary` variable declarations.
* Coefficients and bounds in scientific notation (`1.5e2`, `1e-3`, `2E5`).
* Backslash (`\`) line comments.

Example — build a model in code, serialize it to LP, and read it back:

```python
from qilisdk.core.model import Model, ObjectiveSense
from qilisdk.core.variables import LEQ, Domain, Variable
from qilisdk.utils.lp_parser import from_lp, to_lp, to_lp_file

x = Variable("x", Domain.REAL, bounds=(0, 10))
y = Variable("y", Domain.INTEGER, bounds=(0, 5))

model = Model("example")
model.set_objective(2 * x + 3 * y, label="profit", sense=ObjectiveSense.MAXIMIZE)
model.add_constraint(label="capacity", term=LEQ(x + y, 8))

lp_text = to_lp(model)
print(lp_text)
# Maximize
#  profit: 2 x + 3 y
# Subject To
#  capacity: x + y <= 8
# Bounds
#  0 <= x <= 10
#  0 <= y <= 5
# General
#  y
# End

to_lp_file(model, "example.lp")
round_tripped = from_lp(lp_text)
```

Reading an existing LP file is a one-liner:

```python
from qilisdk.utils.lp_parser import from_lp_file

model = from_lp_file("knapsack.lp")
print(model)
```


changes/207.feature.md
Lumorti, line 1: Maybe add a small example here?

```
from qilisdk.utils.lp_parser import from_lp_file

model = from_lp_file("knapsack.lp")
print(model)
```

(I would have added it as a GitHub suggestion, but you can't include code blocks within a suggested change because the three ticks escapes the suggestion)

docs/locale/ca/LC_MESSAGES/modules/core/core_lp_parser.po
Lumorti, line 1: Thanks for adding translations!! No worries if you ever forget, I'll do a PR occasionally to add any missing ones

docs/modules/core/core_lp_parser.rst
```diff
@@ -0,0 +1,142 @@
+LP File Parser and Serializer
```
Lumorti: Maybe the title could be Linear Program Parser, just in case someone is going through the drop down list and doesn't know what LP is
```suggestion
Linear Program Parser
```

docs/modules/core/core_lp_parser.rst
```diff
@@ -0,0 +1,142 @@
...
+    qubo = model.to_qubo()
+
+Round-tripping a model back to LP text:
+
+.. code-block:: python
+
+    from qilisdk.utils.lp_parser import to_lp, to_lp_file
+
+    print(to_lp(model))         # serialize to a string
+    to_lp_file(model, "out.lp")  # or write directly to disk
```
Lumorti: Nit, only because it was so close
```suggestion
    print(to_lp(model))          # serialize to a string
    to_lp_file(model, "out.lp")  # or write directly to disk
```

docs/modules/core/core_lp_parser.rst
```diff
@@ -0,0 +1,142 @@
...
+Serialization conventions
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:func:`~qilisdk.utils.lp_parser.to_lp` produces grammar-compliant output that
+:func:`~qilisdk.utils.lp_parser.from_lp` can read back. Notable conventions:
+
+- Quadratic and bilinear monomials are expanded into explicit products
+  (``x * x``, ``x * y``) since the LP grammar accepted by the parser has no
+  ``^`` operator.
+- A unit coefficient is elided (``x`` rather than ``1 x``); a ``-1`` coefficient
```
Lumorti: I didn't know the word "elide" existed, the more you know. Could change this to "omitted" if we want to make it a bit more common-English. But if you like it, keep it


```suggestion
- A unit coefficient is omitted (``x`` rather than ``1 x``); a ``-1`` coefficient
```

docs/modules/core/core_lp_parser.rst
```diff
@@ -0,0 +1,142 @@
...
+  of the line.
+
+Serialization conventions
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:func:`~qilisdk.utils.lp_parser.to_lp` produces grammar-compliant output that
+:func:`~qilisdk.utils.lp_parser.from_lp` can read back. Notable conventions:
+
+- Quadratic and bilinear monomials are expanded into explicit products
+  (``x * x``, ``x * y``) since the LP grammar accepted by the parser has no
```
Lumorti, line 90: Does the LP grammar allow non-linear terms? Or it supports them but would throw an error? Or the more general format supports them, but then it's not an LP it's just a P?

docs/modules/core/core_lp_parser.rst
```diff
@@ -0,0 +1,142 @@
...
+     y
+    Binary
+     b1
+     b2
+    End
+
+Parsing it returns a :class:`~qilisdk.core.model.Model` with:
+
+- five real-valued variables (``x1``, ``x2``, ``x3``, ``x4`` plus auto-typed
+  ones, plus integer ``y``),
```
Lumorti, line 131: What does "auto-typed ones" mean here?

src/qilisdk/utils/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+    Variable,
+    _assert_real,
+    _float_if_real,
+)
+
+__all__ = ["from_lp", "from_lp_file", "to_lp", "to_lp_file"]
+
+# === Constants ====================================================================
+
+_INF = 1e30
```
Lumorti: Isn't there a Python INF already? Is this 10^30 an arbitrary bound, or is it CPLEX standard?

src/qilisdk/utils/lp_parser/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+)
+
+__all__ = ["from_lp", "from_lp_file", "to_lp", "to_lp_file"]
+
+# === Constants ====================================================================
+
+_INF = 1e30
+# A monomial token is a Group of [coef_float, var_name_str]; this length lets us
+# distinguish it from a parenthesized sub-expression at parse-walk time.
+_MONOMIAL_LEN = 2
```
Lumorti, line 73: Is this because Sonar screamed at you about magic numbers? haha

src/qilisdk/utils/lp_parser/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+    valid_name = ~py_keyword + name
+
+    colon = Suppress(one_of(": ::"))
+    plus_minus = one_of("+ -")
+    inf = one_of("inf infinity", caseless=True)
+    number = Word(nums + ".")
+    sense = one_of("< <= =< = > >= =>")
+
+    # Section tags.
+    obj_tag_max = one_of("max maximum maximize", caseless=True)
```
Lumorti, line 260: I like this syntax, this is a nice way to do parsing!

src/qilisdk/utils/lp_parser/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+    obj_tag_min = one_of("min minimum minimize", caseless=True)
+    obj_tag = (obj_tag_max | obj_tag_min).set_results_name("objSense")
+    constraints_tag = one_of(["subj to", "subject to", "s.t.", "st"], caseless=True)
+    bounds_tag = one_of("bound bounds", caseless=True)
+    bin_tag = one_of("bin binaries binary", caseless=True)
+    gen_tag = one_of("gen general", caseless=True)
+    end_tag = CaselessLiteral("end")
+
+    # Coefficients (sign + optional number) collapse to a Python float.
+    first_var_coef = Optional(plus_minus, "+") + Optional(number, "1")
```
Lumorti, line 270: Does this mean that the default behavior is to assume float? Since both are optional, so if there was an empty string etc. it would try to get parsed as a float?

src/qilisdk/utils/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+    constraint = Group(labelled + var_expr + sense + rhs)
+    constraints = ZeroOrMore(constraint).set_results_name("constraints")
+
+    signed_inf = (plus_minus + inf).set_parse_action(lambda toks: (1 if toks[0] == "+" else -1) * _INF)
+    signed_number = (Optional(plus_minus, "+") + number).set_parse_action(lambda toks: float("".join(toks)))
+    number_or_inf = signed_number | signed_inf
+    lineq = number_or_inf + sense
+    rineq = sense + number_or_inf
+
+    sensestmt = Group(
```
Lumorti: What's this variable name supposed to mean? sense_statement?

src/qilisdk/utils/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+        else:
+            operand = _extract_arith(e, variable_dict)
+        current = _merge(current, op, operand)
+        op = ""
+
+    return accum + current
+
+
+def _extract_constraint(expr: ParseResults | list, variable_dict: dict[str, BaseVariable]) -> ComparisonTerm:
+    """Split a constraint's tokens at the sense operator and build the comparison.
```
Lumorti: Sense operator?

src/qilisdk/utils/lp_parser/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+        return None
+    if lo_inf:
+        return f"-infinity <= {var.label} <= {_format_number(float(hi))}"
+    if hi_inf:
+        return f"{var.label} >= {_format_number(float(lo))}"
+    return f"{_format_number(float(lo))} <= {var.label} <= {_format_number(float(hi))}"
+
+
+# === Public API (serialization) ===================================================
+
```
Lumorti, line 636: could we move the from_lp and from_lp_file functions here? Just to have the main functions all be together

tests/unit_python/utils/test_lp_parser.py
```diff
@@ -0,0 +1,559 @@
...
+    assert model.objective.label == "my_obj"
+
+
+@pytest.mark.parametrize("keyword", ["Minimize", "Minimum", "Min", "MINIMIZE"])
+def test_minimize_keyword_is_caseless(keyword):
+    """`min`, `minimum`, `minimize` (any case) should all be treated as MINIMIZE."""
+    model = from_lp(f"{keyword}\n x\nSubject To\n x >= 0\nEnd\n")
+    assert model.objective.sense is ObjectiveSense.MINIMIZE
+
+
```
Lumorti, line 64:
```suggestion

@pytest.mark.parametrize("keyword", ["Maximize", "Maximum", "Max", "MAXIMIZE"])
def test_maximize_keyword_is_caseless(keyword):
    """`max`, `maximum`, `maximize` (any case) should all be treated as MAXIMIZE."""
    model = from_lp(f"{keyword}\n x\nSubject To\n x >= 0\nEnd\n")
    assert model.objective.sense is ObjectiveSense.MAXIMIZE
    
```

src/qilisdk/utils/lp_parser.py
fedonman: `_INF = 1e30` duplicates `Domain.REAL.max()`. Pick one source of truth (probably the domain) to avoid drift if either ever changes.

src/qilisdk/core/model.py
```diff
@@ -357,12 +357,12 @@ def _generate_encoding_constraints(
...
             ub_encoding_name = f"{var}_upper_bound_constraint"
             lb_encoding_name = f"{var}_lower_bound_constraint"
-            if ub_encoding_name not in self._encoding_constraints:
+            if ub_encoding_name not in self._encoding_constraints and var.upper_bound < var.domain.max():
                 self._encoding_constraints[ub_encoding_name] = Constraint(
                     label=ub_encoding_name, term=LEQ(var, var.upper_bound)
                 )
                 self._lagrange_multipliers[ub_encoding_name] = lagrange_multiplier
-            if lb_encoding_name not in self._encoding_constraints:
+            if lb_encoding_name not in self._encoding_constraints and var.lower_bound > var.domain.min():
```
fedonman, line 365: Good fix! Add a small test that constructs `Variable("x", REAL, bounds=(0, None))`, calls `_generate_encoding_constraints()`, and asserts `encoding_constraints == []`, so the new behavior doesn't quietly regress.

Lumorti (approved):
Nice PR! Left some suggestions for tweaks, but overall it's seems fine to me.

uv.lock
```diff
@@ -641,22 +641,22 @@ name = "cuda-quantum-cu12"
...
+    { name = "scipy", version = "1.17.1", source = { registry = "https://pypi.org/simple" }, marker = "(sys_platform == 'darwin' and extra == 'extra-7-qilisdk-all') or (sys_platform == 'darwin' and extra == 'extra-7-qilisdk-all-cu12') or (sys_platform == 'darwin' and extra == 'extra-7-qilisdk-cuda') or (sys_platform == 'darwin' and extra == 'extra-7-qilisdk-cuda12') or (extra == 'extra-7-qilisdk-all' and extra == 'extra-7-qilisdk-all-cu13') or (extra == 'extra-7-qilisdk-all-cu12' and extra == 'extra-7-qilisdk-all-cu13') or (extra == 'extra-7-qilisdk-cuda' and extra == 'extra-7-qilisdk-cuda13') or (extra == 'extra-7-qilisdk-cuda12' and extra == 'extra-7-qilisdk-cuda13')" },
```
fedonman, line 662: Why these changes in CUDA dependencies?

docs/modules/core/core_lp_parser.rst
amirazzam2000: yes the LP file format was extended by IBM to allow for non-linear programming.

docs/modules/core/core_lp_parser.rst
amirazzam2000: ones that are not specified in a specific type in the lp file like under Generic or Boolean

changes/207.feature.md
```diff
@@ -0,0 +1 @@
+Added an LP-format serializer (`to_lp` and `to_lp_file` in `qilisdk.utils.lp_parser`) that writes a `Model` back to CPLEX-flavor LP text, round-trippable through the existing `from_lp` parser. Also skips redundant variable-bound encoding constraints when a variable's bound matches its domain's natural extreme.
```
fedonman: Add more information, how to use it etc. Also, include a code example.

src/qilisdk/utils/lp_parser/lp_parser.py
amirazzam2000: yes hahahaha

changes/207.feature.md
fedonman: Second sentence is a bug fix, you could move it to a seperate `207.bugfix.md` file

src/qilisdk/utils/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+        "binary",
+        "gen",
+        "general",
+        "free",
+        "end",
+    ]
+    py_keyword = MatchFirst(map(CaselessKeyword, keywords))
+    valid_name = ~py_keyword + name
+
+    colon = Suppress(one_of(": ::"))
```
fedonman: Is `::` an LP syntax? I think this should be `Suppress(":")`

src/qilisdk/utils/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+    number_or_inf = signed_number | signed_inf
+    lineq = number_or_inf + sense
+    rineq = sense + number_or_inf
+
+    sensestmt = Group(
+        Optional(lineq).set_results_name("leftbound")
+        + valid_name.copy().set_results_name("name")
+        + Optional(rineq).set_results_name("rightbound")
+    )
+    free_var = Group(valid_name.copy().set_results_name("name") + Literal("free"))
```
fedonman: `Literal("free")` is case-sensitive. I think we should use `CaselessKeyword`. `Free` or `FREE` should both be accepted.

src/qilisdk/utils/lp_parser.py
amirazzam2000: well we always hardcode this value for the max real number. If you check the variables.py lines 208 and 219 you will see the same value hardcoded but there was no sonar cube back then so we didn't need to create a global for it hahahahah

src/qilisdk/utils/lp_parser/lp_parser.py
amirazzam2000: I am not sure i understand the question. but yes you want to collapse the parsed information into a float, because float is the most generic real number type. and we don't deal with complex numbers in lp files.

src/qilisdk/utils/lp_parser.py
amirazzam2000: it refers to a comparison statement. I will change the name

src/qilisdk/utils/lp_parser.py
amirazzam2000: this whole thing is because we have something called comparison sense, which specifies comparison operator. i will change it 👍

src/qilisdk/utils/lp_parser.py
```diff
@@ -0,0 +1,715 @@
...
+        "free",
+        "end",
+    ]
+    py_keyword = MatchFirst(map(CaselessKeyword, keywords))
+    valid_name = ~py_keyword + name
+
+    colon = Suppress(one_of(": ::"))
+    plus_minus = one_of("+ -")
+    inf = one_of("inf infinity", caseless=True)
+    number = Word(nums + ".")
```
fedonman: What about scientific notation? I think these won't get parsed: 1.5e2 / 1e-3 / 2E5. We should add tests for these cases.

src/qilisdk/utils/lp_parser/lp_parser.py
```diff
@@ -0,0 +1,712 @@
...
+    # direction" — the Variable then falls back to its domain's natural extreme.
+    bounds_map: dict[str, tuple[float | None, float | None]] = {}
+    for entry in parsed.bounds:
+        label = str(entry.name[0])
+        if len(entry) == _MONOMIAL_LEN and entry[1] == "free":
+            bounds_map[label] = (None, None)
+        else:
+            lo = float(entry.leftbound[0]) if entry.leftbound else 0.0
+            hi = float(entry.rightbound[1]) if entry.rightbound else _INF
+            bounds_map[label] = (_finite_or_none(lo), _finite_or_none(hi))
```
fedonman, line 341: What about `x >= 5` or `x = 5`? I think they currently produce (0, 5) which is wrong. We should add test cases.

src/qilisdk/utils/lp_parser/lp_parser.py
```diff
@@ -0,0 +1,712 @@
...
+    """
+    items = list(expr)
+    for i, e in enumerate(items):
+        if isinstance(e, str) and e in _COMPARISON_FACTORY:
+            lhs = _extract_arith(items[:i], variable_dict)
+            rhs = _extract_arith(items[i + 1 :], variable_dict)
+            return _COMPARISON_FACTORY[e](lhs, rhs)
+    raise ValueError("constraint is missing a comparison operator")
+
+
```
fedonman, line 473: Should label be empty? We could consider an optional `label: str = ""` parameter on `from_lp`, and have `from_lp_file` default it to `Path(filename).stem`. Symmetrically, `to_lp` currently drops `model.label` but we could write it as `\ Model: <label>\n` so the round-trip is lossless.

tests/unit_python/utils/test_lp_parser.py
```diff
@@ -0,0 +1,566 @@
...
+    reparsed = from_lp(text)
+    assert isinstance(_vars_by_label(reparsed)["b"], BinaryVariable)
+
+
+def test_one_sided_lower_bound_renders_compact_form():
+    x = Variable("x", Domain.REAL, bounds=(2, None))
+    model = Model("m")
+    model.set_objective(Term([x], Operation.ADD))
+    model.add_constraint("c", LEQ(x, 10))
+    assert "x >= 2" in to_lp(model)
```
fedonman, line 611: We should also test that round-trip `from_lp(to_lp(model))` recovers the original bounds. Same comment holds for other similar tests.

fedonman (changes requested):

fedonman (changes requested):
Great work!! Left some comments

src/qilisdk/utils/lp_parser/lp_parser.py
amirazzam2000: the thing is that i am not sure there are such thing as model label in lp file format, you might add it in a comment but in general it doesn't have a label like we do.

docs/modules/core/core_lp_parser.rst
```diff
@@ -1,5 +1,5 @@
-LP File Parser and Serializer
------------------------------
+Linear Programming File Parser and Serializer
```
Lumorti, line 1: This is quite a long title, which might have weird wrapping in the docs drop down, have a look

Lumorti (approved):
Nice fixes and improvements!

src/qilisdk/utils/lp_parser/lp_parser.py
fedonman: @amirazzam2000 If it doesn't support it, I suggest we add it as a comment, so the round-trip from qilisdk model doesn't lose any information.

uv.lock
fedonman: @amirazzam2000 Check this.

src/qilisdk/utils/lp_parser/lp_parser.py
amirazzam2000: done!

uv.lock
amirazzam2000: I deleted uv lock and recreated it, because i couldn't figure out why this update happened. I did add a package to pyproject.toml but that package was already installed through matplotlib so it shouldn't have added anything new. I just explicitly listed it so that we make sure it is not removed in the future.

docs/modules/core/core_lp_parser.rst
amirazzam2000: image
I think it looks fine, The reason i changed it is because people might not be aware of what LP is?? I can change it back if you think it's better to have it as LP. 

Let me know!

fedonman (approved):
Great work 💯

PR #210, opened 2026-05-19
Hotfix - Fix Windows Build Workflow

Lumorti:
The "compile with debug" option was causing syntax issues in the Windows build process, it has been removed. The platform workflow has been simplified (no longer installs all optional dependencies etc.), it's now just does `uv sync` as a user would do.

I also added a dispatch trigger, so we can manually run this workflow from a specific branch if we want, something that has been quite useful for the publish docs workflow.

### DIFF
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 jobs:
   build-wheels:
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -32,7 +32,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install qilisdk
-        run: uv -v sync --group dev --extra all-cu13 -Ccmake.build-type=Debug --reinstall
+        run: uv -v sync
 
       - name: Smoke test import
         run: uv run python -c "import qilisdk; print('qilisdk imported successfully')"
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -33,7 +33,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install qilisdk
-        run: uv -v sync
+        run: uv sync --reinstall
 
       - name: Smoke test import
         run: uv run python -c "import qilisdk; print('qilisdk imported successfully')"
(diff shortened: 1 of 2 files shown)


.github/workflows/platforms.yml
```diff
@@ -32,7 +33,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install qilisdk
-        run: uv -v sync --group dev --extra all-cu13 -Ccmake.build-type=Debug --reinstall
```
Lumorti, line 35: Regarding why this errors on Windows: it's just that Windows has different syntax for doing build-type=Debug, but since we don't actually use the debug features here (e.g. profiling) there is no need to do a separate Windows workflow.

.github/workflows/platforms.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
```
Lumorti, line 7: This is the same as in the publish docs workflow, it allows us to manually dispatch the workflow if we want

.github/workflows/platforms.yml
DZPM, line 35: Why removing the `--reinstall`? Looks out of scope for this ticket.

Removing the reinstall It could mask stale files/caches.

.github/workflows/platforms.yml
```diff
@@ -32,7 +33,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install qilisdk
-        run: uv -v sync --group dev --extra all-cu13 -Ccmake.build-type=Debug --reinstall
+        run: uv -v sync
```
DZPM: Why the `-v`? The normal verbosity should be more than enough

.github/workflows/platforms.yml
fedonman: I don't think we need to run this manually, same as tests and linting.

.github/workflows/platforms.yml
Lumorti: I believe that whenever one of the these workflows starts, it makes a new instance, --reinstall I mostly just use it locally when I want to recompile the C++, if it's the first time that someone does uv sync it doesn't really do anything I don't think. But I'll put it back for this PR

.github/workflows/platforms.yml
Lumorti: Completely fair, I'll remove it

.github/workflows/platforms.yml
Lumorti: It won't be often, but in this case I wanted to manually run it for a different branch to test something. Like say I make a hotfix branch for some mac build issues, being able to manually run this workflow for that branch would be quite useful, rather than waiting until I push to main

Lumorti: We could put is as a mandatory thing for all PRs like tests and linting, but since I think it takes quite a while to run maybe it's not so crucial

fedonman (approved):
💯 💯

PR #212, opened 2026-05-20
[SDK-285] Added Missing Translations, Fixed OpenQASM3 Bug

Lumorti:
Added documentation for the OpenQASM 2.0 and 3.0 parsers.

Added missing translations for the QIR docs.

Improved the specificity of the platform build checker.

Fixed a bug when doing an OpenQASM3 round-trip with CNOT and measurement gates.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,15 +19,6 @@ scripts/*.log
 lib*.a
 qili*.lib
 
-# these are generated if you check the docs code
-circuit_custom_font.png
-my_circuit.svg
-test.png
-bell.bc
-bell.ll
-*.qasm
-*.lp
-
 # Profiling stuff
 *.prof
 callgrind*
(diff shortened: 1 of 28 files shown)


.gitignore
```diff
@@ -23,6 +23,10 @@ qili*.lib
 circuit_custom_font.png
 my_circuit.svg
 test.png
+bell.bc
```
amirazzam2000: what is the bell.bc? it seems very specific, do you want to ignore all .bc files?

amirazzam2000 (approved):
Thanks Luke!! 💯 
I left one question tho about the gitignore

.gitignore
Lumorti: It's a file that's created if you run the docs.sh file, I think it must be part of the qir docs

That section of the gitignore is specific to the files that are created when running docs.sh

fedonman (approved):
💯 💯

.gitignore
Lumorti: Fixed

PR #215, opened 2026-05-21
HOTFIX QIR docs

amirazzam2000:


amirazzam2000:
I don't think this is needed

PR #217, opened 2026-05-22
[SDK-288] Improvements for CUDA Benchmarking

Lumorti:
New CudaSamplingMethods - CPU and STATE_VECTOR_MGPU, to tell the CudaBackend to use CPU or multiple GPUs, respectively. Multiple GPU statevector simulation requires a very specific setup, check the latest NVIDIA docs to see what versions of the various libraries they recommend. 

The performance of post-processing CUDA samples has also been improved. When calculating an observable cost function from samples, previously .get_probabilities() was used which if the results contained a state (as per the output of CUDA dynamics), would be 2^n values, rather than just the number of samples.


scripts/missing_translations.sh
Lumorti, line 1: Turns out this was missing a few of the missing/fuzzy translations! Fixed and added the missing translations

src/qilisdk/backends/cuda_backend.py
```diff
@@ -472,23 +473,43 @@ def _apply_digital_simulation_method(self) -> None:
...
+        For the CPU method, it sets the target to use CPU-based simulation.
+
+        Raises:
+            ValueError: If an unsupported sampling method is configured.
         """
         logger.info("Applying sampling simulation method {}", self.sampling_method.value)
-        if self.sampling_method == CudaSamplingMethod.STATE_VECTOR:
-            if cudaq.num_available_gpus() == 0:
+        if self.sampling_method in {CudaSamplingMethod.STATE_VECTOR, CudaSamplingMethod.STATE_VECTOR_MGPU}:
+            float_precision = "fp32" if get_settings().complex_precision == Precision.COMPLEX_64 else "fp64"
```
amirazzam2000: is this right or should it be flipped

amirazzam2000 (approved):
Thanks Luke!

src/qilisdk/backends/cuda_backend.py
Lumorti: Ah oops, good catch!!

PR #220, opened 2026-05-28
[SDK-269] Implement Large Scale Annealing

Lumorti:
Added a new analog simulation method to QiliSim - variational annealing, based on [this paper](https://arxiv.org/pdf/2403.05147) . This method supports any annealing schedule that transitions from an all X start to an Ising model, and  can offer big performance gains for larger numbers of qubits.

For a fully connected uniform Ising model of 60 qubits, it can give a solution within 99% of the optimum in around 5 seconds:

```python
from time import time

# Parameters to try tweaking
nqubits = 60
nsteps = 500
T = 300.0
order = 1
shots = 1000
warmups = 100

# Create the initial and final Hamiltonians
Hz = 0
for i in range(nqubits):
    for j in range(i + 1, nqubits):
        Hz += Z(i) * Z(j)

# Set up the evolution
dt = T / nsteps
schedule = Schedule.linear(Hx, Hz, T, dt)
analog = AnalogEvolution(schedule=schedule, initial_state=InitialState.UNIFORM)
readout = Readout().with_expectation([Hz])

# Initialize the backend with this new method
backend = QiliSim(analog_simulation_method=AnalogMethod.variational_annealing(order=order, shots=shots, warmups=warmups))

# Run it
print("Running evolution for nqubits =", nqubits, "and nsteps =", nsteps)
t0 = time()
results = backend.execute(analog, readout)
t1 = time()
print(f"Execution time: {t1 - t0} seconds")
print(results) # Optimum is -(nqubits / 2) if nqubits is even
```

The speed and accuracy are controlled by a number of parameters:
 - the `order` of the ansatz, 1 for quick fast results like the above case, 2 and above for a more expressive (but slower) ansatz
 - the number of monte carlo `shots`
 - the number of monte carlo `warmups`

The usual things apply as with the other analog simulation backends, a longer time `T` and more a smaller `dt` mean a better evolution.


scripts/install_dev.sh
Lumorti, line 1: removed for now because it never worked well

scripts/cov_python.sh
Lumorti, line 1: removed because we should use the unified one, I never used this or the pure cpp one

src/qilisdk/core/qtensor.py
```diff
@@ -26,13 +27,43 @@
...
 
 
+# Enum containing state presets like +, 0 etc.
+class InitialState(Enum):
+    """
+    Enumeration of symbolic initial states that can be used to specify the initial state of an evolution without needing to construct the corresponding QTensor manually.
+    """
+
+    UNIFORM = "uniform"
+    ZERO = "zero"
```
amirazzam2000, line 45: should we add ONE as well?

src/qilisdk/functionals/analog_evolution.py
```diff
@@ -47,28 +47,43 @@ class AnalogEvolution(PrimitiveFunctional):
...
+        """
+        The initial state of the simulation, if provided.
+
+        Returns:
+            QTensor | InitialState: The initial state.
+
+        Raises:
+            ValueError: If no initial state was provided for this functional.
+        """
+        if self._initial_state is None:
```
amirazzam2000: this should never happen no? initial state is not an optional parameter.

src/qilisdk/readout/readout.py
```diff
@@ -154,10 +151,14 @@ def observables(self) -> list[Hamiltonian | QTensor]:
...
     def qtensor_observables(self) -> list[QTensor]:
+        """Observables as :class:`~qilisdk.core.QTensor` instances.
+
+        Returns the expanded versions if :meth:`expand_observables` has been called;
+        otherwise converts each observable using its natural qubit count.
         """
-        The observables converted to :class:`~qilisdk.core.QTensor` form, populated automatically; not intended to be set manually.
-        """
-        return self._qtensor_observables
+        if hasattr(self, "_qtensor_observables"):
```
amirazzam2000: this will never be true if we don't populate this field. so maybe do something like: 
```python
if not hasattr(self, "_qtensor_observables"):    
     self._qtensor_observables = [o if isinstance(o, QTensor) else o.to_qtensor() for o in self._observables]
return self._qtensor_observables


```

amirazzam2000 (changes requested):
This is super cool! I like the addition of the symbolic expectation values as well! 
Great work 👍

src/qilisdk/backends/backend_config.py
```diff
@@ -118,6 +122,21 @@ class AnalogMethod(BaseSimulatorConfig):
...
+    warmups: int = Field(
+        default=100,
+        ge=0,
+        description="Number of warmup iterations to perform before collecting samples for the variational optimization when `evolution_method='variational_exponential'`.",
+    )
+    order: int = Field(
+        default=2,
+        gt=0,
+        description="Order of the polynomial expansion used in the variational ansatz when `evolution_method='variational_exponential'`.",
+    )
```
fedonman: shouldn't they get prefixed since they belong to specific evolution method? Similar to `arnoldi_*`.

fedonman: Maybe we should re-evaluate if we need config classes for each evolution method. (not in this PR)

src/qilisdk/core/qtensor.py
```diff
@@ -26,13 +27,43 @@
 from qilisdk.yaml import yaml
 
 if TYPE_CHECKING:
+    from qilisdk.analog import Hamiltonian
```
fedonman, line 30: Should the `core` module depend on `analog`? I have the feeling this goes against its design philosophy.

src/qilisdk/readout/readout.py
fedonman: 1. Why did we remove the field initialization from init?
2. If we want caching why don't we make it a cached property?

fedonman (changes requested):
Amazing addition! Left minor comments on the python side.

src/qilisdk/core/qtensor.py
amirazzam2000: There are multiple instances where core depends on other modules however it is always in a type_checking block. This is needed for when we return a Hamiltonian (QUBO.to_hamiltonian() method as an example). 

So I don't see this specific instance of dependency as a problem (given that it's only for type checking). I mean, it won't cause any circular dependency issues.

src/qilisdk/backends/backend_config.py
Lumorti: good idea

src/qilisdk/functionals/analog_evolution.py
Lumorti: good catch, this was a hold-over from the early version in which I allowed initial_state = None to mean "default to all Xs", will remove

src/qilisdk/readout/readout.py
Lumorti: I'll rework this

src/qilisdk/core/qtensor.py
Lumorti: will do!

src/qilisdk/readout/readout.py
```diff
@@ -152,28 +150,25 @@ def observables(self) -> list[Hamiltonian | QTensor]:
...
 
-    @property
-    def qtensor_observables(self) -> list[QTensor]:
-        """
-        The observables converted to :class:`~qilisdk.core.QTensor` form, populated automatically; not intended to be set manually.
-        """
-        return self._qtensor_observables
-
-    def expand_observables(self, nqubits: int) -> None:
+    def expand_observables(self, nqubits: int) -> list[QTensor]:
```
fedonman: typo, should be `expanded_observables`

Lumorti: fixed

fedonman (approved):
💯 💯

amirazzam2000 (approved):
Thanks luke!! This is very exciting!

PR #221, opened 2026-06-01
[SDK-282] Fix MPS Expectation Values

Lumorti:
Fixed a bug in which the `CudaBackend` would crash if the user requested expectation values or a state from an MPS or tensor simulation. Now an error is correctly raised, telling the user to use sampling instead. This information has also been added to the docs.


amirazzam2000 (approved):
Good catch! Thanks 🫶

PR #222, opened 2026-06-01
[SDK-257] Support .controlled and .adjoint chaining

Lumorti:
Support for the following gate operations has been added:
- `Gate.controlled(first_control).controlled(second_control)`
- `Gate.adjoint().controlled(control)`
- `Gate.controlled(control).adjoint()`
- `Gate.adjoint().adjoint()` (redundant, but for completeness)

### DIFF
--- a/changes/222.feature.md
+++ b/changes/222.feature.md
@@ -1,0 +1,5 @@
+Support for the following gate operations has been added:
+- `Gate.controlled(first_control).controlled(second_control)`
+- `Gate.adjoint().controlled(control)`
+- `Gate.controlled(control).adjoint()`
+- `Gate.adjoint().adjoint()` (redundant, but for completeness)
--- a/src/qilisdk/digital/gates.py
+++ b/src/qilisdk/digital/gates.py
@@ -493,6 +493,33 @@ class Controlled(Modified[TBasicGate]):
     def name(self) -> str:
         return "C" * len(self.control_qubits) + self.basic_gate.name
 
+    def adjoint(self) -> Controlled[TBasicGate]:
+        """
+        Returns the adjoint (conjugate transpose) of this controlled gate.
+
+        This method constructs and returns a new controlled gate where the underlying basic gate is replaced by its adjoint.
+        The control qubits remain the same, and the resulting gate's matrix is the conjugate transpose of the current gate's matrix.
+
+        Returns:
+            Controlled: A new Controlled gate instance representing the adjoint of this controlled gate.
+        """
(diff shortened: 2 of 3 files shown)


amirazzam2000 (changes requested):
I'm not sure I like this implementation, I would rather think of something that is a bit more generic. Maybe a chain of modified gates, that then all apply their corresponding matrix modification in the order they were chained. Rather than creating a new class for specifically a control and adjoint chaining.

Lumorti:
It did actually work before, it just wasn't type safe (like if you set the base gate of a controlled gate to be an adjoint gate)

This was the simplest possible solution I could think of that would be type safe

fedonman:
I agree that introducing a new gate for a specific case doesn't feel right. If we need to support this, I would try to find a generic solution.

Maybe the signature of `Modified` needs to change and instead of `self._basic_gate: TBasicGate = basic_gate` it allows for `_basic_gate: TBasicGate | Modified`, so use could do `Gate.controlled(first_control).controlled(second_control).controlled(third_control)....` etc. But the type is still `Controlled[X]` or `Controlled[Controlled[Controlled[X]]]`.

I am not sure, just an idea.

Question: Do we need this feature for a specific use-case or we add it for completeness?

fedonman:
Also, regardless of the implementation, transpilers and exporters/importers should get updated to any change.

Lumorti:
Regarding the use case, it was a linear task I added a while back when Devanshi was getting me to make some Shor's circuits, I think they needed controlled adjoint gates. I was able to do it with Controlled(qubit, Adjoint(base_gate)), which wasn't type safe but it worked. So either we need to forbid the user doing this or allow it, right now it's a middle ground. I'll work on the various fixes you've both mentioned

Lumorti:
I've been messing around today and yesterday trying to get it to work without another class, by allowing Modified to take another Modified, and it's really messy, because then you have to check for Modified[Adjoint[Controlled[Adjoint[Controlled]]]] and all sorts of weird stuff, ty really doesn't like the ability to do infinite chains. Meanwhile a new AdjointControlled class prevents such infinite chaining and let's Modified retain only taking a BasicGate

Lumorti:
Another proposal could be a bigger rework in which we don't have subclasses of Modified - Modified would have _control_qubits, _is_adjoint, _is_exponential properties inside it and thus would just become the generic "it's a basic gate with some extra features" class. Would make it easier in future to add new possible properties to modify too

Lumorti:
Can put this on the backburner for now, it's very non urgent

PR #223, opened 2026-06-01
[SDK-291] Fix QTensor Zero

Lumorti:
The behaviour of QTensor.zero(nqubits) has been fixed: rather than creating an immutable matrix of all zeros (which doesn't have much use), it now creates the |0...0> statevector as a user would expect.

Similarly, QTensor.one(nqubits) has been added to allow easy initialization of the |1...1> state.


amirazzam2000 (approved):
Nice!! This change makes sense 🙌

PR #225, opened 2026-06-01
[SDK-293] Add Model Constructors

Lumorti:
A number of constructors for Model representing common optimization problems have been added:

 - Model.random_ising()
 - Model.knapsack(values, weights, max_weight, label)
 - Model.max_cut(edges, weights, label)
 - Model.graph_coloring(graph, num_colors, label)
 - Model.travelling_salesman(edges, distances, label)
 - Model.factoring(n, label)
 
To help test these binary models, a ``BruteForceSolver`` class has been added that allows a global search for the optimum to help to know what the quantum optimization should achieve. This solver is of course only for small problems as it is exponential in the worst case (as with all of these NP-hard problems).

These can be all be used to help simplify the common optimization workflow, so that now to set up an annealing job for a randomized Ising model looks like:

```python
from qilisdk.analog import Schedule, X
from qilisdk.core import QTensor, Model
from qilisdk.utils.classical_solvers import BruteForceSolver

model = Model.random_ising(num_variables=5, label="Random Ising")
print(model)
print("Brute forced solution:", BruteForceSolver().solve(model))

# Define total time and timestep
T = 100.0
steps = 1000

# Define Hamiltonians
Hz = model.to_qubo().to_hamiltonian()
nqubits = Hz.nqubits

# Set up the schedule, initial state, analog evolution, and readout
analog_evolution = AnalogEvolution(
    schedule=Schedule.linear(Hx, Hz, T, dt=T/steps),
    initial_state=QTensor.uniform(nqubits),
)
readout = Readout().with_expectation([Hz])

# Execute the analog evolution
print("Executing analog evolution...")
results = QiliSim().execute(analog_evolution, readout)
print(results)
```


docs/modules/core/core_model.rst
```diff
@@ -119,6 +119,29 @@ Additional :class:`Constraints<qilisdk.core.model.Constraint>` can be added to r
...
 
+Constructors
+^^^^^^^^^^^^^^^^^^
+
+Rather than building a model from scratch, you can also construct it for certain predefined problems:
+
+ - :func:`Model.random_ising(num_variables, coefficient_range, label, seed)<qilisdk.core.model.Model.random_ising>` - Constructs a random Ising model with specified parameters.
+ - :func:`Model.knapsack(values, weights, max_weight, label, lagrange_multiplier)<qilisdk.core.model.Model.knapsack>` - Constructs a knapsack problem with given weights, values, and capacity.
+ - :func:`Model.max_cut(edges, weights, label)<qilisdk.core.model.Model.max_cut>` - Constructs a max cut problem for a given graph.
+ - :func:`Model.graph_coloring(graph, num_colors, label, lagrange_multiplier)<qilisdk.core.model.Model.graph_coloring>` - Constructs a graph coloring problem for a given graph and number of colors.
```
amirazzam2000, line 130: I would consider adding these on QUBO instead of Model.

src/qilisdk/core/model.py
```diff
@@ -510,6 +518,281 @@ def to_qubo(
...
         )
 
+    @classmethod
+    def knapsack(
+        cls,
+        values: list[Number],
+        weights: list[Number],
+        max_weight: float,
+        label: str = "Knapsack",
+        lagrange_multiplier: float = 100,
```
amirazzam2000, line 548: knapsack has an inequality constraint, so i would add another parameter here to specify the penalization method (slack or unbalanced)

amirazzam2000: I would actually expose these two parameters from add_constraint: 

```
        penalization: Literal["unbalanced", "slack"] = "slack",
        parameters: list[float] | None = None,
```

src/qilisdk/core/model.py
```diff
@@ -510,6 +518,281 @@ def to_qubo(
...
+                model.objective.term
+                + generator.uniform(low=coefficient_range[0], high=coefficient_range[1]) * variables[i]
+            )
+            for j in range(i + 1, num_variables):
+                model.set_objective(
+                    model.objective.term
+                    + generator.uniform(low=coefficient_range[0], high=coefficient_range[1])
+                    * variables[i]
+                    * variables[j]
+                )
```
amirazzam2000, line 626: isn't it more efficient to construct the term first then set_objective?

src/qilisdk/core/model.py
```diff
@@ -510,6 +518,281 @@ def to_qubo(
...
+        for n in nodes:
+            model.add_constraint(
+                f"one_color_{n}",
+                EQ(sum(x[n, k] for k in range(num_colors)), 1),
+                lagrange_multiplier=lagrange_multiplier,
+            )
+        for u, v in edges:
+            for k in range(num_colors):
+                model.add_constraint(
+                    f"conflict_{u}_{v}_{k}", LEQ(x[u, k] + x[v, k], 1), lagrange_multiplier=lagrange_multiplier
```
amirazzam2000: same comment as knapsack here

amirazzam2000: also consider using this formulation instead: 

Image

I would follow the formulation in here for these problems in QUBO format: 
https://www.frontiersin.org/journals/physics/articles/10.3389/fphy.2014.00005/full

amirazzam2000: if we can avoid an inequality constraint then we should do it.

src/qilisdk/core/model.py
```diff
@@ -510,6 +518,281 @@ def to_qubo(
...
+        Raises:
+            ValueError: if ``edges`` and ``distances`` have different lengths.
+        """
+        if len(edges) != len(distances):
+            raise ValueError("edges and distances must have the same length.")
+        n = max(node for edge in edges for node in edge) + 1
+        x = [[BinaryVariable(f"x{i}_{t}") for t in range(n)] for i in range(n)]
+        model = cls(label)
+        objective = sum(
+            distances[k] * (x[i][t] * x[j][(t + 1) % n] + x[j][t] * x[i][(t + 1) % n])
```
amirazzam2000: from my understanding the distances are not necessarily symmetric in the way you defined the input parameters. a user can enter edges: [(1, 2) , (2, 1) , ...] and assign each direction a different distance. either enforce the symmetry or change this into an asymmetric sum.

src/qilisdk/core/model.py
```diff
@@ -510,6 +518,281 @@ def to_qubo(
...
+            model.add_constraint(
+                f"city_{i}", EQ(sum(x[i][t] for t in range(n)), 1), lagrange_multiplier=lagrange_multiplier
+            )
+        for t in range(n):
+            model.add_constraint(
+                f"position_{t}", EQ(sum(x[i][t] for i in range(n)), 1), lagrange_multiplier=lagrange_multiplier
+            )
+        return model
+
+    def brute_force(self) -> tuple[dict[str, Number], dict[BaseVariable, RealNumber]]:
```
amirazzam2000: maybe we can externalize this to a solver module? not necessarily in this PR but in the future when we have the lp exportation we can try to integrate other solvers like groubi

amirazzam2000 (changes requested):
Thanks Luke!! This is a great idea to add these constructors. I left some comments, but overall good job!!

docs/modules/core/core_model.rst
Lumorti: The reason I put them in model is give them in their purest form

If the user wants to convert to a QUBO after, they can do so, giving their list of arguments about penalization methods and multiplier etc.

src/qilisdk/core/model.py
Lumorti: I was actually thinking about getting ride of the lagrange_multiplier here, and leaving it to the user when they do to_qubo()

amirazzam2000: I think it could be a quick config to add here and the user can then modify it in the to_qubo. but to stay consistent with the interface

docs/modules/core/core_model.rst
amirazzam2000: yeah fair, although in some cases we might be able to do smarter optimizations if we build it in QUBO directly. so maybe we can offer some overloads.

src/qilisdk/core/model.py
fedonman: I would add these formulas to the docstrings of every Factory method.

fedonman (changes requested):

src/qilisdk/core/model.py
Lumorti: add_constraint doesn't support specifying the penalization

src/qilisdk/core/model.py
Lumorti: Good idea!

src/qilisdk/core/model.py
Lumorti: So I vote for keep the model as just a plain model, and letting all of the qubification happen later using the to_qubo

src/qilisdk/core/model.py
Lumorti: Will do and will do, good ideas

src/qilisdk/core/model.py
Lumorti: Hmm, I'm not against it, maybe for now I'll put it as  `from qilisdk.utils.classical_solvers import BruteForceSolver` ? Then we have a ClassicalSolver class which has an overloadable .solve() that takes a model and returns a solution

amirazzam2000: I like that, maybe it would be under qilisdk.utils.solvers and then under there there is the BruteForceSolver. 
I like that!

src/qilisdk/core/model.py
amirazzam2000: ohhh right, fair okay :+1:

src/qilisdk/core/model.py
Lumorti: I think this should be fixed by the new format, but please double check

amirazzam2000 (changes requested):
I understand that now the model uses the ising formulation to construct the problems but that goes against the design on the model. What I mean to say is: 

- The QUBO formulations paper defines the problems as a hamiltonian already including the lagrange multiplier etc. which is great! However, our model needs to define separate constraints and cost so that the user can later change and modify the lagrange multipliers. 

- The formulation of QUBO is really nice, but when using the Model object we should keep the inequality constraints as inequality constrains and allow the user the ability to choose how they move them to QUBO. 

- Moreover, if we define these constructors on the Model class, we need to override them in the QUBO class to make sure that we add the extra parameters (like penalization) and that we translate the problems from Model to QUBO.

src/qilisdk/core/model.py
```diff
@@ -510,6 +511,291 @@ def to_qubo(
...
+            )
+
+        dist_terms = []
+        for k, (u, v) in enumerate(edges):
+            w = distances[k]
+            for t in range(n):
+                dist_terms.extend({w * x[u][t] * x[v][(t + 1) % n], w * x[v][t] * x[u][(t + 1) % n]})
+
+        if dist_terms:
+            objective = sum(dist_terms)
```
amirazzam2000, line 828: maybe I would throw an error if this doesn't happen, or I wouldn't defind dist_terms as a list but rather as a number and it can start from 0 and then you just say dist_terms += ...

src/qilisdk/utils/classical_solvers.py
```diff
@@ -0,0 +1,110 @@
...
+        .. code-block:: python
+
+            from qilisdk.core import Model
+            from qilisdk.utils.classical_solvers import BruteForceSolver
+
+            model = Model.knapsack(values=[5, 4], weights=[3, 2], max_weight=3)
+            results, sample = BruteForceSolver().solve(model)
+    """
+
+    def solve(self, model: Model) -> tuple[dict[str, Number], dict[BaseVariable, RealNumber]]:  # noqa: PLR6301
```
amirazzam2000, line 53: maybe here raise a warning for QUBOs/Models larger than 5 or 7 qubits saying hey this will take forever you are testing 2⁷ combinations

amirazzam2000 (approved):
I left some comments but overall looks good!

src/qilisdk/core/model.py
Lumorti: Added some raises!

src/qilisdk/utils/classical_solvers.py
Lumorti: Ah yes good point, I had that before but then forgot to put it back in refactor, will add

Lumorti: Actually wait, that exists already, it checks the total_combinations just below

fedonman (approved):
Thanks Luke! 💯

PR #226, opened 2026-06-01
[SDK-295] Refactor Variables into an Expression AST

### LINKED ISSUE #376: [Feature]: Replace the flattened Term polynomial with an Expression AST
### Feature Summary

Replace `Term` with a real expression tree rooted at a new `Expression` base class, with `Add`, `Mul`, `Pow` and `Constant` nodes and unary function nodes (`Sin`, `Cos`, `Exp`, `Log`, `Tan`, `Sqrt`, `Abs`). Everything symbolic in the SDK, gate angles, Hamiltonian and Schedule coefficients, Model objectives and constraints, becomes an `Expression`.

### Motivation / Use Case

`Term` is a flattened polynomial: a dict of factors to numeric coefficients plus one ADD or MUL operation. It bumps the coefficient of a repeated factor whatever the operation, which is right for a sum and wrong for a product, so `Sin(x) * Sin(x)` collapses to `2.0 * sin[x]` and quietly evaluates to 0.9589 instead of 0.2298. `MathematicalMap` subclasses `Term` to fake a function node, so `Sin(x).degree` reports 1 as if it were linear. There is no derivative, no substitution and no expand, and a fractional or symbolic exponent is impossible: `x ** 0.5`, `2 ** x` and `x ** y` all raise `TypeError`. Gradient work needs `diff`, and there is nowhere to put it in the flat model.

```python
import numpy as np
from qilisdk.core.variables import Parameter, Sin

x = Parameter("x", value=0.5)
p = Sin(x) * Sin(x)
print(p, "->", p.evaluate({x: 0.5}), "should be", np.sin(0.5) ** 2)
# (2.0) * (sin[x]) -> 0.958851077208406 should be 0.22984884706593015
print("degree:", Sin(x).degree)   # 1
x ** 0.5                          # TypeError
```

### Proposed Solution

A new `qilisdk.core.expression` module. Nodes canonicalize on construction (flatten, combine like terms and powers, fold constants, drop identities, order operands) and canonical form is the only definition of `==` and `__hash__`, so `x + y == y + x` still holds. Rewrites are opt-in and never affect equality.

```python
expr = Sin(x) ** 2 + Cos(x) ** 2
expr.diff(x).simplify()      # 0
(x * (y + 3)).expand()       # x*y + 3*x
(x ** y).degree              # NonPolynomialError instead of a wrong answer
```

Two things to settle: whether `Mul` keeps the factored form, which it should, with `expand()` to distribute, so `x*(y+z)` is no longer structurally equal to `x*y + x*z`; and how far `simplify` goes (constant folding, like-term collection and zero/one elimination are enough, trig identities are not).

This is breaking. `Term`, `Operation` and `MathematicalMap` leave the public API, and the YAML tags for symbolic expressions change, so Schedules, Models and Hamiltonians serialized by earlier versions will not load.

#226 already implements this.

fedonman:
Replaces `Term` with a real expression tree. Every symbolic value in the SDK is now an `Expression` built from `Add`, `Mul`, `Pow`, `Constant` and the unary function nodes, which gives us `diff`, `expand`, `simplify`, `substitute` and fractional or symbolic exponents. It also fixes a wrong-answer bug in the old flattened model, where `Sin(x) * Sin(x)` collapsed to `2 * sin(x)`. Breaking: `Term`, `Operation` and `MathematicalMap` are gone, `Pow` is now built with `**`, `Inv(x)` returns `x ** -1` instead of being a class, `Mul` no longer distributes on construction, and expressions serialized by earlier versions will not load. Closes #376.


src/qilisdk/core/variables.py
```diff
@@ -359,25 +315,21 @@ def _check_output(var: Variable, output: Number) -> RealNumber:
...
-        """encode the integer x in Bitwise encoding.
-
-        Args:
-            x (int): the integer to be encoded.
-            N (int): the number of bits to encode x.
+        """Encode the integer ``x`` using ``N`` bits (little-endian).
 
         Returns:
-            list[int]: a binary list representing the Bitwise encoding of the integer x.
+            list[int]: the little-endian binary digits of ``x``.
```
amirazzam2000, line 213: this comment is missing the parameter description.

src/qilisdk/core/variables.py
```diff
@@ -242,105 +252,51 @@ class ComparisonOperation(QiliEnum):
...
-            value (list[int] | int):  a list of binary values or an integer value.
-
-        Returns:
-            tuple[bool, int]: the boolean is True if the sample is a valid encoding,
-                                while the int is the error in the encoding.
-        """
+        """Whether ``value`` is a valid sample in this encoding (and the encoding error)."""
 
 
 def _check_output(var: Variable, output: Number) -> RealNumber:
```
amirazzam2000, line 168: This has inconsistent error messaging which can be confusing.

src/qilisdk/core/variables.py
```diff
@@ -414,8 +363,7 @@ def evaluate(var: Variable, value: list[int] | int, precision: float = 1e-2) ->
 
         binary_dict: dict[BaseVariable, list[int]] = {binary_var[i]: [binary_list[i]] for i in range(len(binary_list))}
```
amirazzam2000: the brackets here are not needed in theory without them it should be a tiny bit faster.

src/qilisdk/core/variables.py
```diff
@@ -776,163 +678,46 @@ def set_bounds(self, lower_bound: float | None, upper_bound: float | None) -> No
...
-        """Replaces the information of the variable with those coming from the dictionary
-        if the variable label is in the dictionary
+        """Replace this variable's domain and bounds.
 
         Args:
             domain (Domain): The updated domain of the variable.
-            bounds (tuple[float | None, float | None]): The updated bounds of the variable. Defaults to (None, None)
+            bounds (tuple[float | None, float | None]): The updated bounds. Defaults to (None, None).
         """
         self._hash_cache = None
```
amirazzam2000: if we are only hashing the variable by its label this shouldn't invalidate the cache

src/qilisdk/core/variables.py
```diff
@@ -1047,7 +837,7 @@ def precision(self) -> float:
         return self._precision
 
     @property
-    def term(self) -> Term:
+    def term(self) -> Expression:
```
amirazzam2000: should this be renamed to `expression`?

src/qilisdk/core/variables.py
```diff
@@ -1086,60 +876,53 @@ def update_variable(
...
 
     def num_binary_equivalent(self) -> int:
-        """
+        """Number of binary variables needed to encode the continuous variable.
+
         Returns:
-            int: the number of binary variables needed to encode the continuous variable.
+            int: the number of binary variables in the variable's encoding.
         """
         return self.encoding.num_binary_equivalent(self, precision=self._precision)
```
amirazzam2000: we should cache this value.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+    """Coerce a value into an :class:`Expression`.
+
+    Returns:
+        Expression | None: ``obj`` if it is already an ``Expression``, a :class:`Constant` wrapping a
+        number, or ``None`` if ``obj`` is neither.
+    """
+    if isinstance(obj, Expression):
+        return obj
+    if isinstance(obj, np.generic):
+        obj = obj.item()
```
amirazzam2000, line 123: we shouldn't modify the user's object, we should just return a new one. This isn't an in-place operation.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+
+    def as_coefficients_dict(self) -> dict[Expression, Number]:
+        """Map each (non-constant) monomial to its numeric coefficient.
+
+        Returns:
+            dict[Expression, Number]: a mapping from monomial to coefficient.
+        """
+        return {self: 1}
+
+    def monomial_factors(self) -> list[tuple[Expression, int]]:
```
amirazzam2000, line 365: we should keep the to_list method as well, because i added it because some people were using it. I know practically this is the same but it doesn't hurt to have an alias who's name is more intuitive.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+        return NotImplemented if rhs is None else Add.build((self, Mul.build((Constant(-1), rhs))))
+
+    def __rsub__(self, other: object) -> Expression:
+        lhs = _coerce(other)
+        return NotImplemented if lhs is None else Add.build((lhs, Mul.build((Constant(-1), self))))
+
+    __isub__ = __sub__
+
+    def __neg__(self) -> Expression:
+        return Mul.build((Constant(-1), self))
```
amirazzam2000, line 428: I think this should expand be default.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+class Constant(Expression):
+    """A numeric literal leaf. Replaces the old ``Term.CONST`` sentinel."""
+
+    def __init__(self, value: Number) -> None:
+        raw: Any = value
+        if isinstance(raw, np.generic):
+            raw = raw.item()
+        if isinstance(raw, bool):
+            raw = int(raw)
+        self.value: Number = _float_if_real(raw)
```
amirazzam2000: if we don't make this a hidden variable then the user can modify it which should invalidate the cache but currently this doesn't happen.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+    def degree(self) -> int:
+        """Polynomial degree. Raises :class:`NonPolynomialError` for non-polynomial expressions."""
+
+    @abstractmethod
+    def diff(self, symbol: BaseVariable) -> Expression:
+        """Symbolic derivative with respect to ``symbol``."""
+
+    @abstractmethod
+    def _sort_key(self) -> tuple:
+        """A total-order key used to order operands deterministically and define equality."""
```
amirazzam2000: can we add an explanation to the tuple structure? the one in constant just seems arbitrary without any explanation

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+        return coefficients
+
+    def _sort_key(self) -> tuple:
+        return (4, tuple(term._sort_key() for term in self.args))  # ruff: ignore[private-member-access]
+
+    def _compute_hash(self) -> int:
+        return qili_hash("Add", self.args)
+
+    def __repr__(self) -> str:
+        return " + ".join(repr(term) for term in self.args)
```
amirazzam2000: this prints things like `-2 + -2 * b` I think this should be handled and replaced with either: 
- `-2 - 2 * b`, 
- or `-2 + (-2) * b`

for clarity. I know this comment should probably be in the Mul class but I just put it here in general.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+
+    Exponents of zero drop out, exponents of one leave the bare base, and an idempotent base (a
+    binary variable) collapses any positive integer power back to itself.
+
+    Returns:
+        list[Expression]: the surviving factors.
+    """
+    factors: list[Expression] = []
+    for base, raw_exponent in powers.items():
+        exponent = Constant(1) if (base.is_idempotent_under_mul and _is_pos_int_const(raw_exponent)) else raw_exponent
```
amirazzam2000, line 217: I don't think it makes sense to have a BinaryVariable with negative exponent. We should throw an error if that happens.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+    def monomial_factors(self) -> list[tuple[Expression, int]]:
+        factors: list[tuple[Expression, int]] = []
+        for factor in self.args:
+            if isinstance(factor, Constant):
+                continue
+            factors.extend(factor.monomial_factors())
+        return factors
+
+    def _sort_key(self) -> tuple:
+        return (3, tuple(factor._sort_key() for factor in self.args))  # ruff: ignore[private-member-access]
```
amirazzam2000: this 3 seems arbitrary. Can we at least define constants to make it more readable.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+
+from .types import Number, RealNumber
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+    from .variables import BaseVariable, Parameter
+
+_TOL = get_settings().atol
+_DIVISION_MESSAGE = "Division by zero is not allowed"
```
amirazzam2000:
```suggestion
_DIVISION_MESSAGE = "Division by zero encountered during Expression evaluation."
```

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+        exponent = _int_exponent(self.exp)
+        if exponent is not None and exponent >= 0:
+            return self.base.degree * exponent
+        raise NonPolynomialError(f"Expression {self!r} is not a polynomial; its degree is undefined.")
+
+    def diff(self, symbol: BaseVariable) -> Expression:
+        base, exp = self.base, self.exp
+        if symbol not in exp.free_symbols():
+            # d/dx b**c = c * b**(c-1) * b'
+            return Mul.build((exp, Pow.build(base, Add.build((exp, Constant(-1)))), base.diff(symbol)))
```
amirazzam2000: update comment to `# d/dx x**c = c * x**(c-1) * x'` or `# (b**c)' = c * b**(c-1) * b'`

amirazzam2000 (changes requested):
Thank you Vyron!! this is a very good refactor to the variables and terms. It streamlines everything. 
I left some minor comments.

changes/226.feature.md
Lumorti, line 1: This could maybe be made a bit clearer, I'm yet to review the rest of the PR but from this I don't immedietly get whether my old Terms like `2 * BinaryVariable + 1` are still going to work (but just now called Expressions rather than Terms) or whether I need to change my syntax.

Lumorti, line 1: Also, regarding the name of `Expression.diff()`, is that standard in other libraries? I thought we tried not to use abbreviations, like maybe it should be `Expression.derivative(), unless it's the `.diff()` is industry standard, to me `diff` makes me think of difference or `git diff` rather than derivatives.

changes/226.removal.md
Lumorti, line 1: Do we normally do remove changelogs? I thought normally we just incorporate it into the main feature log

I'm okay with it being separate, but make sure at the end of this PR that they are both consistent with each other, and make sure that towncrier supports "removal" as a keyword in the final changelog

docs/modules/core/core_terms_maps_comparisons.rst
```diff
@@ -14,29 +18,47 @@ known as :class:`Terms<qilisdk.core.variables.Term>`. For example:
...
+    e3: -1 + b + 2 * x
+    e4: 3 + -1 * (4 + 2 * x + 3 * x**2) + 2 * x
 
-Terms can be evaluated by providing values for the involved variables:
+Construction *canonicalizes* the expression (flattening nested sums/products, combining like terms
+and powers, folding constants, ordering operands deterministically), which is why the numeric
+constant is printed first and ``x + y`` equals ``y + x``. Canonicalization is intentionally cheap:
+products are **not** distributed over sums, so ``e4`` keeps the factored ``-1 * (4 + 2 * x + 3 * x**2)``
+sub-expression. Use :meth:`~qilisdk.core.expression.Expression.expand` to distribute, and
+:meth:`~qilisdk.core.expression.Expression.simplify` to request a simpler (but semantically equal) form:
```
Lumorti: Rewrote to be less LLM-y:

```suggestion
Upon construction we canonicalize the expression (flattening nested sums/products, combining like terms, ordering things etc.), ensuring that ``x + y + 1`` equals ``y + 1 + x``. Products are **not** distributed over sums, so ``e4`` keeps the factored ``-1 * (4 + 2 * x + 3 * x**2)``
sub-expression. Use :meth:`~qilisdk.core.expression.Expression.expand` to distribute:
```

docs/modules/core/core_terms_maps_comparisons.rst
```diff
@@ -14,29 +18,47 @@ known as :class:`Terms<qilisdk.core.variables.Term>`. For example:
...
 
-    t3.evaluate({
+    print(e4.expand())          # distribute the product over the sum
+
+**Output**:
+
+::
+
+    -1 + -3 * x**2
+
```
Lumorti, line 53: Could use an example for `Expression.simplify()` here, the above text mentioned it but didn't use it

docs/modules/core/core_terms_maps_comparisons.rst
```diff
@@ -49,45 +71,143 @@ Terms can be evaluated by providing values for the involved variables:
...
-    To evaluate a term, all participating variables must be assigned valid values within their respective domains and bounds.
+    To evaluate an expression, all participating variables must be assigned valid values within their respective domains and bounds.
+
+Inspecting and differentiating expressions
+===========================================
+
+An :class:`~qilisdk.core.expression.Expression` exposes a small introspection API. You can list the
+named leaves it depends on, isolate just the free :class:`~qilisdk.core.variables.Parameter` leaves,
+read its polynomial :attr:`~qilisdk.core.expression.Expression.degree`, and take a symbolic
+derivative with :meth:`~qilisdk.core.expression.Expression.diff`:
```
Lumorti:
```suggestion
An :class:`~qilisdk.core.expression.Expression` exposes a number of helper functions. You can list the
named leaves it depends on, isolate just the free :class:`~qilisdk.core.variables.Parameter` leaves,
read its polynomial :attr:`~qilisdk.core.expression.Expression.degree`, or take a symbolic
derivative with :meth:`~qilisdk.core.expression.Expression.diff`:
```

docs/modules/core/core_terms_maps_comparisons.rst
```diff
@@ -49,45 +71,143 @@ Terms can be evaluated by providing values for the involved variables:
...
     print(expr.evaluate({theta: 1.0}))
 
-These maps compose naturally with other terms, so you can include them in
-constraints or objectives and rely on the same evaluation and encoding rules
-as other symbolic expressions.
+**Output**:
+
+::
+
+    cos(2 * theta) + sin(theta)
```
Lumorti: I think if we are making the user capitalize the functions then the print should also be capitalized, like `print(Sin(x))` should print `Sin(x)` not `sin(x)`

docs/modules/core/core_terms_maps_comparisons.rst
```diff
@@ -49,45 +71,143 @@ Terms can be evaluated by providing values for the involved variables:
...
+
+
+Mathematical Functions
+======================
+
+Non-polynomial operations are represented by :class:`~qilisdk.core.expression.Function`, the abstract
+base for the unary maths functions. Its concrete subclasses
+:class:`~qilisdk.core.expression.Sin`, :class:`~qilisdk.core.expression.Cos`,
+:class:`~qilisdk.core.expression.Exp`, :class:`~qilisdk.core.expression.Log`,
+:class:`~qilisdk.core.expression.Tan`, :class:`~qilisdk.core.expression.Sqrt` and
```
Lumorti: we list all of the possible operations below, so maybe we don't need them here as well

docs/modules/core/core_terms_maps_comparisons.rst
```diff
@@ -49,45 +71,143 @@ Terms can be evaluated by providing values for the involved variables:
...
+- :class:`~qilisdk.core.expression.Exp` for exponential
+- :class:`~qilisdk.core.expression.Log` for logarithm
+- :class:`~qilisdk.core.expression.Sqrt` for square root
+- :class:`~qilisdk.core.expression.Abs` for absolute value
+
+``Abs`` is the one function with no derivative: it is not differentiable at zero and there is no
+``sign`` node to write its derivative with, so :meth:`~qilisdk.core.expression.Expression.diff`
+raises on it.
+
+Powers are not functions. Use the ``**`` operator, which builds a
```
Lumorti:
```suggestion
For powers, use the ``**`` operator, which builds a
```

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+    """Normalize a numeric evaluation result.
+
+    Returns:
+        Number: a ``float`` for (near-)real values, otherwise the complex value unchanged.
+    """
+    if isinstance(value, RealNumber):
+        return float(value)
+    if isinstance(value, complex) and abs(value.imag) < _TOL:
+        return float(value.real)
+    return value
```
Lumorti, line 110: isn't this the same as `_float_if_real()`?

src/qilisdk/core/expression.py
Lumorti: It's somewhat useless, but since we can support it (in case it gets done automatically by some expansion etc.) then maybe we should. The =1 case is handled here, and the =0 will raise a division warning

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+    """
+    left_terms = left.args if isinstance(left, Add) else (left,)
+    right_terms = right.args if isinstance(right, Add) else (right,)
+    return Add.build(tuple(Mul.build((x, y)) for x in left_terms for y in right_terms))
+
+
+class Expression(ABC):
+    """Abstract base of every node in the expression tree."""
+
+    _TOL = _TOL
```
Lumorti: huh?

src/qilisdk/core/expression.py
Lumorti: I agree!

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+
+@yaml.register_class
+class Mul(Expression):
+    """A canonical n-ary product. Collects like powers and folds the numeric coefficient.
+
+    ``Mul`` does **not** distribute over sums; use :meth:`expand` for that.
+    """
+
+    def __init__(self, args: tuple[Expression, ...]) -> None:
+        # Trusting constructor: ``args`` must already be canonical. Use ``_build`` to normalize.
```
Lumorti: does `_build` exist? should it be `build` instead? or is this an outdated comment from an earlier version?

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,986 @@
...
+        self.arg: Expression = operand  # ty:ignore[invalid-assignment]
+        self._hash_cache: int | None = None
+
+    @staticmethod
+    @abstractmethod
+    def _numeric(value: RealNumber) -> Number:
+        """Numeric kernel applied to an already-evaluated, real operand."""
+
+    @abstractmethod
+    def _derivative(self, operand: Expression) -> Expression:
```
Lumorti: yeah I deffo prefer `.derivative()` to `.diff()`, even if it is more verbose

src/qilisdk/core/model.py
```diff
@@ -761,13 +762,11 @@ def ising(
...
         x = {n: BinaryVariable(f"x{n}") for n in nodes}
-        list_of_terms: list[BaseVariable | Term | Number] = [
-            field_map[n] * x[n] for n in nodes if field_map.get(n, 0) != 0
-        ]
+        list_of_terms: list[Expression] = [field_map[n] * x[n] for n in nodes if field_map.get(n, 0) != 0]
         list_of_terms.extend((1 if couplings is None else couplings[i]) * x[u] * x[v] for i, (u, v) in enumerate(edges))
 
         model = cls(label)
-        model.set_objective(Term(list_of_terms, Operation.ADD), sense=ObjectiveSense.MINIMIZE)
+        model.set_objective(Add.build(tuple(list_of_terms)), sense=ObjectiveSense.MINIMIZE)
```
Lumorti, line 759: Maybe Add(...) should default to this `Add.build(...)` behavior, then use a different internal constructor for when we know the terms are canonical already

That way if the user wants to use these functions themselves it's easier, they can do `Add(Add(x, y), z)` etc. rather than having `.build` everywhere

src/qilisdk/core/expression.py
fedonman: Nothing is mutated, `obj` is the local parameter and those lines just rebind it. `.item()` and `int()` both return new values, so the caller's object is untouched. Renamed the local to `raw` to make that read better.

src/qilisdk/core/expression.py
fedonman: Tried it and backed it out: partial distribution isn't confluent, so `2*b*(x+y)` and `b*(2*(x+y))` stop comparing equal. It's either never distribute or always fully expand, and the latter is what `.expand()` is for. Fixed the printing half though, see your other comment.

src/qilisdk/core/expression.py
fedonman: It's also raised at construction (`x / 0`), not just evaluation, so I used "Division by zero is not allowed in an Expression."

src/qilisdk/core/variables.py
fedonman: Done.

src/qilisdk/core/variables.py
fedonman: Done.

src/qilisdk/core/variables.py
fedonman: Done.

src/qilisdk/core/variables.py
fedonman: Done.

src/qilisdk/core/variables.py
fedonman: Done.

src/qilisdk/core/variables.py
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

src/qilisdk/core/variables.py
```diff
@@ -776,163 +678,46 @@ def set_bounds(self, lower_bound: float | None, upper_bound: float | None) -> No
...
+            bounds (tuple[float | None, float | None]): The updated bounds. Defaults to (None, None).
         """
         self._hash_cache = None
         self._domain = domain
         self.set_bounds(bounds[0], bounds[1])
 
-    @abstractmethod
-    def to_binary(self) -> Term:
-        """Returns the binary representation of a variable.
+    # ------------------------------------------------------------------ Expression interface
```
Lumorti: Nit: do we use this type of comment divider elsewhere? I thought normally it was
normally more like `----- Title -----` when I have seen it, but I could be wrong

src/qilisdk/core/variables.py
Lumorti: well maybe the hash should also include the rest, probably safer to invalidate it and then also make sure we're hashing with all info

Lumorti (changes requested):
Thanks Vyron! Overall I really like the changes, I think it's a nice generalization and will be easier to expand etc. later, left a few comments

Also, I believe there are some Term optimizations included in the Simulated Annealing PR, so if that is merged first please make sure that performance is similar, since for big Ising models there were some slowdowns caused by lack of hashing in the old Term class

src/qilisdk/core/variables.py
amirazzam2000: it used to but then we decided to remove it, i don't recall why but it was giving an error somewhere.

src/qilisdk/core/expression.py
fedonman: Renamed to `derivative()`, no alias kept. The `diff` name came from SymPy, which is what I had followed, but you're right that it clashes with `numpy.diff` (successive differences) and we avoid abbreviations anyway.

changes/226.feature.md
fedonman: Renamed to `derivative()`, no alias kept. The `diff` name came from SymPy, which is what I had followed, but you're right that it clashes with `numpy.diff` (successive differences) and we avoid abbreviations anyway.

docs/modules/core/core_terms_maps_comparisons.rst
fedonman: Done.

docs/modules/core/core_terms_maps_comparisons.rst
fedonman: Done.

docs/modules/core/core_terms_maps_comparisons.rst
fedonman: Done.

src/qilisdk/core/expression.py
fedonman: Done.

changes/226.feature.md
fedonman: Done. Added a line saying `2 * BinaryVariable("b") + 1` still works exactly as before, you just get an `Expression` instead of a `Term`.

docs/modules/core/core_terms_maps_comparisons.rst
fedonman: Done, `Sin(x)` now prints as `Sin(x)`. `NAME` stays lowercase as the internal yaml/hash tag.

docs/modules/core/core_terms_maps_comparisons.rst
fedonman: Done, dropped the inline list.

src/qilisdk/core/expression.py
fedonman: That was `_TOL = _TOL`, a class attribute shadowing the module constant. Removed.

src/qilisdk/core/expression.py
fedonman: Outdated, it's `build`. Fixed.

src/qilisdk/core/variables.py
fedonman: You're right, `sabre_swap_pass.py` and `hamiltonian.py` close them on both sides. Fixed all 6 of mine.

changes/226.removal.md
fedonman: `removal` is a towncrier built-in and we already ship them (`changes/300.removal.md`), CHANGELOG has a "Deprecations and Removals" section. Checked with `towncrier build --draft`.

docs/modules/core/core_terms_maps_comparisons.rst
fedonman: `simplify()` turned out to be a no-op: construction already canonicalizes, so it could never change anything. Removed it rather than fake an example.

src/qilisdk/core/expression.py
fedonman: Close but not the same: `_float_if_real` keeps an int an int, so `Constant(2)` prints `2`; `_finalize` always widens to float for `evaluate()`. Documented both.

src/qilisdk/core/model.py
fedonman: Can't be done in `__init__`: `Add.build((x, Constant(0)))` returns the bare `x` and `Add.build(())` a `Constant`, so it needs `__new__` returning a non-Add, which breaks `isinstance`.

src/qilisdk/core/expression.py
fedonman: Ameer asked for the opposite in this thread, so pick whichever and I'll match it. My reasoning: `b**-1` is 1 at b=1 and undefined at b=0, and positive powers already collapse before the check.

src/qilisdk/core/variables.py
fedonman: Bounds are mutable, so hashing them breaks dicts already keyed on the variable: `{v: 3}` stops resolving after `v.set_bounds(...)`. Kept label-only.

src/qilisdk/core/expression.py
```diff
@@ -683,17 +753,32 @@ class Pow(Expression):
...
     def build(cls, base: Expression, exp: Expression) -> Expression:
         if isinstance(exp, Constant):
             if exp.value == 1:
                 return base
             if exp.value == 0:
                 return Constant(1)
+            if base.is_idempotent_under_mul and isinstance(exp.value, RealNumber) and exp.value < 0:
+                raise NotSupportedOperation(
+                    f"{base!r} is binary, so {base!r}**{exp.value} is 1/{base!r}, which is undefined when "
+                    f"{base!r} is 0. Negative powers of a binary variable are not supported."
```
Lumorti: this doesn't check if base.value == 0, no? so if you have a binary variable b = 1 and you try to do b ** -1 then it tells you it's undefined because the base is zero, which is wrong

the only thing that is undefined is if base.value == 0 and exp.value < 0

Lumorti (approved):
Thanks Vyron, I'm excited to have this in main! It's a really nice refactor. Approved, but left one small comment to take a look at before merging.

src/qilisdk/core/expression.py
fedonman: Nice catch! Fixed it.

changes/226.removal.md
```diff
@@ -0,0 +1 @@
...teger/symbolic powers of a variable, or a function of a variable); `Pow` is now a node built with `**` whose exponent is an `Expression`, so `Pow(x, 2.0)` becomes `x ** 2`; `Inv` is a helper that returns `x ** -1` rather than a class, so `isinstance(expr, Inv)` no longer works; a negative power of a `BinaryVariable` now raises `NotSupportedOperation` instead of building a node that is undefined at 0; expression nodes are immutable, so `Constant.value`, `Add.args`, `Mul.args`, `Pow.base`, `Pow.exp` and `Function.arg` are read-only; `Variable.term` is renamed to `Variable.expression`; `ComparisonTerm` is renamed to `Comparison` and, with `ComparisonOperation` and the `LT`/`LEQ`/`EQ`/`NEQ`/`GT`/`GEQ` helpers, now lives in the new `qilisdk.core.comparison` module rather than `qilisdk.core.variables` (all of them are still re-exported from `qilisdk.core`); `BaseVariable.evaluate` takes an environment mapping like every other node, so `variable.evaluate(bits)` becomes `variable.evaluate({variable: bits})`; and the YAML serialization tags for symbolic expressions changed, so expression-bearing documents (Schedules, Models, Hamiltonians) serialized by earlier versions will not deserialize.
```
amirazzam2000: my understanding that binary power is supported now, so i guess we should remove this line

> a negative power of a `BinaryVariable` now raises `NotSupportedOperation` instead of building a node that is undefined at 0;

```suggestion
**Breaking change.** Removed `Term`, `Operation`, and `MathematicalMap` from `qilisdk.core.variables`; use `Expression` and the node classes (`Add`, `Mul`, `Pow`, `Constant`, `Sin`, `Cos`, ...) instead, importing them from `qilisdk.core` or `qilisdk.core.expression`. Consequences of the new model: `Mul` no longer auto-distributes, so `x * (y + z)` stays factored until `.expand()` is called (it is no longer structurally equal to `x*y + x*z`); `Expression.degree` raises `NonPolynomialError` for non-polynomial expressions (e.g. non-integer/symbolic powers of a variable, or a function of a variable); `Pow` is now a node built with `**` whose exponent is an `Expression`, so `Pow(x, 2.0)` becomes `x ** 2`; `Inv` is a helper that returns `x ** -1` rather than a class, so `isinstance(expr, Inv)` no longer works; expression nodes are immutable, so `Constant.value`, `Add.args`, `Mul.args`, `Pow.base`, `Pow.exp` and `Function.arg` are read-only; `Variable.term` is renamed to `Variable.expression`; `ComparisonTerm` is renamed to `Comparison` and, with `ComparisonOperation` and the `LT`/`LEQ`/`EQ`/`NEQ`/`GT`/`GEQ` helpers, now lives in the new `qilisdk.core.comparison` module rather than `qilisdk.core.variables` (all of them are still re-exported from `qilisdk.core`); `BaseVariable.evaluate` takes an environment mapping like every other node, so `variable.evaluate(bits)` becomes `variable.evaluate({variable: bits})`; and the YAML serialization tags for symbolic expressions changed, so expression-bearing documents (Schedules, Models, Hamiltonians) serialized by earlier versions will not deserialize.
```

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,1064 @@
...
+        return self._arg.free_symbols()
+
+    @property
+    def degree(self) -> int:
+        if self._arg.free_symbols():
+            raise NonPolynomialError(f"Expression {self!r} is not a polynomial; its degree is undefined.")
+        return 0
+
+    def derivative(self, symbol: BaseVariable) -> Expression:
+        return Mul.build((self._outer_derivative(self._arg), self._arg.derivative(symbol)))
```
amirazzam2000, line 912: There are cases where this would fail because the derivative is not supported even when the variable being derived is not included in the expression. 

For example: 
```
x = Variable("x", domain=Domain.REAL)
y = Variable("y", domain=Domain.REAL)

print((Abs(y) + 1).derivative(x))

> NotSupportedOperation: The derivative of Abs is not supported.
```
where in this case the derivative of abs(y) should just be 0. So maybe we can add a check for that: 


```suggestion
    def derivative(self, symbol: BaseVariable) -> Expression:
        if symbol not in self.free_symbols():
             return Constant(0)  
        return Mul.build((self._outer_derivative(self._arg), self._arg.derivative(symbol)))
```

src/qilisdk/core/variables.py
amirazzam2000: can we add a default domain? something like: Domain.INTEGER and then the user can modify it if they want. it makes creating variables smoother. 

I picked integer as the default here because real variables are expensive to binarize.

src/qilisdk/core/expression.py
```diff
@@ -0,0 +1,1064 @@
...
+        if getattr(cls, "NAME", ""):
+            Function._REGISTRY[cls.NAME] = cls
+
+    def __new__(cls, arg: object) -> Expression:
+        operand = _coerce(arg)
+        if operand is None:
+            raise TypeError(f"{cls.__name__} expects an Expression or number, got {type(arg).__name__}")
+        if isinstance(operand, Constant):
+            return Constant(cls._numeric(_assert_real(operand.value)))
+        return super().__new__(cls)
```
amirazzam2000, line 875: This is causing the deepcopy of functions to break. This affects some aspects of the library that used to work but now break. 

After looking into it, it seems that adding this would fix it:
```
   def __getnewargs__(self) -> tuple[Expression]:
       return (self._arg,)
```

code to reproduce: 
```

from qilisdk.digital import Circuit, H, RX
from qilisdk.digital.circuit_transpiler_passes import CancelIdentityPairsPass
from qilisdk.core import Parameter, Sin

p = Parameter("p", 2)
c = Circuit(2)
c.add(H(0))
c.add(RX(0, theta=Sin(p)))
c.add(H(0))

CancelIdentityPairsPass().run(c)

# or 

c_copy = deepcopy(c)
```

amirazzam2000 (commented):
Thanks a lot vyron it looks really good! 
I left some comments with some potential bugs.

changes/226.removal.md
fedonman: Right, removed.

src/qilisdk/core/expression.py
fedonman: Good catch, added the guard. It fixes the composed cases too, `(Abs(y) * x).derivative(x)` now gives `Abs(y)`.

src/qilisdk/core/expression.py
fedonman: Thanks for tracking that down, added. `Function.__new__` needs the operand and `copy`/`pickle` had no way to hand it back. Your repro is in the tests now.

src/qilisdk/core/variables.py
fedonman: Done, `Domain.INTEGER`. Consistent with `Parameter`, which already defaults its domain.

amirazzam2000 (approved):
Thanks a lot Vyron! I am excited to have users start using this 🙌

PR #227, opened 2026-06-03
[SDK-297] Fixing the density matrix simulation for digital backends in qilisim

amirazzam2000:
Fixed a bug in the `QiliSim` matrix-free digital simulator where applying a complex, non-symmetric single-qubit gate (such as `U2` or `U3`) to a *mixed* density matrix computed `U·ρ·U*` (complex conjugate) instead of `U·ρ·U†` (conjugate transpose). This produced a non-Hermitian state and could raise a spurious "imaginary expectation value" error. Real-symmetric gates (`X`, `Z`, `H`) and diagonal gates (`S`, `T`) were unaffected. This most commonly surfaced in `QuantumReservoir` runs, where resetting a qubit turns the state mixed before the next layer's input-encoding gate is applied.


fedonman (approved):
Nice catch! 💯

PR #228, opened 2026-06-03
[SDK-294] Added New Schedule Constructors

Lumorti:
Two new `Schedule` constructors have been added: `Schedule.constant()` and `Schedule.linear_list()`.

Schedule.constant(hamiltonian, total_time, dt) allows for quickly defining the time evolution of a fixed Hamiltonian.

Schedule.linear_list(list_of_hamiltonians, total_time, dt) allows for the definition of a schedule in which one does a linear ramp between successive Hamiltonians. The coefficients look like the following:

image

Usage is the same as with the other Schedule constructors:

```python

nqubits = 3
H1 = -sum(X(i) for i in range(nqubits))
H2 = sum(Z(i) for i in range(nqubits))
H3 = sum(Z(i) * Z((i+1) % nqubits) for i in range(nqubits))

schedule_constant = Schedule.constant(H1, total_time=10.0, dt=0.01)
schedule_constant.draw()

schedule_list = Schedule.linear_list([H1, H2, H3], total_time=10.0, dt=0.01)
schedule_list.draw()
```

Also, a few small tweaks were made to the `docs.sh` and `missing_translations.sh`: `docs.sh` now correctly checks the changelog entries, whilst `missing_translations.sh` now correctly identifies all cases of fuzzy translations.


src/qilisdk/analog/schedule.py
```diff
@@ -231,6 +231,80 @@ def sinusoidal(
...
+        return cls(
+            hamiltonians={"constant": hamiltonian},
+            coefficients={"constant": {0.0: 1}},
+            dt=dt,
+            total_time=total_time,
+            interpolation=Interpolation.LINEAR,
+        )
+
+    @classmethod
+    def linear_list(
```
amirazzam2000: Idk if this one is needed. Like it's too specific for me. Where did you get this schedule from? is it from a paper where they show this is useful? 
because maybe I would start with a linear schedule + a catalyst before i do this.

Lumorti: I feel like it's a natural extension of a linear ramp, no? To allow from H1 -> H2 -> H3? If you want to do anything with more than two Hamiltonians, this is the simplest way to do that

Lumorti: I think we should just add a bunch of pre-built schedules, if you have any more you want me to add, I will, like it would make it really easy for anyone doing annealing research to just try all of the different schedules that could possibly exist

Lumorti: I can also add Schedule.linear_with_catalyst(H_init, H_final, H_catalyst, T, dt) if you want?

changes/228.feature.md
```diff
@@ -0,0 +1,25 @@
...
+
+nqubits = 3
+H1 = -sum(X(i) for i in range(nqubits))
+H2 = sum(Z(i) for i in range(nqubits))
+H3 = sum(Z(i) * Z((i+1) % nqubits) for i in range(nqubits))
+
+schedule_constant = Schedule.constant(H1, total_time=10.0, dt=0.01)
+schedule_constant.draw()
+
+schedule_list = Schedule.linear_list([H1, H2, H3], total_time=10.0, dt=0.01)
```
fedonman: I wouldn't use `list` since it's not informative of what the method is doing, aside that it expects a list. What about `ramp_between` or `linear_ramp` or `chained_linear_ramp` or something similar.

Lumorti: I do prefer all those names, do you think we should also change `Schedule.linear` to be `linear_ramp` too? Or maybe this should be `chained_linear` or `linear_between` to keep it consistent with `.linear`?

amirazzam2000: I like chained_linear as it conveys the idea perfectly.

docs/modules/analog/analog_schedule.rst
```diff
@@ -7,6 +7,8 @@ The simplest way to construct a common schedule is to use of the helper function
 - :func:`quadratic <qilisdk.analog.schedule.Schedule.quadratic>`: Quadratic interpolation between two Hamiltonians.
 - :func:`polynomial <qilisdk.analog.schedule.Schedule.polynomial>`: Polynomial interpolation of arbitrary degree between two Hamiltonians.
 - :func:`sinusoidal <qilisdk.analog.schedule.Schedule.sinusoidal>`: Sinusoidal interpolation between two Hamiltonians.
+- :func:`constant <qilisdk.analog.schedule.Schedule.constant>`: Constant coefficient for a single Hamiltonian.
+- :func:`linear_list <qilisdk.analog.schedule.Schedule.linear_list>`: Linear chaining of multiple Hamiltonians.
```
amirazzam2000: maybe you need to update the name here as well

Lumorti: good catch!

amirazzam2000 (approved):
thank you Luke!! 🙌

changes/228.feature.md
amirazzam2000: before merging could you also update the change log

Lumorti: This found an interesting bug! the doc checker wasn't checking the changelogs for some time, I refactored that script in one the docs PRs a while back and forgot to make it check the changes folder, it was only checking the docs folder

PR #230, opened 2026-06-08
[SDK-300] Experiment Plot Tweaks

Lumorti:
The plotting functions for the experiment classes have been updated:
- fitting is now supported for most experiment types and can be toggled on/off
- each of the experiment plots can also be done in dB if requested
- the tests have been updated to use data that more accurately represents the experiment
- units for various quantities have been added

Usage remains the same as before, but now with optional parameters `fit` and `db`:
```python
experiment_result.plot(db=True, fit=False)
```

image
image
image
image
image
image


amirazzam2000 (approved):
thanks luke!

PR #232, opened 2026-06-09
[SDK-301] More Plot Tweaks

Lumorti:
The various experiment plots have been further improved based on comments from the experimentalists, mostly regarding the units and formatting.

image

image

image

### DIFF
--- a/src/qilisdk/experiments/experiment_result.py
+++ b/src/qilisdk/experiments/experiment_result.py
@@ -217,6 +217,7 @@ class ExperimentResult(FunctionalResult):
 
         x_values = [x_override.transform(x) for x in x_values] if x_override else x_values
         y_values = [y_override.transform(y) for y in y_values] if y_override else y_values
+        z_values = z_override.transform(s21) if z_override else s21
 
         x_edges = np.linspace(x_values[0].min(), x_values[0].max(), len(x_values[0]) + 1)
         y_edges = np.linspace(y_values[0].min(), y_values[0].max(), len(y_values[0]) + 1)
@@ -227,7 +228,7 @@ class ExperimentResult(FunctionalResult):
         ax1.set_ylabel(y_override.label if y_override and y_override.label else y_labels[0])
         ax1.ticklabel_format(axis="both", style="sci", scilimits=(-3, 3))
 
-        mesh = ax1.pcolormesh(x_edges, y_edges, s21.T, cmap="viridis", shading="auto")
+        mesh = ax1.pcolormesh(x_edges, y_edges, z_values.T, cmap="viridis", shading="auto")
         z_label = z_override.label if z_override and z_override.label else r"$|S_{21}|$"
         colorbar_label = z_label + " ∝ Voltage" if not db else z_label + " (dB)"
         fig.colorbar(mesh, ax=ax1, label=colorbar_label)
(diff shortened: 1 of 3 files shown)


src/qilisdk/experiments/experiment_result.py
```diff
@@ -113,8 +130,138 @@ def add_fit(
...
+
+        Args:
+            figure (Figure): The Matplotlib figure to save.
+            save_to (str | Path): The path or directory where the figure should be saved.
+        """
+        save_to = Path(save_to)
+        if save_to.is_dir():
+            save_to /= f"{self.plot_title}_qubit{self.qubit}.png"
+        save_to.parent.mkdir(parents=True, exist_ok=True)
+        figure.savefig(save_to)
```
Lumorti, line 134: Note for reviewers, this (and the plot_1d and plot_2d) was moved outside of plot() for sonarqube refactoring reasons

Lumorti:
The codecov is not technically true, it's just because I moved some functions around that weren't previously tested either, all genuinely new lines _have_ been tested

src/qilisdk/experiments/experiment_result.py
```diff
@@ -43,6 +43,20 @@ def __init__(self, labels: list[str], values: list[np.ndarray]) -> None:
         self.values = values
 
 
+class DimOverride:
```
amirazzam2000: maybe we can unify this with the Dimension class, the label is already it's own override but we can add a transform callable to transform the data we are passing.

amirazzam2000 (changes requested):

src/qilisdk/experiments/experiment_result.py
fedonman: I agree, a new class is not needed. This can be a type alias a callable that takes Dimension and returns another Dimension, `DimensionOverride = Callable[[Dimension], Dimension]`

src/qilisdk/experiments/experiment_result.py
```diff
@@ -55,8 +69,11 @@ class ExperimentResult(FunctionalResult):
     plot_title: ClassVar[str]
     """Default plot title; subclasses provide the concrete label."""
 
-    colorbar_label: ClassVar[str] = r"$|S_{21}|$"
-    """Default colorbar label for 2D plots; can be overridden by subclasses if needed."""
+    dims_override: ClassVar[dict[int, DimOverride]] = {}
```
fedonman: a `list[DimOverride | None]` makes sense too.

amirazzam2000 (approved):
Okay we can just merge this! Thanks Luke 🫶

src/qilisdk/experiments/experiment_result.py
```diff
@@ -113,8 +130,136 @@ def add_fit(
...
+            s21 (np.ndarray): The S21 data to plot, either in linear or dB scale.
+            dims (list[Dimension]): The dimensions of the experiment, used for labeling axes.
+            db (bool): Whether the data is in dB scale, affecting axis labels and fit behavior.
+            fit (bool): Whether to perform and plot the fit using the `add_fit` method.
+            save_to (str | None): Optional path or directory to save the figure.
+            initial_guess (list[float] | None): Optional initial guess passed to `add_fit`.
+        """
+        x_labels, x_values = dims[0].labels, dims[0].values
+        x_override = self.dims_override.get(0)
+        y_override = self.dims_override.get(1)
```
fedonman: This is a side-product of not mapping Dimension to Dimension: There is no `dims[1]` but here we assume it via `dims_override`.

src/qilisdk/experiments/experiment_result.py
amirazzam2000: The dimensions override should be `Callable[[Dimension], [Dimension]]` no?

PR #233, opened 2026-06-09
Hotfix - minor plot tweaks

Lumorti:
The colorbar label no longer forces a Voltage/dB label to be appended to the unit.


amirazzam2000 (approved):
Thanks!

PR #234, opened 2026-06-09
Hotfix - Even More Plot Tweaks

Lumorti:
The db=False/True flag has been removed, since now with the dimension overrides it can cause problems.

connect_points=False/True has been added to plots at the suggestion of Yifei, which can be useful for debugging Rabi plots.

The T2 fit has been made more consistent with the T1 fit, with units corrected.

### DIFF
--- a/src/qilisdk/experiments/experiment_result.py
+++ b/src/qilisdk/experiments/experiment_result.py
@@ -105,9 +105,7 @@ class ExperimentResult(FunctionalResult):
         return 20 * np.log10(self.s21_modulus)
 
     @staticmethod
-    def add_fit(
-        x_values: np.ndarray, y_values: np.ndarray, initial_guess: list[float] | None = None, db: bool = False
-    ) -> None:
+    def add_fit(x_values: np.ndarray, y_values: np.ndarray, initial_guess: list[float] | None = None) -> None:
         """
         Fit a user-provided function to the experimental data.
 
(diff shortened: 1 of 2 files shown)


fedonman (dismissed):
IMO, You remove a nice feature because of a bug. And we will definitely want to add the feature back which will be double effort.

Why not remove the scaling of s21 from Dimensions which is the way to go?

As explained in `qili-runtime`, the following code should not set the `Amplitude (V)` as dimension override:

```python
T1ExperimentResult.dims_override = [
        lambda dim: Dimension(labels=[r"Time ($\mu$s)"], values=[v * 1e-3 for v in dim.values]),
        lambda dim: Dimension(labels=[r"Amplitude (V)"], values=dim.values),
    ]
```

T1 has only **one** dimension: Time (and of course the trivial I/Q which is regarded since we combine the values for s21)

The result data **is not** a dimension.

fedonman (approved):
I leave it to approved, so you can merge if it's pressing, but read the comments please :)

Lumorti:
> IMO, You remove a nice feature because of a bug. And we will definitely want to add the feature back which will be double effort.
> 
> Why not remove the scaling of s21 from Dimensions which is the way to go?
> 
> As explained in `qili-runtime`, the following code should not set the `Amplitude (V)` as dimension override:
> 
> ```python
> T1ExperimentResult.dims_override = [
>         lambda dim: Dimension(labels=[r"Time ($\mu$s)"], values=[v * 1e-3 for v in dim.values]),
>         lambda dim: Dimension(labels=[r"Amplitude (V)"], values=dim.values),
>     ]
> ```
> 
> T1 has only **one** dimension: Time (and of course the trivial I/Q which is regarded since we combine the values for s21)
> 
> The result data **is not** a dimension.

Do you suggest that we do a data override and a dims override as separate things? The current version is like this for simplicity, but I'm happy to change it. I think in one of the plots we do change the scaling of the data, no? Or if not, it's good to have. It's true that in this case we're pretending that the data is a dim, or more that we're treating the dims as graph axes rather than true dims.

We talked about removing dB in the last meeting. Also, it's not so much removing a feature, rather choosing not to include this (new, existing only briefly in a single hotfix commit) feature in the release. Also it complicated everything quite a lot, there were a lot of "if db" checks everywhere.

Lumorti:
Going to merge this for now, the comments are good suggestions though, I'll add a linear task with them as a reminder to have a look at it in future if we keep the experiments around

PR #235, opened 2026-06-10
[SDK-310] Add Phase Plots

Lumorti:
Phase plots for each experiment have been added, enabled via a new argument: `.plot(plot_type = "phase")`.

dB plots are also back (since it was only a few lines extra), which can be enabled with `plot_type = "db"`.

The default axis names have been changed from "|S21|" to "Amplitude (V)" to be more consistent.

### DIFF
--- a/src/qilisdk/experiments/experiment_result.py
+++ b/src/qilisdk/experiments/experiment_result.py
@@ -266,6 +266,7 @@ class ExperimentResult(FunctionalResult):
     def plot(
         self,
         save_to: str | None = None,
+        *,
         initial_guess: list[float] | None = None,
         fit: bool | None = None,
         connect_points: bool = False,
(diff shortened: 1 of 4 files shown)


src/qilisdk/experiments/experiment_result.py
```diff
@@ -248,6 +267,7 @@ def plot(
         initial_guess: list[float] | None = None,
         fit: bool | None = None,
         connect_points: bool = False,
+        plot_type: Literal["amplitude", "phase", "db"] = "amplitude",
     ) -> None:
```
fedonman, line 274: Since parameters are increasing, and some have default `None`, could we enforce using the keywords for all except the `save_to`?

```
def plot(
        self,
        save_to: str | None = None,
        *,
        initial_guess: list[float] | None = None,
        fit: bool | None = None,
        connect_points: bool = False,
        plot_type: Literal["amplitude", "phase", "db"] = "amplitude",
    ) -> None:
```

fedonman (commented):
Awesome 💯 Left a minor comment.

src/qilisdk/experiments/experiment_result.py
Lumorti: good idea

fedonman (approved):
💯 💯

PR #236, opened 2026-06-10
[SDK-308] Add `averages` to all experiments and their results

fedonman:
- Added `averages` as a common field for experiment functionals and experiment results.
- Renamed `TwoTonesAtFluxBiasExperiment` to `TwoTonesAtFixedFluxBiasExperiment`.

### DIFF
--- a/src/qilisdk/experiments/experiment_functional.py
+++ b/src/qilisdk/experiments/experiment_functional.py
@@ -175,7 +175,7 @@ class T2Experiment(ExperimentFunctional[T2ExperimentResult]):
 
 
 @yaml.register_class
-class TwoTonesAtFluxBiasExperiment(ExperimentFunctional[TwoTonesAtFixedFluxBiasExperimentResult]):
+class TwoTonesAtFixedFluxBiasExperiment(ExperimentFunctional[TwoTonesAtFixedFluxBiasExperimentResult]):
     """Two-tone spectroscopy functional for a single qubit.
 
     Sweeps a drive tone frequency while monitoring the readout tone to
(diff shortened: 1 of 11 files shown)


src/qilisdk/experiments/experiment_functional.py
```diff
@@ -60,6 +62,15 @@ def qubit(self) -> int:
         """
         return self._qubit
 
+    @property
+    def averages(self) -> int:
+        """The physical qubit index on which the experiment is performed.
+
+        Returns:
+            int: Index of the qubit.
+        """
```
Lumorti, line 71: This is copied from the qubit property, should be updated for averages

src/qilisdk/experiments/experiment_functional.py
```diff
@@ -238,14 +261,15 @@ def __init__(
 
         Args:
             qubit (int): The physical qubit index on which the experiment is performed.
+            averages (int): Number of averages to acquire for the experiment.
```
Lumorti, line 264: In general I think the word "averages" is strangely used, like I think here "averages" is the number of shots/samples/repeats **per** average, no? The actual number of different averages is determined by the dimension sweep, as in, one average is one data point, obtained by multiple shots

NataliaSirera (approved):

Lumorti (approved):
Thanks for handling this! 🚀️

A few comments, I leave up to you:
 - the term "averages" here is used to mean "shots/repeats", which I don't know if that's standard or not. If that's the term that experimentalists use, then completely fair, but to me it seems strange to call the number of shots used to generate each average the "averages"
 - right now on the results side, we didn't change the format of the results, right? so we're still just getting a single number per point (which as of now is the average over multiple runs), but I feel like maybe they're going to want error bars on the plots, which would require either the standard deviation or a list of all of the different values used to generate the average, which would require much more of a change. I don't know if we want to pre-emptively do this or not, perhaps that's what they're already thinking about when they talk about averages

Note the docstring typo before merging!

src/qilisdk/experiments/experiment_functional.py
fedonman: I had the same comment, but Lorenzo insisted on using the term `averages`.

PR #237, opened 2026-06-11
[SDK-313] Add Printing for Experiments

Lumorti:
`ExperimentResult` can now be printed without error, previously it was using the `FunctionalResult` `__repr__` (since it inherits from `FunctionalResult`), but now it has its own dedicated print.


fedonman (approved):
💯

PR #238, opened 2026-06-15
fix(security): QSDK-01/02 safe-by-default YAML deserialization + QSDK-03 docs

DZPM:
## QSDK-01 + QSDK-02 (🔴 Critical, CWE-502 / CWE-94)

Makes QiliSDK's YAML deserialization **safe by default**, closing two Critical findings from the June 2026 security analysis that share one root cause: the module-level `QiliYAML(typ="unsafe")` loader executes attacker-controlled code at parse time (`!function`/`!lambda` run `dill.loads`; `!type`/`!PydanticModel`/`!defaultdict` and ruamel's `!!python/*` import/instantiate arbitrary names).

- **QSDK-01** — public `deserialize` / `deserialize_from` fed untrusted input straight into that loader (loading a shared file = code execution). Finding: https://app.notion.com/p/37d7eec14c5381e48d4cf683f2f75a28
- **QSDK-02** — the SpeQtrum client routed the server's job-result payload through the same loader inside pydantic before-validators, so a malicious/compromised backend reached code execution on the client during a normal `submit` → `wait_for_job` flow. Finding: https://app.notion.com/p/37d7eec14c5381c99729d0b9da1d6851

### The fix
- New **registry-isolated, data-only** `safe_yaml` loader (`src/qilisdk/yaml.py`): its constructor gets fresh per-instance constructor tables (ruamel's `add_constructor` mutates a shared *class* registry, so a plain second instance is not isolated). It keeps the pure-data tags (`!ndarray`, `!complex`, `!tuple`, `!deque`, ...), **rejects** `!function`/`!lambda`, gates `!type`/`!PydanticModel`/`!defaultdict` **and** ruamel's five `!!python/*` multi-constructors behind a strict fully-qualified-name allow-list, and raises on unknown tags. `register_class` mirrors data tags onto it so all existing `!ClassName` round-trips keep working.
- `deserialize` / `deserialize_from` use `safe_yaml` **by default**, with an additive default-off `trust_code=True` escape hatch for genuinely trusted input. `serialize` and the trusted loader are unchanged.
- **QSDK-02** needs no further source change (the 20 SpeQtrum result/payload validators call the now-safe `deserialize` without `trust_code`); a regression test proves a `!function` server payload is rejected, not executed.

## QSDK-03 — documentation (🟠 Medium, CWE-1188)

Folded in from the now-closed #242. Adds a **"Serialization and Security"** page under Modules/Core (`docs/modules/core/core_serialization.rst`) documenting that `deserialize` / `deserialize_from` are safe by default (data-only loader that rejects code-execution tags, treat like `json.load`) and that the `trust_code=True` opt-in executes arbitrary code and must only be used on fully trusted input, with a prominent `danger` admonition. Finding: https://app.notion.com/p/37d7eec14c5381049d53c63b6d4bb3c2

Bundled here because it documents the behaviour that ships in this PR, so it no longer needs to be a separate stacked PR.

### Verification
113 tests pass across the serialization + SpeQtrum suites (incl. new regression tests); a 726-test sweep over analog/functionals/experiments/etc. confirms no legitimate round-trip regressed; `ruff` + `mypy` clean. Manual adversarial checks confirm every code-bearing tag (incl. nested + all `!!python/*` variants) raises with no sentinel file created.

### Not closed by this PR
- A non-allow-listed pydantic `BaseModel` subclass that previously round-tripped via the generic `!!python/object` tag now raises by default (LOW impact: none flow over the network path; `trust_code=True` restores it). Widen the allow-list if a legitimate round-trip needs it.


amirazzam2000 (commented):
Do we think this is a bit of an overkill? We can simply remove the serialization of lambda and functions once we decouple from qililab.

DZPM:
Rebased and cherry-picked #242 on top of this, with @Lumorti's recommendations.

Now the decision is to, either:
- remove the serialization (per this comment: https://github.com/qilimanjaro-tech/qilisdk/pull/238#pullrequestreview-4504081634)
- or applying this deserialization fix (at least, until the serialization gets removed?)

PR #239, opened 2026-06-16
[SDK-290] Stabilizer State Simulator

Lumorti:
A new digital simulation method has been added to QiliSim, based on stabilizer states. This allow efficient simulation of circuits with small numbers of T gates. For example, to sample from a 1000-qubit GHZ state in around 10s on a laptop:

```python
from time import time
from qilisdk.backends import QiliSim, DigitalMethod

# Settings
nqubits = 1000
nshots = 10000

# Circuit creating a GHZ state
c = Circuit(nqubits)
c.add(H(0))
for i in range(1, nqubits):
    c.add(CNOT(i - 1, i))

# Set everything up
digital = DigitalPropagation(circuit=c)
backend_stab = QiliSim(digital_simulation_method=DigitalMethod.stabilizer())
readout = Readout().with_sampling(nshots)

# Run it
t0 = time()
results_stab = backend_stab.execute(digital, readout)
t1 = time()
samples_stab = results_stab.get_samples()
print(samples_stab)
print(f"Execution time: {t1 - t0} seconds")
``` 

It can also do more general circuits, however either a performance cost (scaling as 2^{number of T gates}) or an accuracy cost (if truncation is enabled via setting `max_states` to something non-zero).


.claude/skills/run-checks/SKILL.md
amirazzam2000, line 1: We should add this in a separate PR when we decide on how we are doing this.

scripts/checks.sh
```diff
@@ -10,6 +10,9 @@
 # Stop if any command fails
 set -euo pipefail
 
+# Make sure we're in the venv
+source .venv/bin/activate
```
amirazzam2000, line 17: this assumes you are running the checks from the base of the library. maybe add something to check this?

src/qilisdk/backends/qilisim.py
```diff
@@ -131,7 +132,10 @@ def get_config(self) -> SolverConfigDict:
         return dict(self._solver_config)
 
     def _execute_digital_propagation(
-        self, functional: DigitalPropagation, readout: list[ReadoutMethod], initial_state: QTensor | None = None
+        self,
+        functional: DigitalPropagation,
+        readout: list[ReadoutMethod],
+        initial_state: QTensor | InitialState = InitialState.ZERO,
```
amirazzam2000, line 138: should this be something that we add to the rest of the backends as well? because as of now the user has no way of modifying this parameter.

Lumorti: the syntax of the function didn't change, just that it now accepts an InitialState instead of a QTensor

I believe the logic with this syntax from before was that in general we don't want to allow circuits with initial states since we can't do that on the hardware, but I think you or someone else wanted it for reservoirs, maybe? so the compromise was to allow it if the user uses the _execute overload directly

scripts/checks.sh
Lumorti: good idea!

.claude/skills/run-checks/SKILL.md
Lumorti: Yes sorry I forgot to remove this

scripts/checks.sh
Lumorti: done

src/qilisdk/backends/qilisim.py
amirazzam2000: Good point, Maybe a cool thing to have is pre-made circuit initializers, were we have some premade circuits that you can prepend to a circuit to initialize the state into a given state. Maybe we can define it in an easy to use way. Not for this PR, just something to consider for the future (maybe as part of the circuit blocks idea we were working on)

Lumorti: Yeah that could be nice, tbf the current InitialState initializers are just zero, uniform or one, which are very easy to implement as prepended circuits, so maybe DigitalPropogation could take either an InitialState or a Circuit. I'll add a linear task for it!

src/qilisdk/backends/backend_config.py
```diff
@@ -337,6 +337,12 @@ class DigitalMethod(BaseSimulatorConfig):
             for statevector simulation. Defaults to ``True``.
     """
 
+    sampling_method: str = Field(
```
amirazzam2000: I know this is annoying but maybe we should call this propagation_method for consistency?

amirazzam2000: or even just method (given that we also change the analog one to be method)

amirazzam2000 (approved):
It looks good! Just left a small comment, we can discuss it! 🙌

src/qilisdk/backends/backend_config.py
Lumorti: we need to seperate digital and analog since maybe we use both for reservoirs etc., but yeah good idea it shouldn't be sampling anymore

PR #240, opened 2026-06-17
fix(security): QSDK-04 coerce from_numpy input to contiguous complex128

DZPM:
## Summary

Fixes **QSDK-04** (Medium, CWE-843 / CWE-125) from the QiliSDK security analysis.

Finding: https://app.notion.com/p/37d7eec14c5381abae83fa006b1c4cd5

`from_numpy()` reinterpreted the raw numpy buffer as `complex128` after only checking `ndim == 2`, never verifying itemsize / dtype / contiguity. A non-`complex128` array (`float64`, or every gate matrix under `QILISDK_COMPLEX_PRECISION=COMPLEX_64`) or a non-contiguous view strides 16 bytes per element over a smaller / mismatched allocation, reading out of bounds; under `COMPLEX_64` this was a deterministic OOB read on any circuit run.

## Fix

Coerce the buffer to a C-contiguous `complex128` array (`py::array::c_style | py::array::forcecast`) before reading. This covers every caller (`QTensorCpp`, `parse_gates`, observables) since they all route through `from_numpy`, and makes the `COMPLEX_64` path produce correct values instead of reading past the buffer. Adds a regression test.

Verified locally: built and 140 QTensor tests pass (incl. the new regression).

### DIFF
--- a/tests/unit_python/core/test_qtensor.py
+++ b/tests/unit_python/core/test_qtensor.py
@@ -63,6 +63,14 @@ def test_from_numpy_coerces_mismatched_dtype_and_strides():
     assert np.allclose(result, np.ascontiguousarray(view))
 
 
+def test_from_numpy_raises_when_array_is_not_convertible_to_complex128():
+    from qtensor_module import QTensorCpp  # noqa: PLC0415
+
+    bad = np.array([["not-a-number"]], dtype=object)
+    with pytest.raises(ValueError, match=r"Input array must be convertible to a 2D complex128 array\."):
+        QTensorCpp(bad)
+
+
 def test_constructor_valid_sparse():
     """QTensor should accept a valid SciPy sparse matrix."""
     sparse_mat = csc_array([[1, 0], [0, 1]])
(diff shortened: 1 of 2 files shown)


tests/unit_python/core/test_qtensor.py
```diff
@@ -43,6 +43,26 @@ def test_constructor_valid_ndarray():
...
+        assert np.allclose(result, expected), f"dtype {np.dtype(dtype).name} mismatched"
+
+    # Non-contiguous complex128 view: strides must be honoured, not ignored.
+    base = np.arange(64, dtype=np.complex128).reshape(8, 8)
+    view = base[::2, ::2]
+    assert not view.flags["C_CONTIGUOUS"]
+    result = np.asarray(QTensorCpp(view).as_numpy())
+    assert np.allclose(result, np.ascontiguousarray(view))
+
+
```
Lumorti, line 65: @copilot could you please also add a test here checking the throw in the case of the array not being able to be converted?

Lumorti (approved):
Thanks David/Claude/Copilot! Good catch

PR #241, opened 2026-06-17
fix(security): QSDK-05/06 reject out-of-range qubit indices at the trust boundary

DZPM:
## Summary

Fixes **QSDK-05** (Medium, CWE-125 / CWE-787) and the closely related **QSDK-06** (Low, CWE-787) from the QiliSDK security analysis.

Findings: [QSDK-05](https://app.notion.com/p/37d7eec14c5381c0b1ebc1866a31833e) and [QSDK-06](https://app.notion.com/p/37d7eec14c5381da8770db5cf7e8ec3b)

The matrix-free kernels computed a shift mask from an unvalidated qubit index, and `parse_measurements` wrote `final_qubits_to_measure[target]` with no bounds check. The Python layer only rejected `qubit >= nqubits` (never `qubit < 0`), and deserialization reconstructs `Circuit` / `Gate` / `Pauli` objects without re-running `__init__`, so a crafted or negative index reached the C++ kernels (undefined shift, wild mask, out-of-bounds state access, QSDK-05) and the measurement vector (out-of-bounds write, QSDK-06).

## Fix

- **C++ (keystone):** validate every control / target / Pauli / measurement index against the circuit size in `parse_gates`, `parse_hamiltonians_matrix_free` and `parse_measurements`, before it reaches the kernels (covers the deserialization-bypass path). Defensively guard the shift in `MatrixFreeOperator::apply`.
- **Python (defense-in-depth):** reject `qubit < 0` in `Circuit._add` / `_insert` and `PauliOperator.__init__`.
- Regression tests for the Python guards and the C++ parser guard via a deserialization-style bypassed gate.

Verified locally: built and 605 digital/analog + 16 backend + 5 new regression tests pass; a bypassed out-of-range gate now raises `ValueError` instead of reading out of bounds.

### DIFF
--- a/src/qilisdk_cpp/backends/qilisim/representations/matrix_free_operator.cpp
+++ b/src/qilisdk_cpp/backends/qilisim/representations/matrix_free_operator.cpp
@@ -14,6 +14,7 @@
 
 #include "matrix_free_operator.h"
 #include <sstream>
+#include <stdexcept>
 #include "../../../libs/pybind.h"
 
 // GCOV_EXCL_BR_START
@@ -36,7 +37,18 @@ void MatrixFreeOperator::apply(DenseMatrix& output_state, MatrixFreeApplicationT
 
     // Precompute things that are used in all branches
     int num_qubits = static_cast<int>(std::log2(output_state.rows()));
-    long long mask = 1LL << (num_qubits - 1 - target_qubits[0]);
+    // QSDK-05 defense-in-depth: indices are validated at parse time, but this
+    // kernel is also reachable via deserialized objects, so guard the shift so
+    // a bad target can never produce an undefined shift / wild mask.
+    if (target_qubits.empty()) {
+        throw std::out_of_range("MatrixFreeOperator has no target qubit.");
+    }
+    int shift = num_qubits - 1 - target_qubits[0];
+    if (shift < 0 || shift >= 64) {
+        throw std::out_of_range("MatrixFreeOperator target qubit " + std::to_string(target_qubits[0]) +
+                                " is out of range for a " + std::to_string(num_qubits) + "-qubit state.");
+    }
+    long long mask = 1LL << shift;
     long N = output_state.rows();
     long stride = mask;
     long half = N >> 1;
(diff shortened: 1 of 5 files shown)


src/qilisdk_cpp/backends/qilisim/representations/matrix_free_operator.cpp
```diff
@@ -36,7 +37,18 @@ void MatrixFreeOperator::apply(DenseMatrix& output_state, MatrixFreeApplicationT
...
     int num_qubits = static_cast<int>(std::log2(output_state.rows()));
-    long long mask = 1LL << (num_qubits - 1 - target_qubits[0]);
+    // QSDK-05 defense-in-depth: indices are validated at parse time, but this
+    // kernel is also reachable via deserialized objects, so guard the shift so
+    // a bad target can never produce an undefined shift / wild mask.
+    if (target_qubits.empty()) {
+        throw std::out_of_range("MatrixFreeOperator has no target qubit.");
+    }
+    int shift = num_qubits - 1 - target_qubits[0];
+    if (shift < 0 || shift >= 64) {
```
Lumorti: @copilot why is this 64? is that a magic number?

DZPM: Good catch. The `64` is the bit-width of the `1LL << shift` mask, so the guard is really "is `shift` a valid bit position for the mask". One nuance on the autofix: the mask is a **signed** `long long` (`1LL`), so the precise bound is `std::numeric_limits<long long>::digits` (= 63 value bits), not the `unsigned long long` digits (= 64) the autofix used. The two differ only at `shift == 63` (the sign bit), which is reachable only with a 64-qubit statevector (2^64 amplitudes), so it never happens in practice — and this guard is belt-and-suspenders anyway, since indices are already validated against the circuit size at the parse boundary (`parse_gates` / `parse_hamiltonians_matrix_free`). Happy to switch it to `std::numeric_limits<long long>::digits` (63) if you'd prefer the exact-for-the-type bound, or change the mask to `1ULL` so 64 is correct.

Lumorti: @copilot please implement the above tweak, changing the mask to 1ULL

Lumorti (approved):
Thanks! Very nice check to have

PR #242, opened 2026-06-17
docs(security): QSDK-03 add a serialization security page

DZPM:
## Summary

Closes the documentation half of **QSDK-03** (Medium, CWE-1188) from the QiliSDK security analysis.

Finding: https://app.notion.com/p/37d7eec14c5381049d53c63b6d4bb3c2

Adds a "Serialization and Security" getting-started page documenting that `deserialize` / `deserialize_from` are **safe by default** (a data-only loader that rejects code-execution tags) and that the `trust_code=True` opt-in **executes arbitrary code** and must only be used on fully trusted input, with a prominent `danger` admonition.

## Note

The safe-by-default loader + `trust_code` opt-in themselves ship in #238 (QSDK-01/02). This page documents that behaviour, so **it should merge after #238**.


Lumorti:
I would suggest that this not be in the introduction to QiliSDK, it's not something new users need to know about. Maybe it should be in the GitHub readme since it's developer specific, or in the "Core" section.

DZPM:
@Lumorti good point, moved it to core_serialization

Lumorti:
It's good to have this, and I'm happy with it in being core (thanks for moving it). My only worry now is with this we are exposing the idea to users that _somewhere_ in the Qili stack we execute unsafe code, since otherwise we wouldn't have this option. Are we certain the stack is safe if the user starts putting malicious code? If so, then it's safe to make users aware of this.

DZPM:
It's a by design limitation of YAML serialization (& pickle, & others). To rebuild an arbitrary object you have to allow constructing arbitrary objects, and that's where code execution goes in.
 
 To answer directly: yes, the default is safe.
 - The data-only loader from #238 only reconstructs a whitelist of known QiliSDK and primitive types and rejects code-execution tags.
 - An arbitraty/malicious YAML loaded the normal way can't run code. You'd have to explicitly pass `trust_code=True` to opt into the unrestricted loader, on input you chose to trust.

 It's not a "100% forever" guarantee (new YAML exploits will appear), but applying the "defense in depth" concept:
 - the safe loader is the layer that removes the accidental-execution risk, and `trust_code` is the deliberate escape hatch with a loud warning (useful when adding/editing QiliSDK primitives)
 - having one centralized place for the loader means every future YAML mitigation lands in a single spot that's easy to maintain.


Alternatively, we just remove the serialization, as @amirazzam2000 said: https://github.com/qilimanjaro-tech/qilisdk/pull/238#pullrequestreview-4504081634

DZPM:
Superseded by #238: the QSDK-03 serialization-security docs have been folded into that PR (cherry-picked both commits) so the safe-by-default loader and its documentation ship together as a single PR. Closing this one.

PR #245, opened 2026-06-18
[SDK-258] Add Time-dependent Lindblad Noise

Lumorti:
Support for time-dependent Lindblad noise has been added, for now only supported in QiliSim. Usage is as follows:

```python
from qilisdk.noise import LindbladGenerator
import numpy as np
J1 = QTensor(np.array([[0, 1], [0, 0]]))
T = 100
lindblad_noise = LindbladGenerator(jump_operators=[J1], rates=[lambda t: 0.005 * (1 + np.sin(t / T))])
```

By extension, `QutipBackend` now also supports general noise for analog evolution (but for now still not for digital).


src/qilisdk/backends/qutip_backend.py
```diff
@@ -79,13 +80,25 @@ class QutipBackend(Backend):
...
                 ODE solver. Defaults to ``10_000``.
+            noise_model (NoiseModel | None): Must be ``None``. The QuTiP
+                backend does not support noisy simulation; accepted only so
+                that passing a noise model raises a clear error rather than
+                an opaque ``TypeError``.
+
+        Raises:
+            NotImplementedError: If a noise model is provided.
         """
+        if noise_model is not None:
```
amirazzam2000: actually can we support linblad noise here? it should be a very easy mapping from qilisdk to qutip. And we already solve the master equation

amirazzam2000 (commented):
I really like this approach it's simple and does the job! Thanks luke!! 

I just left a comment about QuTip if you think it will be too much to add analog noise there then tell me and I approve. But I think having QuTip for benchmarking and comparing could be a good idea.

src/qilisdk/backends/qutip_backend.py
Lumorti: sure, I'll look into it

Lumorti: Actually, right now our QuTiP backend doesn't support any noise model:

`NotImplementedError: QutipBackend does not support noise models (neither digital channels nor Lindblad/analog noise). Use QiliSim for noisy simulations.`

Should I add support for all noise models here? Or maybe I make a separate task for that?

amirazzam2000: I was thinking about only supporting analog as this is what QuTip is known for. I don't mind the fact that the digital in QuTip is not supported.

Lumorti: done!

fedonman (approved):
LGTM 💯 Thank you @Lumorti 

Note: For completeness, I would add a task for adding digital noise to QutipBackend.

PR #247, opened 2026-06-19
[SDK-321] Fix Variational Shots

Lumorti:
Fixed a bug in which the shots were not set as expected for the new Variational Annealing method.


Lumorti:
(This is basically a one-line change PR, the rest is the linter catching stuff in main)

src/qilisdk_cpp/backends/qilisim/analog/time_evolution.cpp
Lumorti, line 1: (This is the actual code change, the shots were being overwritten by the `get_num_monte_carlo_trajectories()` shots rather than the `config.get_shots()` above)

src/qilisdk/backends/backend_config.py
```diff
@@ -123,12 +123,12 @@ class AnalogMethod(BaseSimulatorConfig):
         description="Tolerance for the adaptive integrator method when `evolution_method='integrate_rk45_matrix_free'`.",
     )
     variational_shots: int = Field(
-        default=1000,
+        default=100,
```
DZPM, line 126: Should we add the 10x drop to the changelog?

DZPM (approved):
LGTM. Consider documenting the "silent" 10x drop, as it'll have some impact in the user behaviour

src/qilisdk/backends/backend_config.py
Lumorti: This is actually done since it was defaulting to 100 before (the default of the monte carlo shots), so this change is actually keeping it consistent with what it was before!

PR #248, opened 2026-06-22
[SDK-181] Add check to make sure Schedule has a total time

Lumorti:
Added a check to ensure that a Schedule has some sort of max time on construction, either via total_time or via the coefficients.


src/qilisdk/analog/schedule.py
```diff
@@ -38,6 +38,7 @@
 InterpDict = dict[str, "Interpolator"]
 
 _DEFAULT_DT = 0.1
+_DEFAULT_T = 0.1
```
amirazzam2000: what is the purpose of this variable?

amirazzam2000 (approved):
Thanks for catching this!

src/qilisdk/analog/schedule.py
Lumorti: Literally just noticed I left this in, clearing it up now!

PR #250, opened 2026-06-22
[SDK-307] Add support for global optimizers

Lumorti:
Support has been added for the SciPy global optimizers: `direct`, `dual_annealing`, `differential_evolution`, `shgo` and `basinhopping`, as per https://docs.scipy.org/doc/scipy/tutorial/optimize.html#global-optimization. They are quite slow, but they are guarenteed to find the global minimum, useful for testing small variational programs and ensuring that the ansatz covers the true solution. Usage is the same as choosing any other SciPy optimizer:

```python
from qilisdk.optimizers.scipy_optimizer import SciPyOptimizer
optimizer = SciPyOptimizer(method="shgo")
```

### DIFF
--- a/src/qilisdk/optimizers/scipy_optimizer.py
+++ b/src/qilisdk/optimizers/scipy_optimizer.py
@@ -50,7 +50,7 @@ class SciPyOptimizer(Optimizer):
                     - 'trust-constr
                     - 'dogleg'
                     - 'trust-ncg'
-                    - 'trust-exact
+                    - 'trust-exact'
                     - 'trust-krylov'
                     - 'basinhopping' (global)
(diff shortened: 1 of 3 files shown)


fedonman (approved):
LGTM 💯 Thanks Luke!

src/qilisdk/optimizers/scipy_optimizer.py
```diff
@@ -51,7 +51,12 @@ def __init__(
                     - 'dogleg'
                     - 'trust-ncg'
                     - 'trust-exact
```
Lumorti:
```suggestion
                    - 'trust-exact'
```

PR #252, opened 2026-06-23
[SDK-319] Add GPU Support to Variational Annealing in QiliSim

Lumorti:
GPU support has been added to the Variational Annealing method in QiliSim. No extra dependencies have been added, it tries to load the CUDA libraries as they would be installed via the QiliSDK CUDA optional install, and defaults to CPU if that fails. Usage is via a flag in the `ExecutionConfig`:

```python
backend = QiliSim(
    analog_simulation_method=AnalogMethod.variational_annealing(), 
    execution_config=ExecutionConfig(gpu=True),
)
```

For a 60-qubit test case (an X to ZZ chain, second-order ansatz, defaults nshots), the time taken goes from 108s (8-core CPU), to 18s (8-core CPU + RTX 4060 GPU).


fedonman (approved):
Great feature @Lumorti! Thanks a lot for the effort.

PR #253, opened 2026-06-25
[SDK-328] Fix QUBO and TrotterizedSchedule import paths

fedonman:
- `QUBO` is now importable directly from `qilisdk.core` (`from qilisdk.core import QUBO`), alongside `Model`. It was previously only accessible via `qilisdk.core.model`.

- Resolved a circular import between `qilisdk.digital` and `qilisdk.utils.trotterization`. `trotter_evolution` and `_commuting_trotter_evolution` are now imported lazily inside the methods that use them (`TrotterizedSchedule` and `QAOA`), so importing `from qilisdk.digital import TrotterizedSchedule` works regardless of import order.


amirazzam2000 (approved):
Thanks Vyron! 🙌

PR #255, opened 2026-06-26
[SDK-315] Make QiliSim Faster

Lumorti:
QiliSim statevector performance has been improved for both digital and analog simulation:

 - For a 27 qubit random circuit of 1000 gates, sampling time is down from 44s -> 16s (**2.75x speedup**).
 - For a 25 qubit X to ZZ annealing task, time to reach 99% of the solution is down from 122s -> 79s (**1.54x speedup**).

This was done by analyzing the assembly via `perf`, then improving the memory accesses and ensuring that we correctly use SIMD instructions. We also implement gate-fusion, where nearby gates that act on similar sections of the statevector are grouped and applied together to improve memory access.

Performance can be pushed if the user compiles specific to their machine with the new compile flag
 `-Ccmake.define.build_native=ON`, which for newer CPUs can offer some slight benefit. If the user cares more about performance than precision, they can also compile using single-precision floats rather than doubles with the new flag `-Ccmake.define.single_precision=ON`. This is fine for sampling (since finite sampling is inherently noisy) but can give non-physical values for expectation values or state tomography. On my laptop with both flags the above random circuit takes 14s instead, so a ~10% speedup.


amirazzam2000 (commented):
Missing change log, and also maybe mention the flags in the docs (maybe in the QiliSim section or the building from source section).

A question here, if they install from Pypi shouldn't these flags already be on? specifically the building for the specific platform.

Lumorti:
No, we can't build for the specific processor for Pypi, we build once for general hardware then ship that binary. Compiling for my exact make of processor and then shipping that via Pypi would mean that it would fail to run on anyone who doesn't also have my exact processor

src/qilisdk/backends/backend_config.py
```diff
@@ -333,6 +333,17 @@ class DigitalMethod(BaseSimulatorConfig):
...
+            small set of qubits into a single dense multi-qubit operation. This
+            reduces the number of passes over the statevector (the memory
+            bandwidth bottleneck for large qubit counts) at the cost of extra
+            arithmetic per fused block. Supersedes ``combine_single_qubit_gates``
+            when enabled, and only applies to the matrix-free statevector path
+            (disabled automatically when a noise model is present). Defaults to
+            ``True``.
+        max_fused_qubits (int): Maximum number of qubits a single fused block may
+            span. Larger values fuse more aggressively but build larger dense
+            matrices (``2**max_fused_qubits`` square). Defaults to ``4``.
```
fedonman, line 346: These additions should be explained in a separate changelog. Current changelog only mentions build flags.

CMakeLists.txt
```diff
@@ -1,15 +1,29 @@
...
 #   uv pip install -v -e ./ -Ccmake.define.tidy=ON
+
 # To enable (very) verbose simulator output:
 #   uv pip install -v -e ./ -Ccmake.define.verbose=ON
+
 # To enable the test suite:
 #   uv pip install -v -e ./ -Ccmake.define.tests=ON
+
 # To enable coverage flags:
 #   uv pip install -v -e ./ -Ccmake.define.coverage=ON
```
fedonman: Shouldn't all these commands explained in the documentation as well?

fedonman (changes requested):
Thanks @Lumorti, great work! Left two minor comments.

CMakeLists.txt
fedonman: I would state what are the default flags, e.g. "Running `uv pip install qilisdk` is equivalent to `uv pip install qilisdk -Ccmake.define.coverage=OFF -Ccmake....`" and then list all the flags one by one like you do here.

Lumorti: yes, good idea, originally they weren't because we didn't have a "building from source" section, but now we do so I'll update that

src/qilisdk/backends/backend_config.py
Lumorti: true! I'll add a note about them, I didn't mention them because I did a lot of benchmarking and max_fuse_qubits=4 seems optimum and is what cirq etc. use, so I didn't want to encourage the user to tweak them, but I should mention them in both the changelog and docs

fedonman (approved):
LGTM 💯

PR #257, opened 2026-06-30
[SDK-329] Add check for changelog

Lumorti:
The `checks.sh` pre-commit check script now checks to make sure a changelog file is added in the branch diff, assuming the branch has a non-hotfix name like `sdk-XXX-description`.


fedonman (approved):
Nice! 💯

PR #262, opened 2026-07-06
[SDK-344] Fix Default Gate Time

Lumorti:
Fixed a bug in which the default gate time was not being set correctly in QiliSim, as well as improving how per-gate noise is handled.


fedonman (approved):
Thank you for fixing this! 💯

PR #263, opened 2026-07-07
[SDK-345] Reservoir Input Label Warning

Lumorti:
A warning has been added if a label is given to reservoir that it doesn't recognize.


src/qilisdk_cpp/backends/qilisim/analog/iterations_integrate.cpp
Lumorti, line 1: You might have noticed this line keeps being reformatted by clang format in various PRs, for now I'm just turning off formatting for this line, I think it's a difference between different clang-format versions on my home laptop versus work one, but for now this is a quick fix

fedonman (approved):
Thank you Luke!

PR #264, opened 2026-07-07
[SDK-330] Allow Initial State for Circuits

Lumorti:
The initial state for a `DigitalPropagaton` functional can now be set, just as in the case of the `AnalogEvolution` functional. It can be set to either a `Circuit`, `InitialState` (i.e. `InitialState.ZERO`, `InitialState.ONE` or `InitialState.UNIFORM`), or a `QTensor`. Note that the `QTensor` input will not be supported on real hardware. Usage:

```python
from qilisdk.digital import Circuit

c = Circuit(5)
digital = DigitalPropagation(circuit=c, initial_state=InitialState.ONE)
backend = QiliSim()

results = backend.execute(digital, readout)
print(results.get_samples())
```


fedonman (approved):
Smart implementation! Thanks Luke! 💯

PR #266, opened 2026-07-08
[SDK-342] Improve QTensor Performance

Lumorti:
Previously `QTensor` used a row-sparse matrix format, which was better some quantum objects (e.g. gates) but not so good for things like statevectors. Now, the C++ side of `QTensor` detects the best sparsity pattern based on the initialization, allowing faster initialization and manipulation. The before and after, comparing versus `numpy` for some 25-qubit state initializatons:

```
# Task    |  Numpy  | QTensor Before | QTensor After
# --------|---------|----------------|----------------
# zero    | 0.00003 | 0.37700        | 0.00001
# uniform | 0.09614 | 1.19820        | 0.08174
# ket     | 0.00002 | 0.38340        | 0.00001
# bra     | 0.00001 | 0.00002        | 0.00000
```

---
_Recreated from #265 — the original PR was auto-closed as "merged" during a git mishap; the branch and its contents are unchanged._


amirazzam2000 (approved):
Thanks Luke!

PR #267, opened 2026-07-08
[SDK-258] Add Time-dependent Lindblad Noise

Lumorti:
Support for time-dependent Lindblad noise has been added, for now only supported in QiliSim. Usage is as follows:

```python
from qilisdk.noise import LindbladGenerator
import numpy as np
J1 = QTensor(np.array([[0, 1], [0, 0]]))
T = 100
lindblad_noise = LindbladGenerator(jump_operators=[J1], rates=[lambda t: 0.005 * (1 + np.sin(t / T))])
```

By extension, `QutipBackend` now also supports general noise for analog evolution (but for now still not for digital).

---
_Recreated from #245 — the original PR was auto-closed as "merged" during a git mishap; the branch and its contents are unchanged._


amirazzam2000 (approved):
You can merge this after you resolve the conflicts. Thanks!! :raised_hands:

PR #269, opened 2026-07-08
[SDK-330] Allow Initial State for Circuits

### LINKED ISSUE #346: [Feature]: Let DigitalPropagation take an initial state
### Feature Summary

Add an `initial_state` argument to `DigitalPropagation` so a circuit can run from an arbitrary ket, density matrix or `InitialState`.

### Motivation / Use Case

`AnalogEvolution` and `QuantumReservoir` both take an initial state. `DigitalPropagation` takes only `circuit`, so you cannot run a circuit on a prepared state, a mixture, or the output of a previous run. QiliSim already supports this: `QiliSim._execute_digital_propagation` accepts `initial_state` and forwards it to C++, but the handler lambda in `Backend` passes only `(functional, readout)`, so nothing public can reach it.

### Proposed Solution

```python
psi0 = (ket(0) + ket(1)).unit()
QiliSim().execute(DigitalPropagation(circuit, initial_state=psi0), Readout().with_sampling(nshots=1000))
```

Add `initial_state: QTensor | InitialState = InitialState.ZERO` to `DigitalPropagation.__init__`, expose it as a property, include it in the yaml and repr round-trip, and validate its qubit count against `circuit.nqubits`. Then drop the extra parameter from `QiliSim._execute_digital_propagation` and read `functional.initial_state` instead.

`QutipBackend` and `CudaBackend` hardcode the all-zero state and need updating too, or they should raise on a non-zero state rather than silently starting from zero.

Lumorti:
The initial state for a `DigitalPropagaton` functional can now be set, just as in the case of the `AnalogEvolution` functional. It can be set to either an `InitialState` (i.e. `InitialState.ZERO`, `InitialState.ONE` or `InitialState.UNIFORM`), or a `QTensor`. Note that the `QTensor` input will not be supported on real hardware. Usage:

```python
from qilisdk.digital import Circuit

c = Circuit(5)
digital = DigitalPropagation(circuit=c, initial_state=InitialState.ONE)
backend = QiliSim()

results = backend.execute(digital, readout)
print(results.get_samples())
```


src/qilisdk/backends/cudaq_backend.py
```diff
@@ -263,13 +289,17 @@ def _execute_digital_propagation(
                 f"Only Sampling Readouts are supported for {self.sampling_method.value.upper()} simulation."
             )
 
+        # If an initial state is given, convert it to a circuit, since CUDA doesn't support any other initial state format
+        circuit = self._circuit_from_initial_state(functional.initial_state, functional.circuit.nqubits)
```
amirazzam2000, line 346: when this pr is approved, there has to be PRs in speqtrum/qili runtime so that this behavior is propagated down without issues.

amirazzam2000 (commented):
I will say we should put this PR on hold until we can open PRs in the Runtime to make sure that the execution can be maintained across the stack. But wait till we update the runtimes with the latest changes first before making the PRs.

src/qilisdk/functionals/digital_propagation.py
```diff
@@ -47,13 +48,26 @@ class DigitalPropagation(PrimitiveFunctional):
 
     result_type: ClassVar[type[FunctionalResult]] = FunctionalResult
 
-    def __init__(self, circuit: Circuit) -> None:
+    def __init__(
+        self,
+        circuit: Circuit,
+        initial_state: QTensor | InitialState | Circuit = InitialState.ZERO,
```
fedonman: I understand why we added `Circuit` here. It is feasible because, in most cases, the state is converted into a circuit anyway, as in `CudaBackend._circuit_from_initial_state`. 

However, from a public API and contract perspective, I do not like this approach.

First, it mixes up two distinct concepts: a state and a circuit.

Second, the same result can already be achieved using `Circuit.prepend/append`, giving users two different ways to express the same operation. A clean API should provide one clear and obvious way to perform an operation.

Third, it's not unified with `AnalogEvolution.initial_state` or `QuantumReservoir.initial_state`.

The cleanest solution would be to remove `Circuit` from the accepted types here.

PS. Moreover, from a theoretical perspective, allowing it introduces a recursive question: What is the initial state of the given circuit? How can I change it? Why allow one circuit and not a list of circuits? And so on...

fedonman (changes requested):
Thank you Luke! I left a single (but important) comment.

src/qilisdk/functionals/digital_propagation.py
Lumorti: I agree with this logic, I will remove it. I think we only included it because it was easy to do, but you're correct that it's already easy with prepend anyway

fedonman (approved):
LGTM 👍  Thanks Luke. One question, shouldn't we update the related documentation? Or is handled in a different PR? What's the policy on this?

Lumorti:
Good catch, I forgot to update the docs here, will do that now

Lumorti:
Added this to `speqtrum.py`, because I don't think we're going to support state initialization on hardware for a while, so with this we don't need to do any PR for runtime:

```python
if functional.initial_state is not InitialState.ZERO:
  logger.warning(
      "[SpeQtrum] DigitalPropagation was submitted with a custom initial state, which is not supported "
      "on hardware and will be ignored. The circuit will be executed starting from the zero state."
  )
```

fedonman (approved):
Thanks Luke!

Lumorti:
Note: we agreed at some point that the check for initial state being ignored should be at the runtime level rather than at the SDK level, so that should be removed from this PR and added into runtime

PR #270, opened 2026-07-08
SDK[339]: surface maintenance messages from PublicAPI

Lumorti:
Updated SpeQtrum job submission to display maintenance messages from PublicAPI and fixed the execution type when submitting a quantum reservoir payload.


DZPM (approved):
LGTM 👍

src/qilisdk/speqtrum/speqtrum.py
amirazzam2000: We should probably change this as well

amirazzam2000 (changes requested):
Thanks!! And good catch on the reservoir!!

NataliaSirera (approved):

amirazzam2000 (approved):
Thanks Fiona!! 🫶

PR #259, opened 2026-07-02
[SDK-338] Added New Classical Solvers

Lumorti:
Two new classical solvers have been added: ScipySolver and ScipSolver. SCIP is one of the leading mixed-integer programming solvers which doesn't need a license (compared to something like Gurobi). It can be installed via an optional extra:

```bash
uv pip install qilisdk[scip]
```

Usage is the same as with the BruteForceSolver:
```python3
from qilisdk.core import Model
from qilisdk.utils.classical_solvers import BruteForceSolver, ScipySolver, ScipSolver
model = Model.random_ising(10)
# solver = BruteForceSolver()
# solver = ScipSolver()
solver = ScipySolver(method="nelder-mead")
solution = solver.solve(model)
print(solution)
```
For a random Ising model with 13 qubits:
```
Solver                               Time           Solution
------------------------------------------------------------------
Brute force                         7.4171 s        -5.1688
Scipy (Nelder-Mead)                 0.6702 s        -4.4559
Scipy (Powell)                      0.7337 s        -4.4559
Scipy (TNC)                         0.1107 s        -2.8488
Scipy (COBYLA)                      0.1357 s        -4.4323
Scipy (differential_evolution)      4.3912 s        -5.1688
Scipy (basinhopping)                1.4425 s        -3.3048
SCIP                                0.4231 s        -5.1688
QiliSim (statevector)               1.4903 s        -5.0092
QiliSim (VA, order=1)               1.0353 s        -4.5232
QiliSim (VA, order=2)               2.3634 s        -5.1386
```

For a random Ising model with 30 qubits:
```
Solver                               Time          Solution
------------------------------------------------------------------
SCIP                                2.3561 s       -28.1468
QiliSim (VA, order=1)               3.4792 s       -28.0689
QiliSim (VA, order=2)              32.7094 s       -28.1075
```

PR #271, opened 2026-07-08
[SDK-338] Added New Classical Solvers

Lumorti:
Two new classical solvers have been added: ScipySolver and ScipSolver. SCIP is one of the leading mixed-integer programming solvers which doesn't need a license (compared to something like Gurobi). It can be installed via an optional extra:

```bash
uv pip install qilisdk[scip]
```

Usage is the same as with the BruteForceSolver:
```python3
from qilisdk.core import Model
from qilisdk.utils.classical_solvers import BruteForceSolver, ScipySolver, ScipSolver
model = Model.random_ising(10)
# solver = BruteForceSolver()
# solver = ScipSolver()
solver = ScipySolver(method="nelder-mead")
solution = solver.solve(model)
print(solution)
```
For a random Ising model with 13 qubits:
```
Solver                               Time           Solution
------------------------------------------------------------------
Brute force                         7.4171 s        -5.1688
Scipy (Nelder-Mead)                 0.6702 s        -4.4559
Scipy (Powell)                      0.7337 s        -4.4559
Scipy (TNC)                         0.1107 s        -2.8488
Scipy (COBYLA)                      0.1357 s        -4.4323
Scipy (differential_evolution)      4.3912 s        -5.1688
Scipy (basinhopping)                1.4425 s        -3.3048
SCIP                                0.4231 s        -5.1688
QiliSim (statevector)               1.4903 s        -5.0092
QiliSim (VA, order=1)               1.0353 s        -4.5232
QiliSim (VA, order=2)               2.3634 s        -5.1386
```

For a random Ising model with 30 qubits:
```
Solver                               Time          Solution
------------------------------------------------------------------
SCIP                                2.3561 s       -28.1468
QiliSim (VA, order=1)               3.4792 s       -28.0689
QiliSim (VA, order=2)              32.7094 s       -28.1075
```

---
_Recreated from #259 — the original PR was auto-closed as "merged" during a git mishap; the branch and its contents are unchanged._


Lumorti:
Previous comments from Vyron:

> Should the module be in utils? Is it a utility? Or is it integral to QiliSDK and we should move it to core? E.g. under solvers module. (classical seems reductant to me)

I originally called it classical to note the fact that it's the way of solving the problem classically, like with no quantum at all, compared  to the rest of the library which (even in the case of hyrid methods like VQEs that use ScipyOptimizer) involve something quantum.

> Question: What is the difference between solvers and optimizers? E.g. What's the difference between ScipySolver and ScipyOptimizer?

I agree the naming similarity is confusing, the intent is that the Optimizers are an object that take a cost function and some parameters and then they return you a new set of parameters trying to minimize the cost function, they're what we plug into the VQE to do the classical minimization. Meanwhile the Solvers are objects where you give them a Model and they do something return the optimal solution, they're self-contained and do the full solve, whilst Optimizers are just one component of a bigger algorithm. But that's not like a fixed naming scheme, happy to change it if you have any suggestions.

> I am not so sure if we should make this optional or part of the core.

I just figured I would add it as an optional to avoid new core depends, since we have the ScipySolver and BruteForceSolver in the core already, in the same way that openqasm etc. was made optional

amirazzam2000 (approved):
Amazing!! Let's see how we compare against these solvers 💪

PR #272, opened 2026-07-08
[SDK-315] Make QiliSim Faster

Lumorti:
QiliSim statevector performance has been improved for both digital and analog simulation:

 - For a 27 qubit random circuit of 1000 gates, sampling time is down from 44s -> 16s (**2.75x speedup**).
 - For a 25 qubit X to ZZ annealing task, time to reach 99% of the solution is down from 122s -> 79s (**1.54x speedup**).

This was done by analyzing the assembly via `perf`, then improving the memory accesses and ensuring that we correctly use SIMD instructions. We also implement gate-fusion, where nearby gates that act on similar sections of the statevector are grouped and applied together to improve memory access.

Performance can be pushed if the user compiles specific to their machine with the new compile flag
 `-Ccmake.define.build_native=ON`, which for newer CPUs can offer some slight benefit. If the user cares more about performance than precision, they can also compile using single-precision floats rather than doubles with the new flag `-Ccmake.define.single_precision=ON`. This is fine for sampling (since finite sampling is inherently noisy) but can give non-physical values for expectation values or state tomography. On my laptop with both flags the above random circuit takes 14s instead, so a ~10% speedup.

---
_Recreated from #255 — the original PR was auto-closed as "merged" during a git mishap; the branch and its contents are unchanged._


amirazzam2000 (approved):
Thanks luke! 🚀

PR #273, opened 2026-07-08
[SDK-319] Add GPU Support to Variational Annealing in QiliSim

Lumorti:
GPU support has been added to the Variational Annealing method in QiliSim. No extra dependencies have been added, it tries to load the CUDA libraries as they would be installed via the QiliSDK CUDA optional install, and defaults to CPU if that fails. Usage is via a flag in the `ExecutionConfig`:

```python
backend = QiliSim(
    analog_simulation_method=AnalogMethod.variational_annealing(), 
    execution_config=ExecutionConfig(gpu=True),
)
```

For a 60-qubit test case (an X to ZZ chain, second-order ansatz, defaults nshots), the time taken goes from 108s (8-core CPU), to 18s (8-core CPU + RTX 4060 GPU).

---
_Recreated from #252 — the original PR was auto-closed as "merged" during a git mishap; the branch and its contents are unchanged._


src/qilisdk/backends/backend_config.py
```diff
@@ -282,6 +282,12 @@ class ExecutionConfig(BaseSimulatorConfig):
             "Whether to apply state collapse immediately after measurements. If `False`, measurements are recorded but the state is not collapsed."
         ),
     )
+    gpu: bool = Field(
```
amirazzam2000, line 288: maybe list somewhere the methods that support GPU in QiliSim

amirazzam2000 (approved):
I approved it but please take a look at my comment. I think it's important to have this table somewhere and preferably accessible through method docs and the qilisdk docs, so that the user can find it easily cuz as of now this is only supported for variational annealing and it should be very clear.

src/qilisdk/backends/backend_config.py
Lumorti: good catch

PR #276, opened 2026-07-10
[SDK-337] Add Return Execution Time

Lumorti:
When executing a job, there is now a new field that can be returned: the time taken for execution. For now this is only used for local simulation jobs, but later support may be added for real quantum hardware execution time. Example:

```python

c = Circuit(5)
c.add(H(0))
c.add(CNOT(0, 1))
func = DigitalPropagation(c)
readout = Readout().with_sampling(nshots=100)
backend = QiliSim()
res = backend.execute(func, readout)
print(res)
```


amirazzam2000 (approved):
Nice Addition 👍

PR #278, opened 2026-07-13
[SDK-333] Add More Logging

Lumorti:
More logging has been added to QiliSDK, using the loguru library as before. Different logging levels are now used throughout the library (TRACE, DEBUG, INFO, SUCCESS, WARNING, ERROR, CRITICAL) and can be enabled either via changed the `logging_config.yaml` or via the configure function as follows:

```python3
from qilisdk.logging import configure_logging
configure_logging(level="DEBUG")
```


Lumorti:
image

(an example of how the formatting looks)

src/qilisdk_cpp/libs/logging.h
```diff
@@ -0,0 +1,34 @@
...
+// Thin wrappers around the globally-available Python loguru `logger` (see pybind.h).
+//
+// IMPORTANT: these call back into Python and therefore require the GIL to be held.
+// Only call them from the master thread and NEVER from inside an OpenMP parallel
+// region, otherwise concurrent Python calls from worker threads will crash.
+//
+// The message string is built eagerly at the call site regardless of the active log
+// level, so keep these calls coarse (per-execution / per-phase) and out of hot loops.
+namespace qilisdk {
+
```
fedonman, line 27:
```suggestion
enum class LogLevel : int {
    TRACE = 5,
    DEBUG = 10,
    INFO = 20,
    SUCCESS = 25,
    WARNING = 30,
    ERROR = 40,
    CRITICAL = 50,
};

void set_log_threshold(int severity);   // Python sets this when logging is configured
bool log_enabled(LogLevel level);       // guard hot-loop log calls with this
```

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -263,6 +271,8 @@ void sampling(const std::vector<Gate>& gates, int n_qubits, const SparseMatrixCo
             gate_matrix = gate.get_full_matrix(n_qubits);
         }
 
+        qilisdk::log_trace("[Sampling, C++] Applying gate " + gate.get_id() + (gate_is_new ? " (cache miss)" : " (cache hit)"));
```
fedonman, line 274: This is in the per-gate loop, so we build the string and call into Python for every gate, even at DEBUG or higher, where it's dropped. We should guard it with the `log_enabled` (check logging.h/cpp)


```suggestion
          if (qilisdk::log_enabled(qilisdk::LogLevel::TRACE)) {
    qilisdk::log_trace("[Sampling, C++] Applying gate " + gate.get_id() + (gate_is_new ? " (cache miss)" : " (cache hit)"));
}
```

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -234,6 +240,8 @@ void sampling(const std::vector<Gate>& gates, int n_qubits, const SparseMatrixCo
                 }
             }
 
+            qilisdk::log_trace("[Sampling, C++] Measuring " + std::to_string(num_measurements) + " qubits at gate index " + std::to_string(i));
```
fedonman, line 243: Same here.
```suggestion
            if (qilisdk::log_enabled(qilisdk::LogLevel::TRACE)) {
    qilisdk::log_trace("[Sampling, C++] Measuring " + std::to_string(num_measurements) + " qubits at gate index " + std::to_string(i));
}
```

src/qilisdk_cpp/backends/qilisim/digital/sampling.cpp
```diff
@@ -475,6 +495,8 @@ void sampling_matrix_free(const std::vector<Gate>& gates, int n_qubits, const Sp
     if (!initially_was_statevector && is_statevector) {
         state = state * state.adjoint();
     }
+
+    qilisdk::log_debug("[Sampling, C++] Applied " + std::to_string(optimized_gates.size()) + " gates, matrix-free circuit sampling complete");
 }
 
 #include <iostream>
```
fedonman, line 502: Same here.
```suggestion
        if (qilisdk::log_enabled(qilisdk::LogLevel::TRACE)) {
    qilisdk::log_trace("[Sampling, C++] Measuring " + std::to_string(num_measurements) + " qubits at gate index " + std::to_string(i));
}
```

src/qilisdk_cpp/libs/logging.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "pybind.h"
+
+// GCOV_EXCL_BR_START
+
+#pragma GCC visibility push(default)
+
+namespace qilisdk {
+
+namespace {
+// Forward a message to the given loguru level. Safe to call before the logger has
```
fedonman: Sample implementation.
```suggestion
#include <atomic>

namespace {
std::atomic<int> g_log_threshold{static_cast<int>(LogLevel::WARNING)};
}

void set_log_threshold(int severity) {
    g_log_threshold.store(severity, std::memory_order_relaxed);
}

bool log_enabled(LogLevel level) {
    return static_cast<int>(level) >= g_log_threshold.load(std::memory_order_relaxed);
}
```

We should expose this at `qilisim_bindings.cpp` as well, like `m.def("set_log_threshold", &qilisdk::set_log_threshold);
`.

src/qilisdk/backends/backend.py
```diff
@@ -121,6 +122,7 @@ def execute(self, functional: Functional, readout: Readout[Any, Any, Any]) -> Re
         readout_list = readout.to_list()
         if not readout_list:
             raise ValueError("At least one readout method must be provided in the Readout.")
+        logger.debug("[Backend] Readout methods: {}", [type(ro).__name__ for ro in readout_list])
```
fedonman: loguru evaluates the args eagerly, so this list gets built on every dispatch even with DEBUG off. Make it lazy:
```suggestion
        logger.opt(lazy=True).debug("[Backend] Readout methods: {}", lambda: [type(ro).__name__ for ro in readout_list])
```

Do the same on other places that we do comprehensions etc.

src/qilisdk/logging.py
```diff
@@ -0,0 +1,210 @@
...
+    if filename is not None:
+        file_params: dict[str, Any] = {
+            "level": level if level is not None else base_level,
+            "colorize": False,
+        }
+        if base_format is not None:
+            file_params["format"] = base_format
+        if base_filter is not None:
+            file_params["filter"] = base_filter
+        logger.add(str(Path(filename).expanduser()), **file_params)
```
fedonman, line 208:
```suggestion
# Tell the C++ layer the lowest level we'll actually emit.
try:
    from qilisdk_cpp.qilisim_module import set_log_threshold  # fix import path

    if level is not None:
        effective = level.upper()
    else:
        sink_levels = [s.level for s in settings.sinks] or ["WARNING"]
        effective = min(sink_levels, key=lambda name: logger.level(name).no)
    set_log_threshold(logger.level(effective).no)
except ImportError:
    pass  # C++ ext not built       
    logger.add(str(Path(filename).expanduser()), **file_params)
```

src/qilisdk/core/parameterizable.py
```diff
@@ -246,6 +250,7 @@ def set_parameter_values(
         Raises:
             ValueError: If ``values`` does not match the number of parameters selected by ``where``.
         """
+        logger.trace("[Parameterizable] Setting {} parameter values", len(values))
```
fedonman: If someone turns on TRACE they probably want the actual values, not just the count:

```suggestion
        logger.opt(lazy=True).trace("[Parameterizable] Setting parameter values {}", lambda: values)
```

src/qilisdk/core/parameterizable.py
```diff
@@ -262,6 +267,7 @@ def set_parameters(self, parameters: dict[str, float]) -> None:
         Raises:
             ValueError: If an unknown parameter label is provided or constraints are violated.
         """
+        logger.trace("[Parameterizable] Setting {} parameters by label", len(parameters))
```
fedonman: same here.

src/qilisdk/core/parameterizable.py
```diff
@@ -293,6 +299,7 @@ def set_parameter_bounds(self, ranges: dict[str, tuple[float, float]]) -> None:
         Raises:
             ValueError: If an unknown parameter label is provided.
         """
+        logger.trace("[Parameterizable] Setting bounds for {} parameters", len(ranges))
```
fedonman: Same here.

src/qilisdk/core/interpolator.py
```diff
@@ -401,6 +408,7 @@ def add_time_point(
         Raises:
             ValueError: If the coefficient type is unsupported or the callable uses invalid variables.
         """
+        logger.trace("[Interpolator] Adding time point at t={}", time)
```
fedonman: Same here.

src/qilisdk/core/interpolator.py
```diff
@@ -440,6 +449,7 @@ def set_parameters(self, parameters: dict[str, int | float]) -> None:
         Args:
             parameters (dict[str, int | float]): Mapping from parameter labels to numeric values.
         """
+        logger.trace("[Interpolator] Setting {} parameters by label", len(parameters))
```
fedonman: Same here

src/qilisdk/core/interpolator.py
```diff
@@ -450,6 +460,7 @@ def set_parameter_bounds(self, ranges: dict[str, tuple[float, float]]) -> None:
         Args:
             ranges (dict[str, tuple[float, float]]): Bounds keyed by parameter label.
         """
+        logger.trace("[Interpolator] Setting bounds for {} parameters", len(ranges))
```
fedonman: same here.

fedonman (changes requested):
Thanks for adding this! Implementation is great, I left some comments, but they basically come down to three things:

- The C++ traces in the per-gate loop build the string and call into Python on every gate, even at higher log levels where it just gets dropped.
- A couple of Python logs build their payload eagerly (loguru doesn't defer arg evaluation), so a lazy lambda would avoid the work when the level's off.
- TRACE only logs counts/types right now, not the actual objects. DEBUG should be used for summaries (length of list), TRACE for actual objects (elements of list)

fedonman (approved):
Thanks! 💯 💯

PR #280, opened 2026-07-15
[SDK-350] Disable All State Normalization

Lumorti:
The normalization of the state in all simulation methods in QiliSim can now be disabled:
```python
from qilisdk.backends import QiliSim, ExecutionConfig
backend = QiliSim(execution_config=ExecutionConfig(normalize_state=False))
```


amirazzam2000 (changes requested):
Thanks Luke! Two things: 
1- Why don't we stop the computation once the state diverges, I did some test and I got many state diverging warning and then a none state at the end. we should probably just stop and throw an error the minute the state diverges to avoid using resources on a diverged state. 
2- If normalize_state is True and the initial state is not a valid density matrix then we should try to normalize it before starting and if we fail we should raise an error. Because I just tested it and it only gave me a diverge warning but it allowed the evolution with a non-valid density matrix.  

But the functionality works I tested it! thanks again 🫶

amirazzam2000 (approved):
Thanks Luke!! :heart_hands:

PR #282, opened 2026-07-16
[SDK-354] Fix Analog Global Noise

Lumorti:
A bug has been fixed in which global jump operator noise was expanded as L (x) L rather than two separate jump operators L (x) I and I (x) L, which resulted in global analog noise being applied inconsistently.


amirazzam2000 (approved):
Thanks! 🫶

PR #283, opened 2026-07-16
[SDK-331] Make Qilisim Even Faster

Lumorti:
QiliSim performance has been generally increased via a variety of micro-optimizations and improvements to gate fusion. The Arnoldi step now does matrix-free operations by default, increasing performance, although the integrators are still faster. Reservoirs have benefited especially from these performance improvements.

Approximate performance improvements:
- digital statevector simulation: ~1.5x
- analog integration: ~1.2x
- reservoirs: ~10x


amirazzam2000 (approved):
Thank you Luke! 🚀

PR #284, opened 2026-07-20
[SDK-351] Add ML Datasets

Lumorti:
Some generators of common machine-learning datasets have been added to QiliSDK. These are useful when working with quantum reservoirs. They all have a generate method that allows generating samples as follows:

```python
from qilisdk.ml.datasets import MackeyGlass
inputs, targets = MackeyGlass().generate(100)
```

You can also draw the datasets:
```python
MackeyGlass.draw(inputs, style="2d")
```

An example of this use with a quantum reservoir is as follows:
```python
import numpy as np

from qilisdk.analog import Schedule, X, Y, Z
from qilisdk.backends import AnalogMethod, QiliSim
from qilisdk.digital import RY, Circuit
from qilisdk.ml.datasets import NARMA

NQUBITS = 5
NSTEPS = 500  # one reservoir layer per time step
WASHOUT = 50  # steps dropped while the reservoir forgets its initial state
NTRAIN = 300  # training steps; the remaining steps are the test set

# The NARMA task: a random drive u(t) and the nonlinear response y(t) to predict.
inputs, targets = NARMA(order=2, seed=42).generate(NSTEPS)
u, y = inputs[:, 0], targets[:, 0]

# The Hamiltonian to use
J = 1
hx = 0.5
hz = 1.05
hamiltonian = (
    J * sum(X(i) * X(i + 1) for i in range(NQUBITS - 1))
    + hz * sum(Z(i) for i in range(NQUBITS))
    + hx * sum(X(i) for i in range(NQUBITS))
)

# Each layer encodes one input sample as an RY rotation on qubit 0, evolves the whole
# register, and then resets qubit 0. That reset is what gives the reservoir a fading
# memory: information injected at step t leaks into the other qubits and slowly decays.
input_encoding = Circuit(NQUBITS)
input_encoding.add(RY(0, theta=ReservoirInput("u", 0.0)))

# Set up the reservoir
layer = ReservoirLayer(
    evolution_dynamics=Schedule.constant(hamiltonian, total_time=5.0, dt=0.5),
    input_encoding=input_encoding,
    qubits_to_reset=[0],
)
reservoir = QuantumReservoir(
    initial_state=QTensor.zero(NQUBITS),
    reservoir_layer=layer,
    input_per_layer=[{"u": float(2 * np.arcsin(np.sqrt(sample / 0.5)))} for sample in u],
)

# The expectation values measured after each layer are the features of the readout
observables = [op(q) for op in (X, Y, Z) for q in range(NQUBITS)]
observables += [(Z(i) * Z(j)) for i in range(NQUBITS) for j in range(i + 1, NQUBITS)]

# Set up the simulation
backend = QiliSim(analog_simulation_method=AnalogMethod.direct())
result = backend.execute(reservoir, Readout().with_expectation(observables=observables))

# The intermediate readouts cover layers 0 to L-2 and the final one is layer L-1, so
# together they give exactly one feature row per input sample.
features = np.array([*result.get_intermediate_expectation_values(), result.get_expectation_values()])

# Only a linear readout is trained (ridge regression with a bias column)
design = np.hstack([features[WASHOUT:], np.ones((NSTEPS - WASHOUT, 1))])
target = y[WASHOUT:]
x_train, y_train = design[:NTRAIN], target[:NTRAIN]
x_test, y_test = design[NTRAIN:], target[NTRAIN:]
penalty = 1e-8 * np.eye(x_train.shape[1])
penalty[-1, -1] = 0.0  # never penalize the bias
weights = np.linalg.solve(x_train.T @ x_train + penalty, x_train.T @ y_train)

# Score the prediction with the normalized mean squared error.
predictions = x_test @ weights
nmse = np.mean((y_test - predictions) ** 2) / np.var(y_test)
print(f"test NMSE: {nmse:.4f}")

import matplotlib.pyplot as plt
plt.plot(y_test, label="y_test")
plt.plot(predictions, label="prediction", linestyle="--")
plt.legend()
plt.show()
```


src/qilisdk/ml/datasets/henon.py
```diff
@@ -0,0 +1,89 @@
...
+        needed = npoints + self._horizon
+        total = self._washout + needed
+        traj = np.empty((total, 2), dtype=np.float64)
+        traj[0] = np.asarray(self._initial_state, dtype=np.float64)
+        for i in range(total - 1):
+            x, y = traj[i]
+            traj[i + 1] = (1.0 - self._a * x * x + y, self._b * x)
+
+        sampled = traj[self._washout :]
+        return build_prediction_sample(sampled, self._horizon)
```
amirazzam2000, line 98: The returned data samples need to be transposed in order to be able to have the x as input[0] and y as input[1].

src/qilisdk/ml/datasets/lorenz.py
```diff
@@ -0,0 +1,140 @@
...
+        traj = integrate_lorenz(
+            sigma=self._sigma,
+            rho=self._rho,
+            beta=self._beta,
+            initial_state=self._initial_state,
+            dt=self._dt,
+            n_steps=n_steps,
+        )
+        sampled = traj[self._washout :: self._sample_every][:needed]
+        return build_prediction_sample(sampled, self._horizon)
```
amirazzam2000, line 162: same here it need transposing

src/qilisdk/ml/datasets/logistic_map.py
```diff
@@ -0,0 +1,90 @@
...
+    Logistic map, a simple one-dimensional chaotic system:
+
+    .. math::
+
+        x_{n+1} = r\, x_n (1 - x_n),
+
+    which becomes chaotic as the growth rate :math:`r` approaches 4 (the
+    default :math:`r = 3.9` sits well inside the chaotic regime).
+    :meth:`generate` returns a ``horizon``-step-ahead prediction task, so
+    ``inputs`` and ``targets`` are shaped ``(npoints, 1)``.
```
amirazzam2000, line 32: the shape of the target should also be mentioned in the generate sample docstring for ease of access in all of them.

src/qilisdk/ml/datasets/dataset.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Raises:
+        ValueError: If ``horizon`` is not positive.
+    """
+    if horizon < 1:
+        raise ValueError(f"horizon must be a positive integer, got {horizon}.")
+    return DatasetSample(inputs=series[:-horizon], targets=series[horizon:])
+
+
+class Dataset(ABC):
```
amirazzam2000, line 119: we should add a plot method to this class because it would be nice to visualize the datasets how they are supposed to be visualized, which is a bit different per dataset: 

for example plotting Lorenz:

```
fig = plt.figure()
ax = plt.axes(projection="3d")

ax.plot3D(inputs.T[0], inputs.T[1], inputs.T[2])
```

some of the other 2d distributions also have special way of being visualized.

amirazzam2000: maybe there could be two overrides of the plot method one that accepts a DataSample and one that accepts a numpy array with the correct shape.

amirazzam2000 (changes requested):
Thanks a lot Luke!! this is a very useful addition. I left some comments but there is no rush for this 🙏

changes/284.feature.md
```diff
@@ -0,0 +1,6 @@
+Some generators of common machine-learning datasets have been added to QiliSDK. These are useful when working with quantum reservoirs. They all have a generate method that allows generating samples as follows:
+
+```python
+from qilisdk.ml.datasets import MackeyGlass
+inputs, targets = MackeyGlass().generate(100)
+```
```
fedonman, line 7: Can you provide some examples on how these datasets are used with reservoirs? In changelog as well as in documentation. From the POV of a reviewer or external user, this PR adds a module that is isolated from the rest of the codebase.

src/qilisdk/ml/datasets/dataset.py
```diff
@@ -0,0 +1,149 @@
...
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import TYPE_CHECKING, Callable, Iterator, TypeAlias, TypeVar, cast
+
+import numpy as np
+
+if TYPE_CHECKING:
+    from numpy.typing import NDArray
+
+FloatArray: TypeAlias = "NDArray[np.float64]"
```
fedonman: Shouldn't this follow our global setting for precision?

fedonman (changes requested):
Thanks Luke! Left two comments.

src/qilisdk/ml/datasets/dataset.py
Lumorti: Agree that plotting would be nice, you mean like the following?

```
Dataset.draw(DatasetSamples OR numpy array)
```

or would you prefer it to be DatasetSamples.draw() or a seperate DatasetRenderer class?

changes/284.feature.md
Lumorti: Good idea

src/qilisdk/ml/datasets/dataset.py
amirazzam2000: my concern with DatasetSamples.draw() is that we would need to do a different draw method per dataset but the DatasetSampels are a generic class, which is why I thought adding it to the Dataset would make more sense

src/qilisdk/ml/datasets/dataset.py
Lumorti: Done!

src/qilisdk/ml/datasets/logistic_map.py
Lumorti: Good idea

src/qilisdk/ml/datasets/henon.py
Lumorti: Since the inputs/targets are a list of x/y coordinates, doesn't it make more sense to have them be inputs[0] is the first pair, so inputs[0][0] is the first x and inputs[0][1] be the first y? As now stated in the generation the length of the arrays is (npoints, 2), which for me makes more sense than (2, npoints).

Lumorti: Is the opposite common in ML?

src/qilisdk/ml/datasets/mackey_glass.py
```diff
@@ -0,0 +1,160 @@
...
+from typing import TYPE_CHECKING, Any, ClassVar
+
+import numpy as np
+
+from qilisdk.ml.datasets.dataset import Dataset, DatasetSample, build_prediction_sample, rk4_step
+
+if TYPE_CHECKING:
+    from qilisdk.utils.visualization.dataset_renderers import Transform
+
+_ATTRACTOR_DELAY = 17
```
amirazzam2000: Why is this a fixed value and not an input of draw, it just feels like a missed opportunity. It can allow such a quick an easy way for plotting the data.

amirazzam2000 (approved):
I will approve it, but i still think that we should add an override to the draw method for some of the datasets, not all. But Mackey glass is a good example that i highlighted. 

and as always thanks a lot Luke for adding this :heart_hands:

src/qilisdk/ml/datasets/mackey_glass.py
Lumorti: Fair enough! I added an overload for mackey_glass, we can add more if we want specific non-transform changes for the different datasets

fedonman (approved):
Thanks Luke!

PR #285, opened 2026-07-20
[SDK-358] Add Qtensor Magic

Lumorti:
A new function has been added to QTensor: `QTensor.magic()`. This computes how "magic" a state is, correlating to how hard it is to simulate in the stabilizer state formalism. The method used is the efficient method used in this paper: https://arxiv.org/abs/2601.07824. Usage:

```python
state = QTensor.zero(3)
magic = state.magic()
print(magic)
```


src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -2132,6 +2132,120 @@ double QTensorCpp::entropy_renyi(double alpha) {
...
 
+static void _fast_walsh_hadamard_transform(std::vector<Complex>& a) {
+    /*
+    Apply the unnormalized fast Walsh-Hadamard transform (H_2)^{tensor N} to a vector in place.
+
+    The length of the vector must be a power of two. The transform is computed in O(N * 2^N) time using
+    the standard butterfly iteration, where H_2 = [[1, 1], [1, -1]].
+    */
+    const size_t n = a.size();
+    for (size_t h = 1; h < n; h <<= 1) {
```
amirazzam2000, line 2143: I see three nested loops with no open MP parallelization, is there a reason for that? cuz I see that the results don't need to be computed sequentially. Maybe I am missing something?

amirazzam2000 (approved):
Thanks Luke! 🪄 ✨

src/qilisdk_cpp/core/qtensor.cpp
Lumorti: I think they do need to be computed sequentially, no?

also, the parellelization happens outside of this function, at the higher level (in `magic()`), so this function is already being ran in parallel

PR #286, opened 2026-07-21
[GPU] Optimize sr solve

fedonman:


fedonman:
@Lumorti Did you have a chance to check this?

benchmarks/.gitignore
amirazzam2000, line 1: should we ad this benchmark in the qili-benchmark repo. I don't know if this should go in QiliSDK

fedonman: No we shouldn't, its for Luke to grab the profiling code.

amirazzam2000: But do you plan to leave this in QiliSDK forever or is it just for the purposes of sharing code?

src/qilisdk_cpp/libs/gpu.cpp
```diff
@@ -300,23 +303,76 @@ std::mutex& gpu_mutex() {
...
+    return pool;
+}
+
+// Minimum problem "work" (~ N_s·p², the Gram-matrix flop count) below which the
+// GPU path is declined in favour of the CPU. Small problems can't amortise the
+// GPU's fixed overhead (kernel launches, H2D/D2H, synchronisation) and, on a
+// consumer GPU, the 1/64-rate FP64 units, so Eigen on the CPU is faster there.
+// Default calibrated on a consumer GPU (RTX 3050 Laptop); override (e.g. lower
+// it on a datacenter GPU, or set 0 to disable gating) via QILISDK_GPU_MIN_WORK.
+// Re-read on every call (like QILISDK_DISABLE_GPU) so it can be toggled at
```
Lumorti, line 364: `QILISDK_DISABLE_GPU` doesn't seem to exist in this PR, did it add it and later remove it?

Probably there is no need for it, things should be CPU by default, GPU should be enabled by choice (e.g. via the QilSim(gpu=True)

src/qilisdk_cpp/libs/gpu.cpp
```diff
@@ -300,23 +303,76 @@ std::mutex& gpu_mutex() {
...
+// runtime, e.g. by tests that want to exercise the GPU path at small sizes.
+double gpu_min_work() {
+    if (const char* env = std::getenv("QILISDK_GPU_MIN_WORK")) {
+        const double v = std::atof(env);
+        if (v >= 0.0) {
+            return v;
+        }
+    }
+    return 1.5e8;
+}
```
Lumorti, line 374: The size gate is interesting, because in general with GPU stuff there isn't a size gate, e.g. if we use CudaBackend for a circuit of 3 qubit it's much slower than CPU, CUDA-Q doesn't default to CPU for small systems. I think this is by design, like it's nice to be able to see that GPU has an overhead for small systems and only excels for larger ones.

What I would maybe do at some point is have some sort of GPU=auto or simulation_method=auto setting that allows for this type of CPU/GPU gating and finds the optimum settings/hyperparameters to simulate as best as possible.

PR #287, opened 2026-07-21
[SDK-359] QiliSim Warn on Diverge

Lumorti:
Previously in analog simulation in QiliSim if the dt was set to high and the integrator diverged, it returned all zeros. It now raises a warning and returns quiet (i.e. not crashing anything) NaNs.

Also fixed a bug in which trying to use the adaptive integrator in a quantum reservoir wouldn't be detected correctly.


amirazzam2000 (approved):
Thank you luke for adding this, from my tests it seems like it's working 👍

PR #288, opened 2026-07-21
[SDK-361] make backends import lazy

amirazzam2000:
The optional simulation backends are now imported lazily: their heavy third-party dependencies (`cudaq` for `CudaBackend`/`CudaSamplingMethod`, `qutip` for `QutipBackend`) are only loaded the first time the backend is actually accessed, not when `qilisdk.backends` (or the default `QiliSim`) is imported. This keeps import times and memory usage low when you only use one backend.

```python
import sys

from qilisdk.backends import QiliSim  # neither cudaq nor qutip is imported

assert "cudaq" not in sys.modules
assert "qutip" not in sys.modules

# The optional dependency is imported only on first access of its backend:
from qilisdk.backends import QutipBackend  # now qutip is imported

assert "qutip" in sys.modules
```


fedonman (approved):
Looks good! Next python release will support this natively 😺 https://peps.python.org/pep-0810/

PR #292, opened 2026-07-23
Add Dependabot config for GitHub Actions

### LINKED ISSUE #293: [Feature]: Add Dependabot to keep GitHub Actions dependencies current
### Feature Summary

Add `.github/dependabot.yml` so Dependabot keeps all GitHub Actions dependencies up to date, including the SHA-pinned reference to the sharedreusable backport workflow (qilimanjaro-tech/workflows) and the action versions used across the other workflows. Bumps are grouped into a single weekly PR.

### Motivation / Use Case

The backport caller pins the reusable workflow to a full commit SHA (for supply-chain safety and to satisfy zizmor's `unpinned-uses`). A pinned SHA is immutable, so without automation it silently goes stale and never picks up fixes or action-version bumps made in the shared repo. Dependabot opens a PR whenever a newer release of the shared workflow exists, so we keep the security benefit of SHA-pinning without the manual chore of updating it.

### Proposed Solution

```python

```

### Alternatives Considered

_No response_

### Priority

Low - Nice to have

### Additional Context

Requires granting Dependabot read access to the private `qilimanjaro-tech/workflows` repo, and cutting incrementing version tags (`v1.0.0`, `v1.0.1`, ...) in that repo so Dependabot has releases to compare against.

fedonman:
Adds `.github/dependabot.yml` so Dependabot keeps the repository's GitHub Actions dependencies up to date, including the SHA-pinned reference to the shared reusable backport workflow. Updates are grouped into a single weekly pull request.

Closes #293.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# Keeps GitHub Actions dependencies up to date, including the SHA-pinned
+# reference to the shared backport workflow (qilimanjaro-tech/workflows) and the
+# action versions used across the other workflows. Bumps are grouped into a
+# single weekly PR.
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    groups:
+      github-actions:
+        patterns: ["*"]
+    commit-message:
+      prefix: "ci"
+    open-pull-requests-limit: 5
--- a/changes/292.misc.md
+++ b/changes/292.misc.md
@@ -1,0 +1,1 @@
+Added Dependabot configuration to keep the repository's GitHub Actions dependencies up to date.


DZPM (approved):
Great feature!

Notice this will try to update **major** versions too.
We have plenty of tests, so it should be ok.

The only concern is that a failing major update will block the PR, and the other updated won't be applied. It's reasonable 👍

fedonman:
> Great feature!
> 
> Notice this will try to update **major** versions too. We have plenty of tests, so it should be ok.
> 
> The only concern is that a failing major update will block the PR, and the other updated won't be applied. It's reasonable 👍

It's very rare case, because Github Actions maintainers try to keep backwards compatibility, at least for the previous version. If it ever happens, we will catch it when workflows try to run.

DZPM:
I didn't meant a major version in github actions, but in our own dependencies.
Like we could have a PR for 10 dependency fixes, the the PR will fail if a single dependency breaks the tests, and that's more probably to happen if updating major versions.

It's a acceptable tradeoff :+1:

fedonman:
This config is only about updating the GitHub Actions, not the project's dependencies.

PR #294, opened 2026-07-23
Increment the Version Number

Lumorti:
0.2.1 to 0.2.2


fedonman (changes requested):
You need to run `uv lock` in order to propagate the change to `uv.lock`. Also remove the "hotfix" from the title, we don't fix anything.

fedonman (approved):
Thanks Luke! 💯

PR #297, opened 2026-07-23
[SDK-364] Improve Ising Model Generation Speed

Lumorti:
The performance of `Model.random_ising()` has been improved, a 50 qubit fully-connected Ising model used to take 30s, it now takes 0.08s.


amirazzam2000 (approved):
That's a smart trick to speed up term construction because before it used to do simplification at each step, but this would do it only once!! Thanks Luke! 🙌

PR #305, opened 2026-07-28
[SDK-369] Fix Missing Parallel Pragmas

Lumorti:
Fixed a few cases of some parallelized loops using `#pragma omp for` instead of `#pragma omp parallel for`. Whilst this results in 2-3x speedup for those cases, those particular code paths are not currently used, and so instead this is just future-proofing their use.


amirazzam2000 (approved):
nice!!! good job Luke! 🙌

PR #307, opened 2026-07-29
[SDK-370] Fixed Cos and Sin having the same hash

Lumorti:
A bug was fixed in which some `MathematicalMaps` would have identical hashes, and thus some Terms (or anything using `MathematicalMaps` as keys) might have unintentionally combined them.


amirazzam2000 (approved):
Good catch!! Thanks luke 💯

fedonman (approved):
💯 💯

PR #308, opened 2026-07-29
[SDK-336] Improve Monte-Carlo Sampling

Lumorti:
Regarding Monte-Carlo simulations in QiliSim, previously in order to sample or calculate expectation values from the final state we had to construct the full density matrix from the Monte-Carlo ensemble. Now the expectation values are calculated based on the matrix-free application of the operator to each element of the ensemble, whilst samples are now taken directly.


amirazzam2000 (approved):
thanks Luke 💪

PR #312, opened 2026-07-31
[SDK-371] Random Knapsack

Lumorti:
Adding model constuctors for randomized problems (previously we only had `Model.random_ising()`):
- `Model.random_knapsack()`
- `Model.random_max_cut()`
- `Model.random_graph_coloring()`
- `Model.random_travelling_salesman()`

Usage:

```python
from qilisdk.core import Model
model = Model.random_knapsack(3)
```

### DIFF
--- a/changes/312.feature.md
+++ b/changes/312.feature.md
@@ -1,0 +1,13 @@
+Adding model constuctors for randomized problems (previously we only had `Model.random_ising()`):
+- `Model.random_knapsack()`
+- `Model.random_max_cut()`
+- `Model.random_graph_coloring()`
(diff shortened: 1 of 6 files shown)


fedonman (approved):
Thanks Luke! I was thinking if it made sense to use a "namespace" for all random methods like numpy does, e.g. `Model.random.knapsack()` or `Model.random.graph_coloring()` but we can discuss it separately, not important for this PR.

Lumorti:
Would you prefer I do that in this PR?

This would be the first use of this type of syntax in the library, are there other places we should use it?

fedonman:
No, you can merge as is.

PR #315, opened 2026-08-03
[SDK-373] Add Hamiltonian Constructors

Lumorti:
Added some convenience constructors for Hamiltonians:

```python
from qilisdk.analog import Hamiltonian

H = Hamiltonian.transverse_field_ising(nqubits=2, x_coefficient=1.3, zz_coefficient=-2)
H = Hamiltonian.transverse_field(nqubits=2, x_coefficient=1.3)
H = Hamiltonian.longitudinal_field(nqubits=2, z_coefficient=1.3)
H = Hamiltonian.ising(nqubits=2, zz_coefficient=2.0)
H = Hamiltonian.ising_chain(nqubits=4, zz_coefficient=2.0, periodic=True)
H = Hamiltonian.ising_grid(rows=2, columns=3, zz_coefficient=2.0)
H = Hamiltonian.heisenberg(nqubits=2, xx_coefficient=1.0, zz_coefficient=0.3)
```

Any coefficient can also be given as a list holding one value per term, rather than a single value
shared by every term:

```python
H = Hamiltonian.ising(nqubits=3, zz_coefficient=[0.5, -1.0, 2.0])
```


Lumorti:
I understand that the name "constant" is a bit strange, but I couldn't think of what else, like maybe "with_coefficients" is more appropriate but it's a bit long, "uniform" would be confusing since that sounds more like uniform random etc.

If you have any ideas for better names I'm open to it

src/qilisdk/analog/hamiltonian.py
```diff
@@ -591,6 +594,138 @@ def get_commuting_partitions(self) -> list[dict[tuple[PauliOperator, ...], compl
...
+        for coefficient, word in terms:
+            for operators in cls._place_pauli_word(cls._parse_pauli_word(word, nqubits), nqubits):
+                elements[operators] += coefficient
+        return cls(elements)
+
+    @classmethod
+    def random(
+        cls,
+        nqubits: int,
+        terms: Sequence[str],
```
amirazzam2000: Do we want terms to be strings? or maybe we can add two entries: 
- local fields 
- coupling 

and then we can provide hamiltonian objects there?

src/qilisdk/analog/hamiltonian.py
```diff
@@ -591,6 +594,138 @@ def get_commuting_partitions(self) -> list[dict[tuple[PauliOperator, ...], compl
...
+        """
+        if not word:
+            return [(PauliI(0),)]
+        return [
+            tuple(starmap(_get_pauli, zip(word, qubits, strict=True)))
+            for qubits in combinations(range(nqubits), len(word))
+        ]
+
+    @classmethod
+    def constant(cls, nqubits: int, terms: Sequence[tuple[Number, str]]) -> Hamiltonian:
```
amirazzam2000: I wouldn't add this, i don't see it's utility i would rather add constructors for known hamltonians, for instant transverse field Ising model where the user would specify the ranges of the coupling and each of the local fields. And then if the user wants it a constant instead of providing a range they can directly provide a number.

amirazzam2000 (changes requested):
This is a very good idea and a clear continuation of the constructors we have been adding. Thanks Luke!

src/qilisdk/analog/hamiltonian.py
```diff
@@ -631,6 +701,451 @@ def get_commuting_partitions(self) -> list[dict[tuple[PauliOperator, ...], compl
...
+            {
+                (_get_pauli("Z", first), _get_pauli("Z", second)): zz()
+                for first, second in combinations(range(nqubits), 2)
+            }
+        )
+        elements.update({(_get_pauli("Z", qubit),): z() for qubit in range(nqubits)})
+        return cls(elements)
+
+    @classmethod
+    def xy(
```
amirazzam2000: is that the technical name for this hamiltonian?

src/qilisdk/analog/hamiltonian.py
```diff
@@ -631,6 +701,451 @@ def get_commuting_partitions(self) -> list[dict[tuple[PauliOperator, ...], compl
 
         return partitions
 
+    @classmethod
+    def transverse_field(cls, nqubits: int, x_coefficient: Coefficient = 1.0, seed: int = 1) -> Hamiltonian:
```
amirazzam2000: what if the user wants to provide a list of coefficients one per qubit. something like a J_ij matrix for the interacting terms and an h_i vector for the local fields. Just suggesting ways people might want to use this.

amirazzam2000 (commented):
just leaving some questions for discussion.

src/qilisdk/analog/hamiltonian.py
Lumorti: I was thinking about it, but it's hard to handle, since if the user gives a list of two floats, do we treat that as a random range or the coefficients for two qubits?

amirazzam2000: well we can define the random differently, something like Random(x, y) which would make it clearer for the user and for us what they want. And Random could just be a simple data object.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -623,6 +677,391 @@ def get_commuting_partitions(self) -> list[dict[tuple[PauliOperator, ...], compl
...
+            Hamiltonian: the transverse-field Hamiltonian.
+
+        Raises:
+            ValueError: if ``nqubits`` is not greater than zero.
+            ValueError: if a list of coefficients does not hold one value per term.
+        """
+        _validate_nqubits(nqubits)
+        logger.debug("[Hamiltonian] Building transverse field over {} qubits", nqubits)
+        x = _coefficients(x_coefficient, nqubits, "x_coefficient")
+        return cls({(_get_pauli("X", qubit),): x[qubit] for qubit in range(nqubits)})
```
amirazzam2000: why using `_get_pauli` instead of just using `PauliX`

src/qilisdk/analog/hamiltonian.py
```diff
@@ -236,6 +236,60 @@ class PauliI(PauliOperator):
...
+        ValueError: if ``nqubits`` is not greater than zero.
+        ValueError: if ``nqubits`` is below ``minimum``.
+    """
+    if nqubits <= 0:
+        raise ValueError(f"nqubits must be greater than zero, got {nqubits}.")
+    if nqubits < minimum:
+        raise ValueError(f"{label} Hamiltonians need at least {minimum} qubits, got {nqubits}.")
+
+
+def _coefficients(coefficient: Coefficient, count: int, name: str) -> list[complex]:
```
amirazzam2000: nitpick but i would rename this to _parse_coefficients or _check_coefficients just to be more meaningful.

src/qilisdk/analog/hamiltonian.py
```diff
@@ -623,6 +677,391 @@ def get_commuting_partitions(self) -> list[dict[tuple[PauliOperator, ...], compl
...
+        Raises:
+            ValueError: if ``nqubits`` is less than 2.
+            ValueError: if a list of coefficients does not hold one value per term.
+        """
+        _validate_nqubits(nqubits, minimum=2, label="Ising")
+        logger.debug("[Hamiltonian] Building Ising model over {} qubits", nqubits)
+        pairs = list(combinations(range(nqubits), 2))
+        zz = _coefficients(zz_coefficient, len(pairs), "zz_coefficient")
+        z = _coefficients(z_coefficient, nqubits, "z_coefficient")
+        elements: dict[tuple[PauliOperator, ...], complex | Term | Parameter] = {
```
amirazzam2000: update this with the new expression

changes/315.feature.md
```diff
@@ -0,0 +1,20 @@
+Added some convenience constructors for Hamiltonians:
+
+```python
+from qilisdk.analog import Hamiltonian
+
+H = Hamiltonian.transverse_field_ising(nqubits=2, x_coefficient=1.3, zz_coefficient=-2)
```
amirazzam2000, line 6: can we also add an example of the list case in the changelog so that it has some visibility.

amirazzam2000 (approved):
Thanks a lot Luke! ⚡

I left some comments but I will approve so that you can merge when they are fixed.

changes/315.feature.md
Lumorti: Isn't this already there? it's just below

amirazzam2000: okay I am blind sorry.

src/qilisdk/analog/hamiltonian.py
Lumorti: good idea

src/qilisdk/analog/hamiltonian.py
Lumorti: good catch, it was added for caching, but I just checked and it doesn't actually improve speed at all, so I removed it

src/qilisdk/analog/hamiltonian.py
Lumorti: Done!

changes/315.feature.md
Lumorti: haha no worries

PR #328, opened 2026-08-05
Add QTensor Sample

### LINKED ISSUE #327: [Feature]: Add QTensor.sample()
### Feature Summary

Add the ability to sample from a state, either a state or a density matrix.

### Motivation / Use Case

It's just nice to have, in case you get a state from a state tomography readout and then decide later that you want samples from it.

### Proposed Solution

This routine already exists in the C++ side, it just needs to be exposed to the Python.

### Alternatives Considered

_No response_

### Priority

Low - Nice to have

### Additional Context

_No response_

Lumorti:
The ability to sample from a QTensor statevector/density matrix has been added, this uses the same method and output format as the sampling readout:

```python
state = QTensor.ghz(5)
samples = state.sample(100)
print(samples)
```

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/327


Lumorti:
https://github.com/qilimanjaro-tech/qilisdk/issues/327

fedonman (changes requested):
Check comment at Issue.

amirazzam2000 (approved):
Thanks Luke! 🙌

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -1929,6 +1930,84 @@ py::list QTensorCpp::probabilities_python() const {
...
+    for (Eigen::Index i = 0; i < dim; ++i) {
+        if (probs_data[i] < 0.0) {
+            probs_data[i] = 0.0;
+        } else {
+            total_prob += probs_data[i];
+        }
+    }
+    if (total_prob <= 0.0) {
+        throw py::value_error("Cannot sample from a QTensor with zero total probability");
+    }
```
fedonman: Any negative diagonal entry gets clamped to zero, not just numerical noise, so sampling an operator that is not a state at all comes back with plausible looking counts instead of an error. The Python docstring already promises a `ValueError` when "the probabilities do not form a valid distribution", so a tolerance on the clamp would make that true:

```python
import numpy as np

print(QTensor(np.diag([1.0, -1.0])).sample(10, seed=1))   # {'0': 10}, this is Pauli Z
print(QTensor(np.diag([-0.5, 0.6, 0.6, 0.3])).sample(10, seed=1))  # happily samples
```

```suggestion
    // Accumulate the total probability, clamping away negative entries within tolerance
    double total_prob = 0.0;
    double min_prob = 0.0;
#ifndef _WIN32
#if defined(_OPENMP)
#pragma omp parallel for reduction(+ : total_prob) reduction(min : min_prob) schedule(static)
#endif
#endif
    for (Eigen::Index i = 0; i < dim; ++i) {
        if (probs_data[i] < 0.0) {
            min_prob = std::min(min_prob, probs_data[i]);
            probs_data[i] = 0.0;
        } else {
            total_prob += probs_data[i];
        }
    }
    if (min_prob < -default_atol) {
        throw py::value_error("Cannot sample from a QTensor with negative probabilities");
    }
    if (total_prob <= 0.0) {
        throw py::value_error("Cannot sample from a QTensor with zero total probability");
    }
```

`SampleTest.NegativeProbability_Ignored` and `test_qtensor_sample_negative_probability_ignored` both use `-1e-14`, so they still pass. Worth a test for the case that now raises.

src/qilisdk_cpp/core/qtensor.cpp
```diff
@@ -1929,6 +1930,84 @@ py::list QTensorCpp::probabilities_python() const {
...
+#if defined(_OPENMP)
+#pragma omp parallel for schedule(static)
+#endif
+#endif
+        for (Eigen::Index i = 0; i < dim; ++i) {
+            probs_data[i] *= inv_total_prob;
+        }
+    }
+    // Note that this consumes the probabilities, turning them into a cumulative distribution in-place
+    return sample_from_probabilities(probs.data(), probs.size(), get_nqubits(), nshots, seed);
```
fedonman, line 1995: `sample_from_probabilities` seeds one generator per OpenMP thread, so the same seed gives different counts depending on how many threads happen to be active. On the readout path the thread count at least travels with the seed in `ExecutionConfig`, but `QTensor.sample` has no thread knob, so there is no way to get a reproducible draw. It also means an earlier `QiliSim` run changes later results, because `omp_set_num_threads` is process global:

```python
from qilisdk.backends import QiliSim, ExecutionConfig
from qilisdk.digital import Circuit, H

s = QTensor.uniform(3)
before = s.sample(1000, seed=42)
c = Circuit(1)
c.add(H(0))
QiliSim(execution_config=ExecutionConfig(num_threads=1, seed=1)).execute(DigitalPropagation(circuit=c), Readout().with_sampling(nshots=1))
print(before == s.sample(1000, seed=42))  # False
```

The fix belongs in `sample_from_probabilities` in `backends/qilisim/utils/random.cpp`: partition the shots into a fixed number of blocks and seed each block from `{seed, block}` instead of `{seed, thread_id}`, then the result is the same for any thread count. `test_qtensor_sample_is_deterministic_with_seed` only compares two calls in one process so it does not catch this.

fedonman (changes requested):
Thanks Luke. Left some comments.

fedonman (approved):
Thank you Luke!

PR #329, opened 2026-08-06
[SDK-164] Add ability to update schedule coefficients

Lumorti:
Added the ability to update the coefficient of a schedule without reconstructing the whole thing:

```python

sched = Schedule.linear(X(0), Z(0), total_time=10, dt=1)
sched.set_coefficient("driver", 0.5)
print(sched)

# Prints:

# Schedule(
#   hamiltonians={
#     'driver': X(0),
#     'problem': Z(0),
#   },
#   coefficients={
#     'driver': {0.0: 0.5, 10.0: 0.5},
#     'problem': {0.0: t * (0.1), 10.0: t * (0.1)},
#   },
#   dt=1,
#   total_time=None,
#   interpolation=None
# )
```


src/qilisdk/analog/schedule.py
```diff
@@ -537,6 +537,43 @@ def update_hamiltonian(
         if self._max_time is not None:
             self._coefficients[label].set_max_time(self._max_time)
 
+    def set_coefficient(
```
amirazzam2000, line 540: One thing about this, you might want to set the coefficient of a single point during the evolution. This would overwrite all the coefficients of the schedule for that hamiltonian.

amirazzam2000: also how is this different than calling `update_hamiltonian` with `new_hamiltonian` set to none?

amirazzam2000 (changes requested):
Thanks Luke! I left some questions in the comments 👍

src/qilisdk/analog/schedule.py
Lumorti: It's not that unique, there are many ways this can be done, I just did a quick PR for this issue that was in the backlog since a long time, the only description was "ability to set the coefficient of a schedule" so I tried to do just that

Lumorti: Do you remember what the original idea was? Happy to tweak or change or remove this entirely

PR #330, opened 2026-08-06
[SDK-234] Gates now return QTensors

### LINKED ISSUE #323: [Feature]: Gates should return a QTensor not a numpy array
### Feature Summary

Gates should return a QTensor rather than a numpy array.

### Motivation / Use Case

To be more consistent with the rest of the library.

### Proposed Solution

the `_generate_matrix()` for each gate which currently returns a np.ndarray should cast this numpy array to a QTensor.

### Alternatives Considered

_No response_

### Priority

Low - Nice to have

### Additional Context

_No response_

Lumorti:
All gates now return QTensors rather than a numpy array when their ` .matrix `property is queried, this allows for better compatibility with the rest of the library,

```python
from qilisdk.digital import X
gate = X(0)
print(gate.matrix)
# prints a QTensor, not a numpy array
```

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/323


amirazzam2000 (approved):
This has been sitting in the backlog for a while thanks for adding it 🙌

fedonman (approved):
Great feature 💯 Since this is a breaking change, after merging, make sure to update downstream libraries (e.g. qili-runtime) that use `.matrix`.

Lumorti:
Actually very good point Vyron, I just did a search through qililab and .matrix is used quite a bit on QiliSDK gates, as well as the class Rmw(BasicGate) which would need changing. But I think this is all code that should be removed once things are decoupled, right? Qililab shouldn't interface with SDK objects, so maybe this is a PR to merge after the decoupling

PR #332, opened 2026-08-06
[ISSUE-331] Add Single-Pauli Hamiltonian Constructors

### LINKED ISSUE #331: [Feature]: Hamiltonian.X and other single constructors
### Feature Summary

Right now if you want to generate a Hamiltonian containing a single Pauli you need to use:
```python
from qilisdk.analog import X
H = X(0)
```
but this can cause issues with the X gate from `qilisdk.digital`. I propose we start moving towards using the following instead, following the recent addition of Hamiltonian constructors:
```python
from qilisdk.analog import Hamiltonian
H = Hamiltonian.X(0)
```
We should keep the old X/Y/Z for now, but then we can slowly move to depreciating them.

### Motivation / Use Case

Although it's a bit longer than the original constructors, it saves any confusion with PauliX and X (gate) in the library.

### Proposed Solution

```python

```

### Alternatives Considered

_No response_

### Priority

Low - Nice to have

### Additional Context

_No response_

Lumorti:
Some more constructors for Hamiltonians have been added, now for single Paulis, which should be used over the loose X, Y and Z constructors from before, since those can be confused with the X, Y and Z from the digital module. The old functions remain there for backwards compatability, but we might depreciate them at some point. The new usage is as follows:

```python
from qilisdk.analog import Hamiltonian
h = Hamiltonian.X(0) + 2 * Hamiltonian.Z(1)
print(h)
```

This resolves https://github.com/qilimanjaro-tech/qilisdk/issues/331

### DIFF
--- a/changes/332.feature.md
+++ b/changes/332.feature.md
@@ -1,0 +1,7 @@
+Some more constructors for Hamiltonians have been added, now for single Paulis, which should be used over the loose X, Y and Z constructors from before, since those can be confused with the X, Y and Z from the digital module. The old functions remain there for backwards compatability, but we might depreciate them at some point. The new usage is as follows:
+
+```python
+from qilisdk.analog import Hamiltonian
+h = Hamiltonian.X(0) + 2 * Hamiltonian.Z(1)
+print(h)
+```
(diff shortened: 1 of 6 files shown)


fedonman (changes requested):
See comment in Issue.

PR #347, opened 2026-08-07
Added seed to Circuit.random()

### LINKED ISSUE #334: [Bug]: Circuit.random parameters are frozen and unseeded
### Bug Description

`Circuit.random` gives each parameter `bounds=(value, value)` and a label built from the float value, so `set_parameter_values` on a random circuit always raises. It also has no `seed` argument and draws from the global `random` module. Both come from `src/qilisdk/digital/circuit.py:437`.

Leave bounds unset (or default to `(-pi, pi)`), label parameters by index instead of embedding the float, and add `seed: int | None = None` backed by a local `random.Random(seed)`.

```python
from qilisdk.digital import CNOT, RX, Circuit

c = Circuit.random(2, {RX}, {CNOT}, 4)
print(c.get_parameter_names())  # ['theta-1.3707532506504077']
c.set_parameter_values([0.1] * c.nparameters)
```

### Expected Behavior

Parameters of a random circuit can be assigned, and `Circuit.random` accepts a `seed` instead of using the global RNG.

### Actual Behavior

```shell
  File "src/qilisdk/core/variables.py", line 1185, in check_value
    raise ValueError(f"The value provided ({value}) is outside the bound of the parameter {self.bounds}")
ValueError: The value provided (0.1) is outside the bound of the parameter (-1.3707532506504077, -1.3707532506504077)
```

### System Information

```shell
QiliSDK Version: 0.2.2
Python Version: 3.13.11
Numpy Version: 2.4.4
```

Lumorti:
Added the ability to specify a seed to `Circuit.random()`, rather than relying on the global rng.
Also a fixed a bug in which Parameterized gates generated in random circuits had their ranges specified incorrectly so trying to adjust their parameters later would fail. Usage remains the same, except now with the new optional `seed` argument:

```python
from qilisdk.digital import Circuit, X, RX, S, RY, U1, U2, U3, CNOT
c = Circuit.random(
    nqubits=3,
    single_qubit_gates={X, RX, S, RY, U1, U2, U3},
    two_qubit_gates={CNOT},
    ngates=10,
    seed=42,
  )
print(c)
```

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/334


amirazzam2000 (approved):
Thanks Luke! 🫶

PR #348, opened 2026-08-07
QiliSim Seed Reuse

### LINKED ISSUE #344: [Bug]: QiliSim reuses one seed for every execute()
### Bug Description

The seed is resolved once when the backend is built and passed unchanged to every execution, so repeated `execute()` calls on the same backend return identical shot counts. `ExecutionConfig._validate_seed` turns `None` into one fixed number via `secrets.randbelow(2**15)`, `QiliSim.__init__` freezes it into `self._solver_config`, and every sampling call on the C++ side re-seeds a fresh engine from `config.get_seed()`. The same seed also feeds the mid-circuit draw and the final draw within one run, so those are correlated too.

Treat the user seed as a root seed and derive a distinct sub-seed per execution and per sampling call. `secrets.randbelow(2**15)` should also be widened, which means updating the `0 <= seed < 2**15` assertion in `test_qilisim_specific_integration.py`.

```python
from qilisdk.backends import ExecutionConfig, QiliSim
from qilisdk.digital import Circuit, H

c = Circuit(nqubits=2)
c.add(H(0))
c.add(H(1))
ro = Readout().with_sampling(nshots=200)

b = QiliSim(execution_config=ExecutionConfig(num_threads=1))
print(b.execute(DigitalPropagation(circuit=c), readout=ro).get_samples())
print(b.execute(DigitalPropagation(circuit=c), readout=ro).get_samples())
```

### Expected Behavior

Consecutive `execute()` calls on the same backend give different samples.

### Actual Behavior

```shell
{'00': 57, '01': 46, '10': 52, '11': 45}
{'00': 57, '01': 46, '10': 52, '11': 45}
```

### System Information

```shell
QiliSDK Version: 0.2.2
Python Version: 3.13.11
Numpy Version: 2.4.4
```

Lumorti:
Previously after creating a QiliSim instance, each simulation would use the same seed for subsequent runs, now the seed is advanced after each run, so the following would produce different outputs (although still reproducible with the same base seed):

```python
backend = QiliSim(execution_config=ExecutionConfig(seed=42))
results_1 = backend.execute(...)
results_2 = backend.execute(...)
# results_1 != results_2
```

If you want the previous behavior, simply create a new backend each time to reset the random generator with a given seed.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/344


changes/348.bugfix.md
```diff
@@ -0,0 +1,11 @@
+Previously after creating a QiliSim instance, each simulation would use the same seed for subsequent runs, now the seed is advanced after each run, so the following would produce different outputs (although still reproducible with the same base seed):
+
+<!-- SKIP -->
```
fedonman, line 3: what's this line?

src/qilisdk_cpp/backends/qilisim/qilisim.h
```diff
@@ -13,17 +13,34 @@
...
     py::object execute_digital_propagation(const py::object& functional, const py::object& readout, const py::object& noise_model, const py::object& initial_state, const py::dict& solver_params);
     py::object execute_analog_evolution(const py::object& functional, const py::object& readout, const py::object& noise_model, const py::dict& solver_params);
     py::object execute_quantum_reservoir(const py::object& functional, const py::object& readout, const py::object& noise_model, const py::dict& solver_params);
+
+    // Restart the random stream with a given seed
+    void reset_seed(int seed) {
+        root_seed = seed;
+        seed_rng.seed(static_cast<uint64_t>(static_cast<uint32_t>(seed)));
+        seed_rng_initialised = true;
+    }
```
fedonman: This doesn't seem to do anything. The next execute reads the seed off the Python config, sees it does not match root_seed, and resets the stream right back, so reset_seed(99) on a backend built with seed=42 gives you the seed-42 stream:

```python
from qilisdk.backends import ExecutionConfig, QiliSim
from qilisdk.digital import Circuit, H

c = Circuit(nqubits=2)
c.add(H(0))
c.add(H(1))
ro = Readout().with_sampling(nshots=500)

def run(reset_to):
    b = QiliSim(execution_config=ExecutionConfig(seed=42, num_threads=1))
    first = b.execute(DigitalPropagation(circuit=c), readout=ro).get_samples()
    b.execute(DigitalPropagation(circuit=c), readout=ro)
    b.qili_sim.reset_seed(reset_to)
    return first, b.execute(DigitalPropagation(circuit=c), readout=ro).get_samples()
```

If we only want to reset to the Python config, then why don't we drop the parameter?

```suggestion
    // Restart the random stream from the root seed
    void reset_seed() {
        seed_rng.seed(static_cast<uint64_t>(static_cast<uint32_t>(root_seed)));
        seed_rng_initialised = true;
    }
```

src/qilisdk_cpp/backends/qilisim/utils/sample.h
```diff
@@ -23,7 +23,7 @@
 
 // GCOV_EXCL_BR_START
 
-std::map<std::string, int> apply_readout_error(const std::map<std::string, int>& counts, const NoiseModelCpp& noise_model_cpp, int n_qubits);
+std::map<std::string, int> apply_readout_error(const std::map<std::string, int>& counts, const NoiseModelCpp& noise_model_cpp, int n_qubits, int seed = 42);
```
fedonman: Why do we have the default here? The value should come from upstream.

```suggestion
std::map<std::string, int> apply_readout_error(const std::map<std::string, int>& counts, const NoiseModelCpp& noise_model_cpp, int n_qubits, int seed);
```

fedonman (changes requested):
Thank you Luke! Left some comments. I think the docs need also slight rephrasing at `docs/modules/backends/backends_qilisim.rst:218`.

changes/348.bugfix.md
Lumorti: This is the for the docs.sh checker telling it to not check that this code block can run (since it uses some placeholders as examples rather than forming full functionals etc.), I've used it a lot before and it can be found in various changelogs and docs

fedonman: First time I see them. Can't we refactor `docs.sh` to not check the changelog entries? These lines will go into the released CHANGELOG.md (no one will remember to manually edit the file and still might miss them)

fedonman (approved):
Thanks Luke!

PR #352, opened 2026-08-10
Fix Hardware Ansatz Final Layer

### LINKED ISSUE #349: [Bug]: HardwareEfficientAnsatz is missing its final rotation layer
### Bug Description

The class docstring and `_parameter_blocks` both say `(layers + 1)` single-qubit blocks, but `_build_circuit` runs only `layers` iterations and never emits a trailing block. So the circuit ends on an entangling block, which carries no parameters and cannot be tuned, and `layers=0` gives an empty circuit.

Pick the intended construction and make the docstring, the allocator and the build loop agree. Adding `final_rotation_layer: bool = True` would keep both. This changes the parameter count of existing ansatzes, so it needs a changelog entry, and the tests assert the current behaviour while their own docstrings say `(layers + 1)`.

```python
from qilisdk.digital import CNOT, U3

a = HardwareEfficientAnsatz(nqubits=3, layers=2, one_qubit_gate=U3, two_qubit_gate=CNOT)
print(a.nparameters, [g.name for g in a.gates])
print(HardwareEfficientAnsatz(nqubits=3, layers=0, one_qubit_gate=U3, two_qubit_gate=CNOT).gates)
```

### Expected Behavior

`(layers + 1)` single-qubit blocks, so 27 parameters here, and the circuit ends on a rotation layer.

### Actual Behavior

```shell
18 ['U3', 'U3', 'U3', 'CNOT', 'CNOT', 'U3', 'U3', 'U3', 'CNOT', 'CNOT']
[]
```

### System Information

```shell
QiliSDK Version: 0.2.2
Python Version: 3.13.11
Numpy Version: 2.4.4
```

Lumorti:
The `HardwareEfficientAnsatz` can now be customized as to whether it ends on a rotation or an entangling layer, by default the former, customizable via a new parameter:

```python
  nqubits=2,
  layers=2,
  connectivity="Linear",
  structure="grouped",
  final_rotation_layer=False,
)
```

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/349


src/qilisdk/digital/ansatz.py
```diff
@@ -45,8 +45,9 @@ def __init__(self, nqubits: int) -> None:
 @yaml.register_class
 class HardwareEfficientAnsatz(Ansatz):
     """
-    Hardware-efficient ansatz with `(layers + 1)` single-qubit blocks and ``layers``
-    entangling blocks.
+    Hardware-efficient ansatz with ``layers`` entangling blocks and, by default,
+    ``(layers + 1)`` single-qubit blocks: one before each entangling block plus one
+    one at the end, which can be disabled with ``final_rotation_layer=False``.
```
fedonman: One "one" too many.

```suggestion
    ``(layers + 1)`` single-qubit blocks: one before each entangling block plus one
    at the end, which can be disabled with ``final_rotation_layer=False``.
```

changes/352.feature.md
```diff
@@ -0,0 +1,11 @@
...
+
+```python
+ansatz = HardwareEfficientAnsatz(
+    nqubits=2,
+    layers=2,
+    connectivity="Linear",
+    structure="grouped",
+    final_rotation_layer=False,
+)
+```
```
fedonman, line 13: Reads as purely additive, but the default changes behaviour for every existing ansatz. Let's mention this.

```suggestion
The `HardwareEfficientAnsatz` can now be customized as to whether it ends on a rotation or an entangling layer, by default the former, customizable via a new parameter:

```python
    nqubits=2,
    layers=2,
    connectivity="Linear",
    structure="grouped",
    final_rotation_layer=False,
)
```

Note this changes the default circuit: there is now a trailing single-qubit block, so an ansatz has `(layers + 1)` parameterized blocks instead of `layers`. `nqubits=3, layers=2` with `U3` goes from 18 parameters to 27, and `layers=0` now gives one rotation block rather than an empty circuit. Pass `final_rotation_layer=False` for the old shape.
```

fedonman (approved):
I wrote these comments three weeks ago but never submitted them. Approving since they are documentation-only. Thanks Luke.

src/qilisdk/digital/ansatz.py
Lumorti: Good catch!

PR #353, opened 2026-08-10
Fix Circuit With Measurement Hash

### LINKED ISSUE #338: [Bug]: hash(Circuit) raises when the circuit has a measurement
### Bug Description

`Gate.__hash__` hashes `self.matrix`, but `M.matrix` raises `GateHasNoMatrixError`, so hashing a circuit with a measurement fails. It also hashes `tuple(self.get_parameters())`, which yields parameter names rather than values, so `RX(0, 0.1)` and `RX(0, 0.2)` only differ by their matrix.

Hash what `__eq__` compares instead: name, qubits and `get_parameter_values()`, rounded so gates equal under `np.allclose` hash equal.

```python
from qilisdk.digital import Circuit, H, M

c = Circuit(1)
c.add([H(0), M(0)])
hash(c)
```

### Expected Behavior

`hash(c)` returns an int.

### Actual Behavior

```shell
  File "src/qilisdk/digital/gates.py", line 257, in __hash__
    return qili_hash((self.name, self.qubits, self.matrix, tuple(self.get_parameters())))
  File "src/qilisdk/digital/gates.py", line 589, in matrix
    raise GateHasNoMatrixError
qilisdk.digital.exceptions.GateHasNoMatrixError
```

### System Information

```shell
QiliSDK Version: 0.2.2
Python Version: 3.13.11
Numpy Version: 2.4.4
```

Lumorti:
Fixed a bug in which Circuits could not be hashed if they contained a measurement.

Also tweaked the Circuit hash so that it follows the `Circuit.__eq__` method closer with respect to Parameters, such that now both are rounded to the same `atol` when compared.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/338

### DIFF
--- a/changes/353.bugfix.md
+++ b/changes/353.bugfix.md
@@ -1,0 +1,3 @@
+Fixed a bug in which Circuits could not be hashed if they contained a measurement.
+
+Also tweaked the Circuit hash so that it follows the `Circuit.__eq__` method closer with respect to Parameters, such that now both are rounded to the same `atol` when compared.
--- a/src/qilisdk/digital/gates.py
+++ b/src/qilisdk/digital/gates.py
@@ -255,8 +255,8 @@ class Gate(Parameterizable, ABC):
 
     def __hash__(self) -> int:
         atol = get_settings().atol
-        _PARAMETER_HASH_DECIMALS = int(-np.log10(atol))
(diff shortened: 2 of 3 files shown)


fedonman (approved):
Thanks Luke! 💯 💯

PR #355, opened 2026-08-10
Add Support For Monte-Carlo + Noise in QiliSim

### LINKED ISSUE #319: [Feature]: Noise + Monte Carlo support
### Feature Summary

Right now noise models are no supported if Monte Carlo is used. We should add this support to QiliSim.

### Motivation / Use Case

Doing more efficient noisy simulations, right now one has to use full density matrices.

### Proposed Solution

Allow the use of noise models in Monte Carlo, which involves creating a effective Hamiltonian and doing probabilistic jumps every so often.

### Alternatives Considered

_No response_

### Priority

Low - Nice to have

### Additional Context

_No response_

Lumorti:
Monte-Carlo is now supported in combination with a NoiseModel in QiliSim for all methods. Usage:

```python
from qilisdk.noise import NoiseModel, Depolarizing
from qilisdk.backends import QiliSim, ExecutionConfig, MonteCarloConfig, DigitalMethod

# Set up the circuit, here creating a Bell state
circuit = Circuit(nqubits=2)
circuit.add(H(0))
circuit.add(CNOT(0, 1))

# Simple noise model on all qubits
noise_model = NoiseModel()
noise_model.add(Depolarizing(probability=0.15))

# Set up the simulation
monte_carlo = MonteCarloConfig(trajectories=1000)
backend = QiliSim(
    digital_simulation_method=DigitalMethod.statevector(),
    noise_model=noise_model,
    execution_config=ExecutionConfig(seed=5, num_threads=1, monte_carlo=monte_carlo),
)

# Run it
result = backend.execute(DigitalPropagation(circuit=circuit), readout=readout)
print(result)
```

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/319


src/qilisdk_cpp/backends/qilisim/analog/time_evolution.cpp
```diff
@@ -13,14 +13,68 @@
...
+    likely per step. The rate used here is an upper bound, so this only fires when the step is clearly
+    too long.
+
+    Args:
+        jump_drift (SparseMatrix): The drift operator D of the effective Hamiltonian.
+        step_list (std::vector<double>): The schedule time points.
+    */
+    const double longest_dt = schedule_step_extremes(step_list).second;
+    const double expected_jumps = max_jump_rate_bound(jump_drift) * longest_dt;
+    if (expected_jumps > 0.5) {
```
amirazzam2000: I think we should tighten this, and mention in the docs that monte carlo simulation needs finer dt. maybe something like 0.1? and maybe we can actually store this value in the settings of qilisdk so the user can modify it. 

also we should raise the warning once, to avoid cluttering the user. So we should cache that we raised the warning. 

and in the docs we should explain more about this, adding a note that explains that you need smaller dt for monte carlo than the normal evolution to avoid errors.

amirazzam2000 (changes requested):
Thanks Luke!! 👻

I left a comment tho to be discussed before i approve the PR.

src/qilisdk_cpp/backends/qilisim/analog/time_evolution.cpp
Lumorti: all done!

PR #357, opened 2026-08-10
Drop the MKL numpy and scipy builds for the PyPI ones

### LINKED ISSUE #356: [Task]: Drop the MKL numpy and scipy builds for the PyPI ones
### What needs doing

Drop the MKL-linked numpy and scipy and go back to the PyPI ones. In `pyproject.toml` that means removing the `numpy-mkl` index block, the three `[tool.uv.sources]` entries that pin numpy, scipy and `mkl-service` to it, and the `mkl-service` dependency. The platform-split version pins on numpy and scipy only exist to steer the MKL resolution, so they collapse back to one line each. Then relock.

### Why

The MKL wheels came in with #99, when the heavy linear algebra still ran in Python. It does not any more. The simulator is C++ and uses Eigen, which is header-only and never links BLAS, and the GPU path uses cuBLAS/cuSOLVER. What is left of numpy and scipy in the Python layer is sparse matrix bookkeeping, a `4x4` `expm` in `gates.py`, and `scipy.optimize` in the optimizers. None of that is worth a custom BLAS.

fedonman:
Dropped the MKL-linked numpy and scipy builds and the unused `mkl-service` dependency. numpy and scipy now come from PyPI on every platform, which also removes `mkl`, `onemkl-license`, `intel-openmp`, `intel-cmplr-lib-ur`, `tbb`, `tcmlib` and `umf` from installs on Linux and Windows x86_64.

Closes #356.


Lumorti (approved):
Thanks Vyron! 🚀️

PR #358, opened 2026-08-10
Fix QiliSim Reservoirs Resetting the Wrong Qubit

### LINKED ISSUE #354: [Bug]: QuantumReservoir in QiliSim resets the wrong qubit
### Bug Description

`ReservoirLayer(qubits_to_reset=...)` builds its bitmask with the wrong qubit ordering, so it resets
the mirror-image qubit (`q` -> `nqubits - 1 - q`). On 2 qubits, asking to reset qubit 0 resets qubit 1;
with 3+ qubits only the middle qubit lands on itself.

Cause: `src/qilisdk_cpp/backends/qilisim/qilisim.cpp:527` builds the mask as
`reset_mask |= (1ULL << q)`, but qubit 0 is the **most** significant bit of the state index -
operators are assembled as `op ⊗ I ⊗ …` in `expand_operator`, and `collapse_state` correctly uses
`1L << (nqubits - 1 - qubit)` for the same job. The mask is shared by both reset paths
(`reset_trajectories` for Monte Carlo and the density-matrix loop below it), so both are off in the
same way.

```python
import numpy as np

from qilisdk.analog import Schedule
from qilisdk.analog import Z as pauli_z
from qilisdk.backends import ExecutionConfig, QiliSim
from qilisdk.core.qtensor import ket
from qilisdk.functionals.quantum_reservoirs import QuantumReservoir, ReservoirLayer

# |11>, an eigenstate of the Hamiltonian, so only the reset can change <Z>
layer = ReservoirLayer(
    evolution_dynamics=Schedule(hamiltonians={"h": pauli_z(0) + pauli_z(1)}, total_time=0.1, dt=0.05),
    qubits_to_reset=[0],
)
functional = QuantumReservoir(
    initial_state=ket(1, 1),
    reservoir_layer=layer,
    input_per_layer=[{}, {}],  # two layers, so the reset of the first shows up in the result
)
result = QiliSim(execution_config=ExecutionConfig(seed=1, num_threads=1)).execute(
    functional, readout=Readout().with_expectation(observables=[pauli_z(0), pauli_z(1)])
)
print("qubits_to_reset=[0] -> <Z0>, <Z1> =", np.round(result.get_expectation_values(), 4))
```

### Expected Behavior

Qubit 0 is reset to |0>, qubit 1 is left alone, so `<Z0> = +1` and `<Z1> = -1`:

```
qubits_to_reset=[0] -> <Z0>, <Z1> = [ 1. -1.]
```

### Actual Behavior

```shell
qubits_to_reset=[0] -> <Z0>, <Z1> = [-1.  1.]
```

### System Information

```shell
QiliSDK Version: 0.2.2rc1
Python Version: 3.13.6 (main, Aug 14 2025, 16:23:12) [Clang 20.1.4 ]
Platform: Linux 6.8.0-136-generic (#136~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul  3 16:29:11 UTC )
Processor: x86_64
g++ Version: g++ (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
```

Lumorti:
A bug was fixed in which the qubit resets in the QuantumReservoir QiliSim implementation were being done on the wrong qubits, specifically that trying to target qubit `0` would instead target qubit `nqubits-1` and vice-versa because of a endianness typo.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/354


amirazzam2000 (approved):
Good catch! This is a very critical bug in execution 👍

PR #360, opened 2026-08-11
Test Suite Build Performance

### LINKED ISSUE #359: [Bug]: Test Suite Takes Too Long
### Bug Description

The combined recompile + test suite (both the Python and the C++) in total takes more than 5 minutes to run which I think is the reason some GitHub runners are timing out. This should be reduced, perhaps via compiling with a reduced set of optimizations (to hopefully reduce compile times without affecting the test performance) as well as seeing which are the slowest tests.

Run the test suite via the GitHub action in a PR.

### Expected Behavior

It runs reliably.

### Actual Behavior

It sometimes times out.

### System Information

```shell
QiliSDK Version: 0.2.1
Python Version: 3.13.6 (main, Aug 14 2025, 16:23:12) [Clang 20.1.4 ]
Platform: Linux 6.8.0-136-generic (#136~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul  3 16:29:11 UTC )
Processor: x86_64
g++ Version: g++ (Ubuntu 11.4.0-1ubuntu1~22.04.3) 11.4.0
```

Lumorti:
The time taken to build and run the tests and other checks via GitHub actions has been reduced. This is mostly to speed up the dev workflow, which previous required waiting for 5 minutes for checks, which has now been reduced to 4 minutes. The max time limit for the tests has also been increased to 10 minutes, just in case.


.github/workflows/tests.yml
```diff
@@ -11,22 +11,31 @@ on:
...
-          - "3.12"
-          - "3.13"
+        # Only one job needs to produce the coverage
+        include:
+          - python-version: "3.11"
+            coverage: false
+          - python-version: "3.12"
+            coverage: false
+          - python-version: "3.13"
+            coverage: true
```
fedonman: oh nice, I wasn't aware of this option!

.github/workflows/tests.yml
```diff
@@ -42,39 +51,66 @@ jobs:
             **/uv.lock
           python-version: ${{ matrix.python-version }}
 
+      # Most pull requests don't touch the C++ at all, so reuse the compilation
+      - name: Set up ccache
+        uses: hendrikmuhs/ccache-action@d62db5f07c26379fc4b4e0916f098a92573c3b03  # v1.2.23
```
fedonman: how does the cache gets invalidated?

.github/workflows/tests.yml
```diff
@@ -42,39 +49,64 @@ jobs:
...
+          echo "VIRTUAL_ENV=$PWD/.venv" >> "$GITHUB_ENV"
+          echo "$PWD/.venv/bin" >> "$GITHUB_PATH"
+
       - name: Install qilisdk
+        if: ${{ !matrix.coverage }}
         run: |
-          uv -v sync --group dev --extra all-cu13 -Ccmake.define.tests=ON -Ccmake.build-type=Debug -Ccmake.define.coverage=ON --reinstall
+          uv sync --frozen --group dev --extra all-cu13 --reinstall-package qilisdk \
+            -Ccmake.build-type=Release \
+            -Ccmake.define.CMAKE_CXX_COMPILER_LAUNCHER=ccache
```
fedonman: Following previous comment, what's the usage of this, since we already have installed qilisdk in the previous step?

.github/workflows/tests.yml
```diff
@@ -42,39 +49,64 @@ jobs:
...
+          max-size: 500M
+
+      - name: Install qilisdk with coverage
+        if: matrix.coverage
+        run: |
+          uv sync --frozen --group dev --extra all-cu13 --reinstall-package qilisdk \
+            -Ccmake.build-type=Debug \
+            -Ccmake.define.CMAKE_CXX_FLAGS_DEBUG=-O0 \
+            -Ccmake.define.CMAKE_CXX_COMPILER_LAUNCHER=ccache \
+            -Ccmake.define.tests=ON \
```
fedonman, line 58: We install unoptimized and then we install again optimized?

fedonman (commented):
Thanks Luke! I have some questions.

.github/workflows/tests.yml
Lumorti: There are two install commands, one for if the `matrix.coverage` is true and one if it isn't

.github/workflows/tests.yml
Lumorti: There's the `if: matrix.coverage`

.github/workflows/tests.yml
Lumorti: This I'm still investigating, but I believe it works like a normal local C++ build cache where it checks whether files have been updated and if so rebuilds them. I'm gonna test it more on this PR/branch

Lumorti: Supposedly it's just a wrapper for https://ccache.dev/manual/4.13.6.html which sets up the environment and maintains the cache between runners

.github/workflows/tests.yml
Lumorti: In this case `coverage` is just a local variable included in the matrix, it's not specific to GitHub actions or anything, it's branched on later in the workflow

.github/workflows/tests.yml
Lumorti: Actually, I checked the cache statistics and it never seemed to get a cache hit, so for now I've removed it.

fedonman (approved):
Thanks Luke!

PR #361, opened 2026-08-11
Fix Multi Controlled Unitaries

### LINKED ISSUE #339: [Bug]: Multi-controlled gate decomposition produces the wrong unitary
### Bug Description

`DecomposeMultiControlledGatesPass` follows Barenco Lemma 7.5, which needs `V * V == U` exactly. The helpers build `sqrt(U)` and `U†` out of `RX` / `RY` / `RZ`, which only match up to a global phase. Under a control that global phase becomes a relative phase, so the decomposition is a different unitary.

Fix: use the exact diagonal gates where they exist (`sqrt(Z)=S`, `sqrt(S)=T`, `sqrt(U1(phi))=U1(phi/2)`, `adjoint(U1(phi))=U1(-phi)`), and for the rest return the residual phase `alpha` with the gate and emit it on the controls as `C^(k-1)(U1(alpha))`, which is exact and keeps the recursion exact.

```python
import numpy as np
from qilisdk.digital import Circuit, X
from qilisdk.digital.gates import Controlled
from qilisdk.digital.circuit_transpiler_passes import DecomposeMultiControlledGatesPass

c = Circuit(3)
c.add(Controlled(0, 1, basic_gate=X(2)))  # Toffoli
ref = c.to_matrix()
dec = DecomposeMultiControlledGatesPass().run(c).to_matrix()
print(np.round(np.diag(ref.conj().T @ dec), 4))
print(np.allclose(ref, dec))
```

### Expected Behavior

The decomposed circuit equals the original unitary, so `np.allclose(ref, dec)` is `True`.

### Actual Behavior

Toffoli picks up an extra `-i` on the `|11x>` block.

```shell
[1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 1.+0.j 0.-1.j 0.-1.j]
False
```

### System Information

```shell
QiliSDK Version: 0.2.2
Python Version: 3.13.11
Numpy Version: 2.4.4
```

Lumorti:
A bug has been fixed in which decomposed multi-controlled gates produced a wrong unitary, specifically that since it used the square root of certain gates to build the decomposition, so phases were incorrectly ignored.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/339

### DIFF
--- a/changes/361.bugfix.md
+++ b/changes/361.bugfix.md
@@ -1,0 +1,1 @@
+A bug has been fixed in which decomposed multi-controlled gates produced a wrong unitary, specifically that since it used the square root of certain gates to build the decomposition, so phases were incorrectly ignored.
--- a/src/qilisdk/digital/circuit_transpiler_passes/decompose_multi_controlled_gates_pass.py
+++ b/src/qilisdk/digital/circuit_transpiler_passes/decompose_multi_controlled_gates_pass.py
@@ -22,13 +22,19 @@ from qilisdk.digital.gates import BasicGate, Controlled
 
 from .circuit_transpiler_pass import CircuitTranspilerPass
 from .numeric_helpers import (
+    _EPS,
+    _u3_and_phase_from_unitary,
     _unitary_sqrt_2x2,
-    _zyz_from_unitary,
+    _wrap_angle,
 )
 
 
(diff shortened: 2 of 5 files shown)


fedonman (approved):
Looks good! 👏 Thanks Luke!

PR #375, opened 2026-08-13
Fix SABRE layout being applied twice

### LINKED ISSUE #369: [Bug]: SabreSwapPass applies the initial layout twice and misreports the final layout
### Bug Description

Two separate bugs in the topology pipeline.

SabreLayoutPass retargets the circuit to physical qubits and also records the mapping in `context.initial_layout`. SabreSwapPass then picks that mapping up and applies it a second time, so a circuit the layout pass already made routable comes out with extra SWAPs. On the 3-qubit star below over a line of 5 the router adds one SWAP where the chosen layout needs none, and it does that on every seed I tried.

The final layout map is wrong for its own reason. `sabre_swap_pass.py:209` iterates `sorted(layout_hint)`, which holds physical labels, so `result.layout` is keyed on physical qubits instead of logical ones. Reading it either raises KeyError or quietly returns the wrong qubit, and in the second snippet it goes out of range.

```python
from qilisdk.digital.circuit_transpiler_passes import SabreLayoutPass, SabreSwapPass, TranspilationContext

line = [(0, 1), (1, 2), (2, 3), (3, 4)]
c = Circuit(3); c.add(H(0)); c.add(CNOT(0, 1)); c.add(CNOT(0, 2))

ctx = TranspilationContext()
lay = SabreLayoutPass(line, seed=2); lay.attach_context(ctx)
placed = lay.run(c)

sw = SabreSwapPass(line, seed=2); sw.attach_context(ctx)
sw.run(placed)
bare = SabreSwapPass(line, seed=2)
bare.run(placed)
print("layout:", lay.last_layout, "| swaps with ctx:", sw.last_swap_count, "without:", bare.last_swap_count)
print("result.layout:", ctx.final_layout)

p = SabreSwapPass([(0, 1), (1, 2), (0, 2)], initial_layout=[2, 0])
p.attach_context(TranspilationContext())
c2 = Circuit(2); c2.add(H(0)); c2.add(CNOT(0, 1))
p.run(c2)
```

### Expected Behavior

Exactly one pass applies the layout, so routing a circuit SabreLayoutPass already placed needs no swaps. `result.layout` keys are always `range(nqubits)`.

### Actual Behavior

```shell
layout: [2, 1, 3] | swaps with ctx: 1 without: 0
result.layout: {1: 1, 2: 2, 3: 3}
  File "repro.py", line 21, in <module>
    p.run(c2)
  File "src/qilisdk/digital/circuit_transpiler_passes/sabre_swap_pass.py", line 210, in run
    logical_qubit: self.last_final_layout[logical_qubit] for logical_qubit in sorted(layout_hint)
IndexError: list index out of range
```

### System Information

```shell
Python 3.13.12
Linux x86_64
```

fedonman:
SabreLayoutPass retargeted the circuit and also recorded the mapping in the context, so SabreSwapPass applied it a second time and added swaps that were not needed. It now takes an `apply_layout` flag, and the context tracks whether the layout is already baked in so the router only applies it once. Also keys `final_layout` on logical qubits instead of physical labels, which could give the wrong qubit or raise IndexError. Closes #369.


Lumorti (approved):
Thanks Vyron, LGTM!

PR #377, opened 2026-08-14
Fix Schedule dt

### LINKED ISSUE #371: [Bug]: Schedule.tlist does not step by dt
### Bug Description

`Schedule.tlist` builds the grid with `linspace(0, T, int(T // dt))`, so the spacing comes out as `T/(N-1)`, not `dt`. Ask for `dt=1.0` over `T=10` and the grid steps by 1.111; `T=1.0, dt=0.1` steps by 0.125. All three backends integrate on this list, so `dt` sets the accuracy of every analog run and does not mean what it says. If `T < dt` the grid is empty.

```python

for T, dt in [(10.0, 1.0), (1.0, 0.1), (0.05, 0.1)]:
    tl = Schedule.linear(PauliX(0), PauliZ(0), T, dt=dt).tlist
    spacing = (tl[1] - tl[0]) if len(tl) > 1 else None
    print(f"T={T} dt={dt}: npoints={len(tl)} spacing={spacing}")
```

### Expected Behavior

The spacing equals `dt` and the last point is `T`, so `T=10, dt=1.0` gives 11 points spaced 1.0. If `T / dt` is not close to an integer, snap `dt` and say so, or raise, but do not quietly change the spacing.

### Actual Behavior

```shell
T=10.0 dt=1.0: npoints=10 spacing=1.1111111111111112
T=1.0 dt=0.1: npoints=9 spacing=0.125
T=0.05 dt=0.1: npoints=0 spacing=None
```

### System Information

```shell
Python 3.13.12
Linux x86_64
```

Lumorti:
Previously if you created a Schedule with `T = 1` and `dt= 0.1`, it would instead make 9 steps with `dt = 0.125`, as opposed to the expected 10 steps with `dt = 0.1`. Now the tlist construction behaves as one would expect, as well as rounding the dt if an integer multiple isn't possible (emitting a warning if so).Expand commentComment on line R1Resolved

Two things to watch when upgrading: `len(schedule)` and iterating a schedule now give N+1 items instead of N, and `schedule.dt` is now derived so it can differ from the value you passed in.

```python

T = 1.0
dt = 0.1
tlist = Schedule.linear(PauliX(0), PauliZ(0), T, dt=dt).tlist
print(tlist)
```

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/371


fedonman (dismissed):

changes/377.bugfix.md
```diff
@@ -0,0 +1,10 @@
+Previously if you created a Schedule with `T = 1` and `dt= 0.1`, it would instead make 9 steps with `dt = 0.125`, as opposed to the expected 10 steps with `dt = 0.1`. Now the tlist construction behaves as one would expect, as well as rounding the dt if an integer multiple isn't possible (emitting a warning if so).
```
fedonman, line 1: Two more things worth mentioning here, both visible to users on upgrade.

```suggestion
Previously if you created a Schedule with `T = 1` and `dt= 0.1`, it would instead make 9 steps with `dt = 0.125`, as opposed to the expected 10 steps with `dt = 0.1`. Now the tlist construction behaves as one would expect, as well as rounding the dt if an integer multiple isn't possible (emitting a warning if so).

Two things to watch when upgrading: `len(schedule)` and iterating a schedule now give N+1 items instead of N, and `schedule.dt` is now derived so it can differ from the value you passed in.
```

src/qilisdk/analog/schedule.py
```diff
@@ -560,7 +605,7 @@ def __len__(self) -> int:
         Get the total number of discrete time steps in the annealing process.
 
         Returns:
-            int: The number of time steps, calculated as T / dt.
+            int: The number of sampling times in :attr:`tlist`, i.e. ``T / dt + 1``.
```
fedonman, line 633: This is one of them. The `AnnealingAnsatz` fix in this PR shows it can bite, anyone else iterating a schedule to evolve it now does one extra step.

src/qilisdk/analog/schedule.py
```diff
@@ -350,7 +384,18 @@ def _get_coefficients_max_time(self) -> float:
 
     @property
     def dt(self) -> float:
```
fedonman, line 401: And the other. `s.set_dt(0.3)` then `s.dt` gives 0.303. Docstring covers it nicely, just wants a changelog line too.

fedonman (changes requested):
Thanks Luke. Left a couple of comments, only the changelog one needs doing.

src/qilisdk/analog/schedule.py
```diff
@@ -350,7 +384,18 @@ def _get_coefficients_max_time(self) -> float:
...
+        This is the ``dt`` the schedule was built with, snapped to the nearest step that
+        divides :attr:`T` exactly when the two are incommensurate.
+
+        Returns:
+            float: The effective time resolution.
+        """
+        T = self.T
+        if T <= 0:
+            return self._dt
+        return T / self._nsteps(T)
```
amirazzam2000: why not cache this value? i feel like the dt might be accessed a couple of times.

src/qilisdk/analog/schedule.py
Lumorti: Fair! I updated the changelog as suggested

amirazzam2000 (approved):
Thanks Luke! 🌊

fedonman (approved):
Thank you Luke!

PR #378, opened 2026-08-14
Schedule Eigenvalues Zero Fix

### LINKED ISSUE #372: [Bug]: Schedule.eig returns the wrong spectrum when a coefficient is zero
### Bug Description

`Schedule.eig` calls `to_qtensor()` with no `total_nqubits`, so at times where a coefficient is zero the summed Hamiltonian has no terms on those qubits and the matrix shrinks. At the start of a normal anneal the problem coefficient is 0, so the first row of eigenvalues is the spectrum of a smaller system and the rows have different lengths. `draw_eigenvalues` then either plots fewer levels than asked for or raises IndexError. The overlap path in the schedule renderer has the same unpadded call.

```python

drv, prb = PauliX(0).to_hamiltonian(), PauliZ(0) * PauliZ(1)

s = Schedule.linear(drv, prb, 10.0, dt=1.0)
vals, _ = s.eig(levels=4)
print("row lengths:", [len(v) for v in vals], "row 0:", vals[0])

s2 = Schedule.linear(prb, drv, 10.0, dt=1.0)
s2.draw_eigenvalues(levels=4, filepath="out.png")
```

### Expected Behavior

Every row has `levels` entries and holds the spectrum of the full 2-qubit register, and `draw_eigenvalues` plots all 4 levels.

### Actual Behavior

```shell
row lengths: [2, 4, 4, 4, 4, 4, 4, 4, 4, 4] row 0: [-0.9999999999999998, 0.9999999999999998]
  File "repro.py", line 10, in <module>
    s2.draw_eigenvalues(levels=4, filepath="out.png")
  File "src/qilisdk/utils/visualization/schedule_renderers.py", line 310, in plot
    new_eigenvalues.append([full_eigenvalues[j][i] for j in range(len(full_eigenvalues))])
                            ~~~~~~~~~~~~~~~~~~~^^^
IndexError: list index out of range
```

### System Information

```shell
Python 3.13.12
Linux x86_64
```

Lumorti:
A bug was fixed in which `Schedule.eig()` didn't pass nqubits when calling to `Hamiltonian.to_qtensor()`, so if one Hamiltonian had less qubits than the other (i.e. one Hamiltonian was `X(0)` and the other was `X(0) + X(1)`) then there was a mismatch in the array sizes.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/372


fedonman (approved):
Thank you Luke.

PR #381, opened 2026-08-14
Schedule extrapolation fix

### LINKED ISSUE #373: [Bug]: Linear interpolation extrapolates past the defined time points
### Bug Description

Outside the defined time points, `Interpolator` in LINEAR mode keeps extending the slope of the first or last segment instead of holding the endpoint value. A schedule's `T` is the max over all coefficients, so a Hamiltonian defined over a shorter window gets sampled out of its own range on every run. Turning a driver off early is normal, and today its coefficient does not stay at 0, it goes negative and grows. Nothing warns.

```python

s = Schedule(
    hamiltonians={"driver": PauliX(0).to_hamiltonian(), "problem": PauliZ(0).to_hamiltonian()},
    coefficients={"driver": {0.0: 1.0, 2.0: 0.0}, "problem": {0.0: 0.0, 10.0: 1.0}},
    dt=1.0,
)
for t in [0.0, 2.0, 4.0, 10.0]:
    print(f"H({t}) = {s[t]}   driver coeff = {s.coefficients['driver'][t]}")
```

### Expected Behavior

The driver coefficient holds at 0 after t=2, so `H(10.0)` is just `Z(0)`. If extrapolation is wanted it should be opt-in, for example `Interpolator(..., extrapolate=True)`, exposed from `Schedule`.

### Actual Behavior

```shell
H(0.0) = X(0)   driver coeff = 1.0
H(2.0) = 0.2 Z(0)   driver coeff = 0.0
H(4.0) = - X(0) + 0.4 Z(0)   driver coeff = -1.0
H(10.0) = -4 X(0) + Z(0)   driver coeff = -4.0
```

### System Information

```shell
Python 3.13.12
Linux x86_64
```

Lumorti:
Schedules with coefficients that don't span the full time are now kept constant rather than extrapolating. Previously a Schedule like the following would extrapolate it's coefficient to below zero, whilst the user probably expected that it remain at zero. This behavior can be re-enabled by an `extrapolate` arg to Schedule, but by default it's off:

```python
s = Schedule(
    hamiltonians={"driver": PauliX(0).to_hamiltonian(), "problem": PauliZ(0).to_hamiltonian()},
    coefficients={"driver": {0.0: 1.0, 2.0: 0.0}, "problem": {0.0: 0.0, 10.0: 1.0}},
    dt=1.0,
    extrapolate=False,
)
s.draw()
```

Now correctly outputs the following, rather than the driver line going below zero as before:

image

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/373


fedonman (dismissed):

src/qilisdk/analog/schedule.py
```diff
@@ -113,13 +116,17 @@ def __init__(
...
             # Build hamiltonian schedule
             if ham not in coefficients:
-                self._coefficients[ham] = Interpolator({0: 1}, interpolation=interpolation, nsamples=int(1 / dt))
+                self._coefficients[ham] = Interpolator(
+                    {0: 1}, interpolation=interpolation, nsamples=int(1 / dt), extrapolate=extrapolate
+                )
                 continue
             coeff = copy(coefficients[ham])
             if isinstance(coeff, Interpolator):
                 self._coefficients[ham] = coeff
```
fedonman, line 125: `extrapolate` gets dropped here, so `Schedule(..., extrapolate=True)` with a pre-built `Interpolator` still holds instead of extrapolating. Either forward it or say so in the docstring, just not silently.

```suggestion
                self._coefficients[ham] = coeff  # keeps its own extrapolate setting
```

src/qilisdk/analog/schedule.py
```diff
@@ -436,7 +443,9 @@ def _add_hamiltonian_from_dict(
         if label in self._hamiltonians:
             raise ValueError(f"Can't add Hamiltonian because label {label} is already associated with a Hamiltonian.")
         self._hamiltonians[label] = hamiltonian
-        self._coefficients[label] = Interpolator(coefficients, interpolation, nsamples=int(1 / self.dt))
+        self._coefficients[label] = Interpolator(
+            coefficients, interpolation, nsamples=int(1 / self.dt), extrapolate=self._extrapolate
+        )
 
     def _add_hamiltonian_from_interpolator(
```
fedonman, line 457: Same here, `self._extrapolate` never reaches the interpolator.

src/qilisdk/core/interpolator.py
```diff
@@ -516,6 +523,17 @@ def get_coefficient_expression(self, time_step: float) -> Number | Term | Parame
...
+        Warn (once per interpolator) that a coefficient is being evaluated outside its defined time range without extrapolation.
+        """
+        if self._warned_extrapolation or self._tlist is None or len(self._tlist) < 2:  # ruff: ignore[magic-value-comparison]
+            return
+        self._warned_extrapolation = True
+        logger.warning(
+            "[Interpolator] At least one Schedule coefficient is being assumed constant outside its defined time range."
+        )
+
     def _get_coefficient_expression_step(self, time_step: float) -> Number | Term | Parameter:
```
fedonman, line 550: STEP is now inconsistent with LINEAR: `Interpolator({2: 1.0, 4: 0.0}, STEP)[0]` gives `0.0`, not `1.0`, and no warning. Pre-existing, not yours, but the two modes disagree now. Follow-up issue is fine.

src/qilisdk/core/interpolator.py
```diff
@@ -516,6 +523,17 @@ def get_coefficient_expression(self, time_step: float) -> Number | Term | Parame
...
 
+    def _warn_no_extrapolation(self) -> None:
+        """
+        Warn (once per interpolator) that a coefficient is being evaluated outside its defined time range without extrapolation.
+        """
+        if self._warned_extrapolation or self._tlist is None or len(self._tlist) < 2:  # ruff: ignore[magic-value-comparison]
+            return
+        self._warned_extrapolation = True
+        logger.warning(
+            "[Interpolator] At least one Schedule coefficient is being assumed constant outside its defined time range."
```
fedonman: Can we say which range? With a few Hamiltonians there is no way to tell which coefficient this is about.

```suggestion
            "[Interpolator] Coefficient held constant outside its defined range [{}, {}].",
            self._tlist[0],
            self._tlist[-1],
```

docs/modules/analog/analog_schedule.rst
```diff
@@ -48,6 +48,7 @@ Key arguments
     - or a callable returning a coefficient. This callable can take a parameter ``t`` that will be replaced by time. Moreover, any other parameters passed to this callable need to have a default value or have their value specified in the ``**kwargs``.
 - **interpolation** (:class:`~qilisdk.core.interpolator.Interpolation`): ``LINEAR`` (default) or ``STEP`` behavior between provided points.
 - **total_time** (float | Parameter | Term | None): Optional max time that rescales all time points while preserving relative positions.
+- **extrapolate** (bool): Whether to extrapolate a coefficient outside its own time points.
```
fedonman: This drops the default and the warning. The longer wording is still in the po files as obsolete `#~` blocks, looks like it got shortened by accident.

```suggestion
- **extrapolate** (bool): How to evaluate a coefficient outside its own time points. Defaults to ``False``, which holds the first and last values constant and warns the first time it happens. With ``True`` the first and last linear segments are extended instead.
```

fedonman (changes requested):
Thanks Luke. Left some comments, the first one is the only blocker.

src/qilisdk/analog/schedule.py
```diff
@@ -96,7 +98,8 @@ def __init__(
 
         self._hamiltonians = hamiltonians if hamiltonians is not None else {}
         self._coefficients: dict[str, Interpolator] = {}
-        self._interpolation = None
+        self._interpolation = interpolation
```
fedonman, line 101: Nice catch, the field was never assigned so repr always printed `interpolation=None`. Unrelated to this issue though, worth a changelog line.

docs/modules/analog/analog_schedule.rst
Lumorti: Ah good catch, I started changing the wording but forgot to add the default back!

src/qilisdk/analog/schedule.py
Lumorti: Added changelog

src/qilisdk/analog/schedule.py
Lumorti: Added a warning

src/qilisdk/analog/schedule.py
Lumorti: Added a warning

src/qilisdk/core/interpolator.py
Lumorti: Good idea!

src/qilisdk/core/interpolator.py
Lumorti: Made them both consistent, just copied the block from the linear section to clamp and warn

fedonman (approved):
Thanks Luke!

PR #384, opened 2026-08-14
Qubo Bitwise Encoding Error

### LINKED ISSUE #374: [Bug]: to_qubo() crashes for integer variables with the default encoding
### Bug Description

`Variable` defaults to `encoding=Bitwise`, but `Bitwise.encoding_constraint` raises `NotImplementedError`, so the most natural way to write an integer variable gives a model that cannot be converted to a QUBO at all. Bitwise needs no encoding constraint: every bit pattern already decodes to a value inside the bounds. `QUBO._check_variables` also already handles a `None` return, so returning `None` from `Bitwise.encoding_constraint` is enough.

```python
from qilisdk.core.model import Model
from qilisdk.core.variables import BinaryVariable, Domain, Variable

v = Variable("n", Domain.POSITIVE_INTEGER, bounds=(0, 5))  # Bitwise by default
print("decoded:", [v.evaluate([(i >> b) & 1 for b in range(3)]) for i in range(8)])

m = Model("m")
m.set_objective(v + BinaryVariable("b"))
m.to_qubo()
```

### Expected Behavior

`to_qubo()` works with the default encoding. Switching to `OneHot` is the only thing that works today, and nothing says so.

### Actual Behavior

```shell
decoded: [0, 1, 2, 3, 2, 3, 4, 5]
  File "repro.py", line 9, in <module>
    m.to_qubo()
  File "src/qilisdk/core/variables.py", line 424, in encoding_constraint
    raise NotImplementedError("Bitwise encoding constraints are not supported at the moment")
NotImplementedError: Bitwise encoding constraints are not supported at the moment
```

### System Information

```shell
Python 3.13.12
Linux x86_64
```

Lumorti:
Fixed a bug in which integer variables using Bitwise encoding would raise an unnecessary error when doing `Model.to_qubo()`.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/374


fedonman (approved):
Thank you Luke.

PR #388, opened 2026-08-15
Split the publish index input into platform and repository

### LINKED ISSUE #387: [Task]: Split the publish workflow index input into platform and repository
### What needs doing

In `.github/workflows/publish.yml`, rename the `index` dispatch input to `platform` with two choices, `pypi` and `aws`, and add a string input `repository` for the AWS CodeArtifact repository to publish to. The CodeArtifact job currently hardcodes `CA_REPOSITORY: qilimanjaro`, so `repository` takes its place. Delete the pyx `deploy-index` job while doing this, along with the `speqtrum` and `euroqcs` options and the `PYX_API_KEY` secret it is the only user of.

Actions cannot show an input conditionally, so `repository` will always appear in the dispatch form. Default it to `qilimanjaro` and fail the job early if it is empty.

### Why

The `index` input mixes two things, which platform to publish to and which named index inside it, and the CodeArtifact repository cannot be chosen at dispatch time. Publishing to a different CodeArtifact repository means editing the workflow. pyx is no longer used.

fedonman:
Renames the publish workflow's `index` dispatch input to `platform` with two choices, `pypi` and `aws`, and adds a `repository` input that names the CodeArtifact repository to publish to instead of hardcoding `qilimanjaro`. Actions cannot hide an input, so `repository` always shows in the form and the job fails early if it is empty. The pyx `deploy-index` job goes away with the `speqtrum` and `euroqcs` options, so the `PYX_API_KEY` secret can be deleted from the repo settings.

Closes #387


Lumorti (approved):
LGTM, thanks Vyron!

PR #396, opened 2026-08-17
Fix QiliSim Permutation Affecting Original

### LINKED ISSUE #390: [Bug]: QiliSim perturbations mutate the caller's circuit and stack up
### Bug Description

QiliSim applies a `ParameterPerturbation` by writing the perturbed value back into the live gate with `set_parameters` and never restores it, so the caller's `Circuit` is permanently rewritten and repeated `execute()` calls stack the perturbation instead of drawing independently around the original value. `parse_gates` iterates the circuit's gates with no copy and nothing in `qilisim.cpp` puts the originals back. The analog paths do the same to the `Schedule`. CudaBackend snapshots the parameters before perturbing and restores them afterwards, so the two backends disagree and QiliSim is the wrong one.

```python
from qilisdk.digital.circuit import Circuit
from qilisdk.digital.gates import RX
from qilisdk.noise import NoiseModel, OffsetPerturbation

circuit = Circuit(nqubits=1)
circuit.add(RX(0, theta=0.5))
noise = NoiseModel()
noise.add(OffsetPerturbation(offset=0.1), gate=RX, parameter="theta")

backend = QiliSim(noise_model=noise)
for _ in range(4):
    backend.execute(DigitalPropagation(circuit), readout=Readout().with_sampling(nshots=100))
    print(circuit.gates[0].get_parameters())
```

### Expected Behavior

`theta` stays at 0.5 after every run, and each run perturbs from 0.5. CudaBackend on the same input prints `{'theta': 0.5}` four times.

### Actual Behavior

```shell
{'theta': 0.6}
{'theta': 0.7}
{'theta': 0.7999999999999999}
{'theta': 0.8999999999999999}
```

A VariationalProgram or any loop over a perturbed circuit gets a random walk on its angles instead of independent draws, and nothing warns.

### System Information

```shell
qilisdk 0.2.2rc1 (dev checkout)
Python 3.13.12
Linux x86_64
```

Lumorti:
Fixed a bug in which simulating parameter perturbations via QiliSim would perturb the original Circuit/Schedule rather than making copies of the Parameters and perturbing those.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/390


fedonman (approved):
LGTM 👍 Thanks Luke!

Issue #389, opened 2026-08-15
[Bug]: Per-qubit noise hits the wrong qubits on multi-qubit gates


fedonman:
### Bug Description

`get_relevant_kraus_operators` looks up per-qubit Kraus operators over a gate's target list and then expands each one onto every target, so noise attached to a single qubit is applied to all targets of a multi-qubit gate. The qubit the noise was attached to is never carried through, and control qubits are never looked up, so noise on a CNOT control is dropped. CudaBackend is wrong in a different way: it registers the channel against the exact operand list, so it drops the channel instead of over-applying it. The two backends disagree with each other as well as with the right answer.

### Minimal Reproducible Example

```python
from qilisdk.backends import CudaBackend, QiliSim
from qilisdk.digital.circuit import Circuit
from qilisdk.digital.gates import CNOT, SWAP
from qilisdk.functionals import DigitalPropagation
from qilisdk.noise import BitFlip, NoiseModel
from qilisdk.readout import Readout

noise = NoiseModel()
noise.add(BitFlip(probability=1.0), qubits=[0])  # qubit 0 only

for gate in (SWAP(0, 1), CNOT(0, 1)):
    for backend in (QiliSim, CudaBackend):
        circuit = Circuit(nqubits=2)
        circuit.add(gate)
        result = backend(noise_model=noise).execute(
            DigitalPropagation(circuit), readout=Readout().with_sampling(nshots=200)
        )
        print(gate, backend.__name__, result.get_samples())
```

### Expected Behavior

All four cases flip exactly one qubit: `{'10': 200}`.

### Actual Behavior

```shell
SWAP(0, 1) QiliSim {'11': 200}      # applied to both targets
SWAP(0, 1) CudaBackend {'00': 200}  # dropped
CNOT(0, 1) QiliSim {'00': 200}      # dropped, qubit 0 is the control
CNOT(0, 1) CudaBackend {'00': 200}  # dropped
```

Nothing warns, and the noise integration tests only use single-qubit gates, so all of them pass.

### System Information

```shell
qilisdk 0.2.2rc1 (dev checkout)
Python 3.13.12
Linux x86_64
```

Lumorti:
Fixed by the combination of the two above PRs

PR #397, opened 2026-08-17
QiliSim Handles noise on Controls Correctly

Lumorti:
QiliSim now handles noise acting on controls correctly: previously if you had noise on qubit 1 and then did a CNOT controlled on qubit 1, the noise would not apply since it only checked for the targets of the gate, not the controls.

Also fixed qubit-specific noise spreading to all the targets of a multi-target gate: previously if you had noise on qubit 1 and then did a swap between qubits 1 and 2, the noise was applied to both, but it should only be applied to qubit 1.

Global and per-gate noise is now applied independently to each qubit a gate acts on, instead of as one correlated operator across the targets.

This is the QiliSim fix for https://github.com/qilimanjaro-tech/qilisdk/issues/389


src/qilisdk_cpp/backends/qilisim/noise/noise_model.cpp
```diff
@@ -191,62 +236,54 @@ std::string NoiseModelCpp::make_gate_key(const std::string& base_name, int num_c
...
-    for (int qubit : target_qubits) {
-        auto key = std::make_pair(gate_key, qubit);
-        if (cached_kraus_operators_per_gate_qubit.find(key) != cached_kraus_operators_per_gate_qubit.end()) {
-            for (const auto& operator_set : cached_kraus_operators_per_gate_qubit.at(key)) {
-                relevant_operators.push_back(operator_set);
-            }
+        // Add per-qubit Kraus operators, applied only to the qubit they are attached to
+        auto qubit_it = cached_kraus_operators_per_qubit.find(qubit);
+        if (qubit_it != cached_kraus_operators_per_qubit.end()) {
+            append_expanded_sets(relevant_operators, qubit_it->second, {qubit}, nqubits);
```
fedonman: A whole-register set found under a per-qubit key still gets applied once per qubit it was attached to. `nm.add(KrausChannel([X (x) I]), qubits=[0, 1])` on 2 qubits applies `X (x) I` twice, so it cancels and you get `{'00': 100}` instead of `{'10': 100}`. Same on main, so not a regression, but this is the function that decides it now. A per-qubit set that is not single-qubit is ambiguous in the same way the 3-qubit case is, so I would just reject it: give `append_expanded_sets` an `allow_full_register` flag and pass `false` from the two per-qubit calls. Happy with a follow-up issue instead.

changes/397.bugfix.md
```diff
@@ -0,0 +1,3 @@
+QiliSim now handles noise acting on controls correctly: previously if you had noise on qubit 1 and then did a CNOT controlled on qubit 1, the noise would not apply since it only checked for the targets of the gate, not the controls.
+
+Also fixed qubit-specific noise not spreading to all targets of a multi-target gate: previously if one did noise on qubit 1 and then did a swap between qubits 1 and 2, the noise would applied to both 1 and 2, but it should only be applied to 1 since it's qubit specific.
```
fedonman: The second paragraph says the opposite of what it fixes, and it is missing the global-noise change, which also moves results: on main a global `BitFlip(0.5)` on a `SWAP` only ever gives `00` or `11`, now it gives all four outcomes.

```suggestion
Also fixed qubit-specific noise spreading to all the targets of a multi-target gate: previously if you had noise on qubit 1 and then did a swap between qubits 1 and 2, the noise was applied to both, but it should only be applied to qubit 1.

Global and per-gate noise is now applied independently to each qubit a gate acts on, instead of as one correlated operator across the targets.
```

tests/integration/noise/test_noise_backends.py
```diff
@@ -82,6 +82,50 @@ def test_qilisim_backend_bit_flip_two_qubits_sampling(backend_class):
     assert result.get_samples() == {"01": 100}
 
 
+@pytest.mark.parametrize("gate", [SWAP(0, 1), CNOT(0, 1)])
```
fedonman, line 85: #398 adds these same three tests right here with `_cuda_` names, so the two will conflict and leave near-duplicates. Whichever lands second, could you fold them into one set parametrized over `backends` like the rest of the file? The issue was that the two backends disagreed with each other, so one shared test is the thing that keeps them honest.

src/qilisdk_cpp/backends/qilisim/noise/noise_model.cpp
```diff
@@ -191,62 +236,54 @@ std::string NoiseModelCpp::make_gate_key(const std::string& base_name, int num_c
     return base_name + "_c" + std::to_string(num_controls);
 }
 
-std::vector<std::vector<SparseMatrix>> NoiseModelCpp::get_relevant_kraus_operators(const std::string& gate_name, int num_controls, const std::vector<int>& target_qubits, int nqubits) const {
+std::vector<std::vector<SparseMatrix>> NoiseModelCpp::get_relevant_kraus_operators(const std::string& gate_name, int num_controls, const std::vector<int>& gate_qubits, int nqubits) const {
     /*
-    Get all relevant Kraus operators for a given gate and its target qubits.
+    Get all relevant Kraus operators for a given gate, expanded onto the full register.
+
+    Global and per-gate noise is applied independently to every qubit the gate acts on, while
```
fedonman, line 250: Nice to have this written down. Could it also go in `docs/modules/noise/noise_overview.rst`? "which qubits does my channel land on for a CNOT" is the question the issue was really about, and right now the answer only exists in a C++ comment.

src/qilisdk_cpp/backends/qilisim/noise/noise_model.cpp
```diff
@@ -191,62 +236,54 @@ std::string NoiseModelCpp::make_gate_key(const std::string& base_name, int num_c
...
     const std::string gate_key = make_gate_key(gate_name, num_controls);
 
-    // Add global Kraus operators
-    if (!cached_kraus_operators_global.empty()) {
-        for (const auto& operator_set : cached_kraus_operators_global) {
-            relevant_operators.push_back(operator_set);
-        }
-    }
+    // Add global Kraus operators, applied to each qubit the gate acts on
+    append_expanded_sets(relevant_operators, cached_kraus_operators_global, gate_qubits, nqubits);
```
fedonman: Not for this PR: the expansion runs again for every gate, and with one set per qubit there are now more of them. 8 qubits, 140 CNOTs with global noise went 0.29s -> 0.52s for me. Per-qubit Lindblad jumps are already expanded once at parse time (`parsers.cpp:601`); the per-qubit Kraus ones could be too, and the global ones cached per qubit.

fedonman (approved):
Thank you Luke. LGTM, but left some comments to consider.

src/qilisdk_cpp/backends/qilisim/noise/noise_model.cpp
Lumorti: Good idea, added an issue https://github.com/qilimanjaro-tech/qilisdk/issues/399

tests/integration/noise/test_noise_backends.py
Lumorti: Good catch, will update #398

src/qilisdk_cpp/backends/qilisim/noise/noise_model.cpp
Lumorti: It seems quite a small fix, so I include it in this PR

src/qilisdk_cpp/backends/qilisim/noise/noise_model.cpp
Lumorti: Done!

PR #398, opened 2026-08-17
CudaBackend Multi-qubit Noise Fixes

Lumorti:
The way CudaBackend internally handles digital noise has been refactored to be more consistent with how QiliSim handles it, specifically it now has a function which returns all noise types acting on a specific qubit/gate, which are then turned into Kraus operators, expanded as needed, and given to the simulation. This still keeps the efficient CUDA bitflip channels etc. and nothing about the public-facing API has changed. This now allows for things like noise acting on multi-qubit gates, which were previously just ignored.

This is the CudaBackend fix for https://github.com/qilimanjaro-tech/qilisdk/issues/389


Lumorti:
Comment from Vyron on the other PR:

> https://github.com/qilimanjaro-tech/qilisdk/pull/398 adds these same three tests right here with _cuda_ names, so the two will conflict and leave near-duplicates. Whichever lands second, could you fold them into one set parametrized over backends like the rest of the file? The issue was that the two backends disagreed with each other, so one shared test is the thing that keeps them honest.

src/qilisdk/backends/cuda_backend.py
```diff
@@ -595,155 +650,147 @@ def _handle_readout_errors(cudaq_result: dict[str, int], noise_model: NoiseModel
...
-        # Per gate per qubit noise
-        for (gate_type, qubit), noises in noise_model.per_gate_per_qubit_noise.items():
-            self._add_per_gate_per_qubit_noise(
-                gate_type, qubit, noises, cuda_noise_model, all_cuda_gate_names, noise_model.noise_config
-            )
+        for gate_type in self._basic_gate_handlers:
+            gate_name = _CUDA_GATE_NAMES.get(gate_type, gate_type.__name__.lower())
+            if gate_name is None:
+                continue
+            if relevant_noise := self._get_relevant_noise(gate_type, noise_model):
```
fedonman: Can you check this? It passes on
QiliSim and fails on Cuda in both cases:

```python
@pytest.mark.parametrize("backend_class", backends)
@pytest.mark.parametrize(("noisy_gate", "expected"), [(CNOT, {"11": 100}), (X, {"00": 100})])
def test_per_gate_noise_on_controlled_gate(noisy_gate, expected, backend_class):
    circuit = Circuit(nqubits=2)
    circuit.add(CNOT(0, 1))

    noise_model = NoiseModel()
    noise_model.add(BitFlip(probability=1.0), gate=noisy_gate)

    backend = backend_class(noise_model=noise_model, **args_per_backend[backend_class])
    result = backend.execute(DigitalPropagation(circuit), readout=Readout().with_sampling(nshots=100))

    assert result.get_samples() == expected
```

Cuda gives `{'00': 100}` for `gate=CNOT` and `{'11': 100}` for `gate=X`.

src/qilisdk/backends/cuda_backend.py
```diff
@@ -80,40 +80,94 @@ def _complex_dtype() -> np.dtype:
...
+_SWAP_MATRIX = [1, 0, 0, 0, 0, 0, 1, 0, 0, 1, 0, 0, 0, 0, 0, 1]
+
 TBasicGate = TypeVar("TBasicGate", bound=BasicGate)
 BasicGateHandlersMapping = dict[Type[TBasicGate], Callable[[cudaq.Kernel, TBasicGate, cudaq.QuakeValue], None]]
 
 TPauliOperator = TypeVar("TPauliOperator", bound=PauliOperator)
 PauliOperatorHandlersMapping: TypeAlias = dict[Type[TPauliOperator], Callable[[TPauliOperator], ElementaryOperator]]
 
+# Overrides for some gate names, None for gates that don't have a CUDA-Q equivalent
+_CUDA_GATE_NAMES: dict[Type[BasicGate] | Type[Gate], str | None] = {SWAP: _SWAP_OP_NAME, U1: None, U2: None}
```
fedonman: The comment is not quite right, U1 and U2 do have a CUDA-Q equivalent, they are both emitted as `u3`. Dropping them makes CudaBacked and Qilisim disagree. E.g.

```
noise_model = NoiseModel()
noise_model.add(BitFlip(probability=1.0), gate=U1)

circuit = Circuit(nqubits=1)
circuit.add(U1(0, phi=np.pi))

for backend_class in (QiliSim, CudaBackend):
    backend = backend_class(noise_model=noise_model)
    print(backend.execute(DigitalPropagation(circuit), readout=Readout().with_sampling(nshots=100)).get_samples())
```

prints

```
{'1': 100}
{'0': 100}
```

changes/398.bugfix.md
```diff
@@ -0,0 +1 @@
+The way CudaBackend internally handles digital noise has been refactored to be more consistent with how QiliSim handles it, specifically it now has a function which returns all noise types acting on a specific qubit/gate, which are then turned into Kraus operators, expanded as needed, and given to the simulation. This still keeps the efficient CUDA bitflip channels etc. and nothing about the public-facing API has changed. This now allows for things like noise acting on multi-qubit gates, which were previously just ignored.
```
fedonman: A more detailed changelog suggestion.

```suggestion
CudaBackend now handles noise on multi-qubit gates correctly: previously if you had noise on qubit 1 and then did a swap or a CNOT involving qubit 1, the noise was dropped entirely. It is now applied to that qubit alone, whether it is a target or a control.

Global and per-gate noise is now applied independently to each qubit a gate acts on, instead of as one correlated operator across the qubits.

A global multi-qubit Kraus channel is now applied to any gate acting on that many qubits, rather than only to a gate spanning the whole register. Channels that do not fit the gate are skipped with a warning.
```

fedonman (changes requested):
Thank you Luke! Left some comments.

src/qilisdk/backends/cuda_backend.py
Lumorti: By "not having an equivalent" I mean that CUDA-Q doesn't have a U1 object, so it doesn't distinguish between U1 and U3 at the gate level. We convert U1 and U2 to U3. So the user trying to put noise on just U1 I don't think ever worked for CudaBackend, because we would have to put on all U3

I'll investigate to see if there's some way we can add gate noise specific for U1 in CudaBackend, and if not I'll just raise a warning telling the user that their noise has been dropped

changes/398.bugfix.md
Lumorti: This is a duplicate of the QiliSim changelog, no? Maybe I combine both (i.e. "Both QiliSim and CudaBackend now handle noise...")

fedonman: But they are different PRs, no?

Lumorti: Yeah true, I just thought it might be weird if we have two identical features in the changelog just with one line changed, but maybe it's fine, I'll make the above change

src/qilisdk/backends/cuda_backend.py
Lumorti: Fixed in the rework

src/qilisdk/backends/cuda_backend.py
Lumorti: Added a warning if they try to do noise specific to u1/u2

fedonman (approved):
Thank you Luke!

PR #401, opened 2026-08-19
QUBO with negative integers

### LINKED ISSUE #385: [Bug]: to_qubo() rejects real, integer and shifted-bound variables
### Bug Description

`QUBO._check_variables` only accepts binary and 0-based positive integer variables, so a model with a real variable, an integer variable, or any variable with a non-zero lower bound is refused. The encodings already handle all three: `to_binary()` expands them correctly and spans exactly the requested bounds. The real-variable example in the variables docs cannot be converted.

```python
from qilisdk.core import Domain, Model, Variable
from qilisdk.core.variables import Bitwise

x = Variable("x", domain=Domain.REAL, bounds=(1, 2), encoding=Bitwise, precision=1e-1)
print(x.to_binary())  # 0.1 * (10.0 + x(0) + 2 * x(1) + 3.0 * x(3) + 4 * x(2))

model = Model("m")
model.set_objective(x)
model.to_qubo()
```

Same for `Domain.INTEGER` with bounds `(-2, 2)`, and for `Domain.POSITIVE_INTEGER` with bounds `(1, 5)`.

### Expected Behavior

`to_qubo()` converts the model using the binary expansion the encoding already produces.

### Actual Behavior

```shell
ValueError: QUBO models are not supported for variables that are not in the positive integers or binary domains.

# POSITIVE_INTEGER with bounds (1, 5):
ValueError: All variables must have a lower bound of 0. But variable x has a lower bound of 1
```

### System Information

```shell
qilisdk 0.2.2rc1 (dev checkout)
Python 3.13.12
Linux x86_64
```

Lumorti:
QUBO models now support the use of bounded floats and negative integers. Whilst the encoding of such variables existed before, they were over-cautiously disallowed by QUBO, however as long as the variable is bounded and has a finite binary encoding, there is no reason why it cannot be put in QUBO form.

```python
from qilisdk.core import Variable, Domain, Model, GEQ
from qilisdk.core.variables import Bitwise
from qilisdk.utils.classical_solvers import BruteForceSolver
x = Variable("x", Domain.INTEGER, bounds=(-3, 3), encoding=Bitwise)
m = Model("m")
m.set_objective(x + 1)
m.add_constraint("c", GEQ(x, 0))
q = m.to_qubo()
sol = BruteForceSolver().solve(q)
print(sol)
```

A bug was fixed in which GEQ and LEQ resulted in the same QUBO constraint.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/385


src/qilisdk/core/model.py
```diff
@@ -1545,7 +1551,7 @@ def _transform_constraint(
                     f"{label}_slack", domain=Domain.POSITIVE_INTEGER, bounds=(0, ub_slack), encoding=Bitwise
                 )
                 slack_terms = slack.to_binary()
-                out = h + slack_terms
+                out = h - slack_terms
```
Lumorti, line 1547: Note for reviewers: this is the fix for GEQ giving the same constraint as LEQ, previously both generated (h+slack)**2 but for GEQ it should be (h-slack)**2

Added a lot of tests for this

src/qilisdk/core/variables.py
```diff
@@ -383,12 +383,11 @@ def encode(var: Variable, precision: float = 1e-2) -> Term:
         if var.domain is Domain.REAL:
             bounds = (bounds[0] / precision, bounds[1] / precision)
 
-        abs_bound = np.abs(bounds[1] - bounds[0])
-        n_binary = int(np.floor(np.log2(abs_bound if abs_bound != 0 else 1)))
-        binary_vars = [BinaryVariable(var.label + f"({i})") for i in range(n_binary + 1)]
+        n_binary = Bitwise.num_binary_equivalent(var, precision)
+        binary_vars = [BinaryVariable(var.label + f"({i})") for i in range(n_binary)]
```
Lumorti, line 224: Note for reviewers: in order to make everything more consistent within the file, the encodings now all use their function to get the num binary vars, previously the code was duplicated twice per encoding type

In the case of Bitwise here, the n_binary is now the literal number of binary variables, rather than with a -1 as before, so that -1 has been added to the below equations which is the only time it's used and imo doesn't warrant a new var

src/qilisdk/core/model.py
```diff
@@ -34,6 +34,12 @@
 
 _EMPTY_GRAPH_MSG = "The graph must have at least one edge."
 
+# Everything apart from SPIN because for now they have no binary encoding
+_QUBO_SUPPORTED_DOMAINS = {Domain.BINARY, Domain.POSITIVE_INTEGER, Domain.INTEGER, Domain.REAL}
+
+# Used to prevent unbounded vars from expanding into infinite binary vars
+_MAX_ENCODING_SPAN = 2**32
```
amirazzam2000, line 41: we should probably but a much smaller bound.

amirazzam2000 (approved):
Thanks Luke! 🙌

src/qilisdk/core/model.py
Lumorti: ah oops, I merged and then I saw this, I'll add it as a bugfix issue on github

PR #404, opened 2026-08-20
Trim the CI matrices and gate on aggregate checks

### LINKED ISSUE #403: [Task]: Trim the CI matrices and fix the checks that never fail
### What needs doing

Cut the CI matrices down to the legs that give different signal, and stop the ruleset from naming every leg.

- Add an aggregate gate job to each workflow and make that the only required check on `main`.
- Split `code_quality.yml` into a lint job with no project install (ruff, format, cpp-linter) and a type job that runs `ty` once per Python version. Make the clang-format half of cpp-linter fail the job.
- Drop 3.12 from `tests.yml` and keep the floor and the ceiling.
- Cut `platform_builds.yml` to one Python version per OS plus a 3.11 leg on Linux, and give it the concurrency, draft and timeout settings the other two workflows already have.
- Set the ruff `target-version` to match `requires-python`.
- Add `CIBW_TEST_COMMAND` to `publish.yml`.

### Why

A push to a PR starts 18 jobs and about 44 minutes of runner time, most of it the same C++ build over and over. In Code Quality the checks take 2 seconds per job and the remaining 3 minutes are the install and clang-tidy. Ruff behaves the same on all three Python versions because `target-version` is pinned in config, and nothing in `src/` branches on the Python version, so the 3.12 legs are interpolation.

Two things are wrong rather than just wasteful. cpp-linter reports failing checks and the job still goes green, because the action only comments unless you check its output. Its clang-tidy findings are all missing system headers, since it runs with no compile database, so only the formatting side can gate anything today. And nothing ever imports the wheels we publish, so a broken delvewheel repair or rpath would ship. Meanwhile the `main` ruleset pins all 18 job names as required checks, so no matrix can change without every PR blocking on a check that no longer reports.

fedonman:
Closes #403.

A PR used to start 18 jobs and about 44 minutes of runner time, almost all of it the same C++ build repeated. Each workflow now ends in one aggregate gate job, so the ruleset can require `Code Quality`, `Tests` and `Platform Builds` instead of naming every matrix leg. Code Quality is split into a lint job that installs no project and a type job that runs `ty` once per supported version in a single environment, Tests and Platform Builds run the oldest and newest Python on PRs and the full set on main, and the ruff target version now comes from `requires-python` so 3.12-only syntax cannot pass lint. clang-format failures now fail the job, which they never did. On the publish side the wheels get imported before upload and a bad CodeArtifact input fails up front instead of after the builds. Nine jobs on a PR instead of eighteen.

The ruleset needs its required checks switched to the three gate names when this merges, and open PRs will need a rebase to report them.

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -15,9 +15,6 @@ concurrency:
   group: code-quality-${{ github.head_ref || github.ref }}
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
-permissions:
-  contents: read
-
 jobs:
   lint:
     name: Lint
@@ -47,7 +44,7 @@ jobs:
       # code, so linting needs neither the extension built nor the dependencies.
       - name: Install the linters
         run: |
-          uv sync --frozen --only-group dev --no-install-project
+          uv sync --frozen --only-group dev --no-install-project --no-build
           echo "VIRTUAL_ENV=$PWD/.venv" >> "$GITHUB_ENV"
           echo "$PWD/.venv/bin" >> "$GITHUB_PATH"
 
@@ -80,6 +77,9 @@ jobs:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
     timeout-minutes: 10
+
+    permissions:
+      contents: read
 
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
@@ -119,6 +119,7 @@ jobs:
     if: always()
     runs-on: ubuntu-latest
     timeout-minutes: 5
+    permissions: {}
 
     steps:
       - name: Check the results
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -11,22 +11,23 @@ on:
       - "synchronize"
       - "ready_for_review"
 
+concurrency:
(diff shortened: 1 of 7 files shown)


.github/workflows/code_quality.yml
```diff
@@ -68,3 +63,65 @@ jobs:
...
+    needs: [lint, types]
+    if: always()
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions: {}
+
+    steps:
+      - name: Check the results
+        if: contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled')
+        run: exit 1
```
Lumorti: what does this job do?

.github/workflows/tests.yml
```diff
@@ -50,7 +48,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install qilisdk with coverage
-        if: matrix.coverage
+        if: matrix.python-version == '3.13'
```
Lumorti: could we make this a variable like before? since otherwise we have to change many lines if we want to change which version we use for coverage

Lumorti (approved):
LGTM, thanks Vyron! Left a minor comment

.github/workflows/tests.yml
fedonman: Done, added 
```
env:
  # The version that carries coverage and the C++ suite.
  COVERAGE_PYTHON: "3.13"
```

PR #444, opened 2026-08-27
Replace GPUtil

### LINKED ISSUE #443: [Task]: Replace GPUtil with a different library
### What needs doing

GPUtil has some issues on Mac, and in general is depreciated. We should replace it with a different library, such as perhaps https://pypi.org/project/gpuq/.

### Why

Some people have mentioned it can cause some virtual environment issues on mac. Regardless, it is depreciated so we should find a more up-to-date library for this.

Lumorti:
Replaced the deprecated `gputil` dependency with a direct `nvidia-smi` query for the GPU reporting in `about()`.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/443


src/qilisdk/_info.py
```diff
@@ -69,33 +68,38 @@ def about() -> str:
...
-        nvidia_smi_output = subprocess.check_output(  # ruff: ignore[subprocess-popen-with-shell-equals-true]
-            ["nvidia-smi | grep 'Driver'"],  # ruff: ignore[start-process-with-partial-path]
-            shell=True,
+        gpu_query = subprocess.check_output(
+            ["nvidia-smi", "--query-gpu=name,memory.total", "--format=csv,noheader,nounits"],  # ruff: ignore[start-process-with-partial-path]
+            stderr=subprocess.STDOUT,
+        ).decode()
+        name, _, vram = gpu_query.strip().splitlines()[0].partition(",")
+        gpu_name = name.strip()
+        gpu_vram = round(int(vram) / 1024)
```
fedonman: `stderr=subprocess.STDOUT` pipes warnings into the CSV, and the parse takes line 1 whatever it looks like, so a warning line becomes the GPU name. Send stderr to `DEVNULL` for the machine-readable query and only accept a row that actually has a comma.

```suggestion
        gpu_query = subprocess.check_output(
            ["nvidia-smi", "--query-gpu=name,memory.total", "--format=csv,noheader,nounits"],  # ruff: ignore[start-process-with-partial-path]
            stderr=subprocess.DEVNULL,
        ).decode()
        name, separator, vram = gpu_query.strip().splitlines()[0].partition(",")
        if separator:
            gpu_name = name.strip()
            gpu_vram = round(int(vram) / 1024)
```

Fails on this branch (prints `GPU Info: WARNING: infoROM is corrupted`), passes with the change:

```python
def test_gpu_name_needs_a_csv_row(monkeypatch):
    _monkeypatch_all(monkeypatch, gpus=[("Test GPU", 8 * 1024)])

    def fake_check_output(args, **kwargs):
        if args[0] != "nvidia-smi":
            return b"test_output"
        if any(arg.startswith("--query-gpu") for arg in args):
            return b"WARNING: infoROM is corrupted\n"
        return b""

    monkeypatch.setattr("subprocess.check_output", fake_check_output)

    assert "GPU Info: Not Found" in about()
```

fedonman (commented):
Thanks Luke, left one minor comment.

fedonman (approved):
Thank you Luke! The fewer dependencies, the better!

PR #451, opened 2026-08-28
fixing schedule title

### LINKED ISSUE #448: [Bug]: Schedule.draw() has the title "eigenvalues" even though it isn't
### Bug Description

Schedule.draw() has the title "Schedule eigenvalues" even though it isn't

```python
nqubits = 10
T = 20.0
nsteps = 1000
dt = T / nsteps
Hz = sum(Z(i) for i in range(nqubits))
schedule = Schedule.linear(Hx, Hz, T, dt)
schedule.draw()
```

### Expected Behavior

It has a title like "Schedule coefficients"

### Actual Behavior

```shell
It has the title "Schedule eigenvalues"
```

### System Information

```shell
------------------------------------------------------
             Start of QiliSDK Debug Info
------------------------------------------------------
QiliSDK Version: 0.2.2rc1
Python Version: 3.13.11 (main, Jan 14 2026, 19:38:04) [Clang 21.1.4 ]
Numpy Version: 2.5.2
SciPy Version: 1.18.0
QuTiP Version: 5.3.1
CUDA-Q Version: CUDA-Q Version 0.15.1 (https://github.com/NVIDIA/cuda-quantum aca5853a76d499ecc3d5f97c2e06163ae99d9c75)
Platform: Linux 6.17.0-1032-oem (#32-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul 24 05:46:26 UTC 2026)
Processor: x86_64
CPU Info: AMD Ryzen 7 8845HS w/ Radeon 780M Graphics
Number of CPU Cores: 8
Number of Logical Processors: 16
Available Memory: 32 GB
GPU Info: NVIDIA GeForce RTX 4070 Laptop GPU with 7 GB VRAM
CUDA Version: 13.2
NVIDIA Driver Version: 595.84
g++ Version: g++ (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
g++ OpenMP Support: Yes
clang++ Version: Ubuntu clang version 18.1.3 (1ubuntu1)
clang++ OpenMP Support: Yes
QiliSim Import: Success
QTensor Import: Success
------------------------------------------------------
              End of QiliSDK Debug Info
------------------------------------------------------
```

amirazzam2000:
A bug was fixed in which `Schedule.draw()` labelled its plot "Schedule Eigenvalues", because it fell back to the shared renderer default title instead of setting its own. `Schedule.draw()` now titles the plot "Schedule Coefficients" (which is what it actually plots) and `Schedule.draw_eigenvalues()` titles it "Schedule Eigenvalues"; passing an explicit `ScheduleStyle(title=...)` still overrides both.

resolves https://github.com/qilimanjaro-tech/qilisdk/issues/448


src/qilisdk/utils/visualization/schedule_renderers.py
```diff
@@ -77,7 +77,7 @@ def setup_axes(self) -> None:
             for text in leg.get_texts():
                 text.set_color(title_color)
         self.ax.set_title(
-            self.style.title or "Schedule Eigenvalues",
+            self.style.title or "Schedule Coefficients",
```
Lumorti: Maybe the default should be "Schedule" since it's a shared method? That way if later add a new draw method it won't be wrong by default
```suggestion
            self.style.title or "Schedule",
```

changes/451.bugfix.md
```diff
@@ -0,0 +1 @@
+A bug was fixed in which `Schedule.draw()` labelled its plot "Schedule Eigenvalues", because it fell back to the shared renderer default title instead of setting its own. `Schedule.draw()` now titles the plot "Schedule Coefficients" (which is what it actually plots) and `Schedule.draw_eigenvalues()` titles it "Schedule Eigenvalues"; passing an explicit `ScheduleStyle(title=...)` still overrides both.
```
Lumorti:
```suggestion
A bug was fixed in which `Schedule.draw()` labelled its plot "Schedule Eigenvalues", but now titles the plot "Schedule Coefficients" (which is what it actually plots).
```

Lumorti (approved):
Thanks Ameer! 🐶️

Maybe add a simple test to make sure for both methods the title is set correctly, and I left some auto-apply suggestions for minor tweaks.

src/qilisdk/utils/visualization/schedule_renderers.py
amirazzam2000: fair point, I set it to Schedule coefficients because it was inside the `MatplotlibScheduleRenderer` and there is another renderer `MatplotlibEigenvalueRenderer` which i guess should override this, but yeah Schedule is fine either way

fedonman (approved):
Thank you Ameer!

PR #452, opened 2026-08-28
fixing readout builder

### LINKED ISSUE #435: [Bug]: The Readout builders mutate self while the docstring promises a copy
### Bug Description

`Readout.with_sampling`, `with_expectation` and `with_state_tomography` all assign the slot on `self` and `return self`, each with a `# ty:ignore[invalid-return-type]` because the annotation does not match what is returned. The class docstring says the opposite: each `with_*` method "returns a **new** `Readout`" and "The original instance is not mutated." So the documented pattern of building one base spec and branching from it cross-contaminates the branches, `a is base` is True, and a second `with_*` call on the same base raises "already set". The `Readout[S, E, T]` type parameter design only makes sense with real copies. Copying in the builders, so each one constructs a new `Readout`, carries over the slots already set and sets the new one, would match the docstring and let the three `ty:ignore` comments go; keeping the `ValueError` when the slot is already filled on the source. If the mutating style is preferred, then the docstring is what needs fixing, but the two should not keep contradicting each other.

```python

base = Readout()
a = base.with_sampling(nshots=10)
b = base.with_expectation(observables=[])

print("a is base:", a is base)
print("branch a picked up b's readout:", a.expectation is not None)
base.with_sampling(nshots=99)
```

### Expected Behavior

`a is base` is False, branch `a` never sees the readout added on branch `b`, and a second `with_sampling` on the untouched base is fine.

### Actual Behavior

```shell
a is base: True
branch a picked up b's readout: True
ValueError: Sampling readout already set in this specification.
```

### System Information

```shell
Python 3.13.12
Linux-7.0.0-30-generic-x86_64
```

amirazzam2000:
`Readout.with_*` no longer mutates the specification it is called on. The class documents that each `with_*` builder returns a new `Readout` and leaves the original untouched, but `with_*` was setting the sampling slot on `self` and returning it, so a shared base specification was silently modified:

```python

base = Readout()
sampled = base.with_sampling(nshots=1000)

print(base.sampling)  # previously the SamplingReadout, now None
```

`with_*` now populates a copy, so `base` stays reusable for building further specifications.

resolves https://github.com/qilimanjaro-tech/qilisdk/issues/435


Lumorti (approved):
Thanks Ameer, nice fix! 🥳️

Quick question, was this really only specific to sampling and not the others? Since the fix here is the same fix for all readout types. Before merging, could you check if this actually fixed a bug with all readout types rather than just sampling, and if so update the changelog accordingly

fedonman (approved):
Thank you Ameer!

PR #453, opened 2026-08-28
Unbundle CUDA-Q and QuTiP and support Python 3.14

### LINKED ISSUE #289: [Feature]: Support for Python 3.14
### Feature Summary

Add official support for Python 3.14, including CI test coverage and updated package metadata, so QiliSDK can be installed and run on the latest Python release.

Concretely, this would involve:                                                                                                            
  - Bumping the requires-python / classifier range in pyproject.toml to include 3.14.                                                        
  - Adding Python 3.14 to the CI test matrix.                                                                                                
  - Verifying/upgrading dependencies that don't yet ship 3.14-compatible wheels.

### Motivation / Use Case

Python 3.14 is long released, and users on newer environments (fresh installs, updated CI runners, distro upgrades) want to adopt it. Currently, installing QiliSDK on 3.14 either fails on version constraints or is untested, forcing users to pin to an older interpreter. Official 3.14 support lets teams stay current and benefit from the latest performance and language improvements without holding back their whole stack for QiliSDK.

### Proposed Solution

```python

```

### Alternatives Considered

_No response_

### Priority

Medium - Would improve my workflow

### Additional Context

_No response_

fedonman:
QiliSDK no longer installs CUDA-Q or QuTiP. The `cuda12` and `cuda13` extras now only carry the CUDA math libraries QiliSim needs for GPU support, and `CudaqBackend` and `QutipBackend` run against whatever version of their framework you install yourself. `CudaBackend` and `CudaSamplingMethod` are renamed to `CudaqBackend` and `CudaqSamplingMethod`, with the old names kept as deprecated aliases. Python 3.14 is supported everywhere except `CudaqBackend`, since CUDA-Q ships no 3.14 wheels. A new CI job installs the lowest versions our bounds allow, which turned up two wrong floors, `pyqir` and `pyscipopt`.

Closes #289


.github/workflows/tests.yml
```diff
@@ -53,7 +61,7 @@ jobs:
       - name: Install qilisdk with coverage
         if: matrix.python-version == env.COVERAGE_PYTHON
         run: |
-          uv sync --frozen --group dev --extra all-cu13 --reinstall-package qilisdk \
+          uv sync --frozen --group dev --extra all-cu13 --reinstall-package qilisdk --no-build \
```
Lumorti, line 64: doesn't this mean we won't get C++ coverage?

.github/workflows/tests.yml
```diff
@@ -101,9 +109,144 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
 
+  core-tests-lowest:
```
Lumorti, line 114: I don't get this section, isn't the lowest version also fully tested above, with coverage too?

.github/workflows/tests.yml
```diff
@@ -101,9 +109,144 @@ jobs:
...
+        with:
+          enable-cache: true
+          cache-dependency-glob: |
+            **/pyproject.toml
+            **/uv.lock
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install qilisdk with the external backends
+        run: |
+          uv sync --frozen --group dev --group backends --extra all-cu13 --reinstall-package qilisdk --no-build \
```
Lumorti, line 187: Won't the --no-build here mean that we don't do the QiliSim CUDA tests?

.github/workflows/tests.yml
```diff
@@ -101,9 +109,144 @@ jobs:
...
+        run: |
+          uv sync --resolution lowest-direct --group dev --extra all-cu13 --reinstall-package qilisdk \
+            -Ccmake.build-type=Release
+          echo "VIRTUAL_ENV=$PWD/.venv" >> "$GITHUB_ENV"
+          echo "$PWD/.venv/bin" >> "$GITHUB_PATH"
+
+      - name: Run Python tests
+        run: pytest -n auto tests/unit_python/ $(printf -- '--ignore=%s ' $EXTERNAL_BACKEND_TESTS)
+
+  external-backends:
```
Lumorti, line 156: I feel like there's a lot of duplication here, we just need one test suite that:
 - if it's 3.11 it enables coverage
 - if it's 3.11 - 3.13 it enables tests for the external deps

Right now I think we're running the whole test suite multiple times per python version, when we only need to run it once per version (enabled coverage/externals depending on the version)

changes/453.misc.md
Lumorti, line 1: setuptools was required by something in the about() method without actually being imported, so please double check that. it might be that you already have the thing it needs installed elsewhere on your system already, I think I added it as a dep since someone had issues running about() and the fix was to make sure setuptools was installed so then they had some certain CPU/GPU checking command

Lumorti (changes requested):
Thanks Vyron 🚀️

The changes in the library seem fine, but I left some comments about the testing workflow. Regardless I think we should discuss this a bit more in a daily at some point just to confirm this is the best solution before merging.

.github/workflows/tests.yml
fedonman: No, `--no-build` only blocks building dependencies from sdists, uv still builds the project.

.github/workflows/tests.yml
fedonman: Different "lowest". That job is `--resolution lowest-direct` (oldest dependency versions) which is what caught the wrong `pyqir` and `pyscipopt` floors. I added comments in the workflow to clarify.

.github/workflows/tests.yml
fedonman: No, this job only runs the four `EXTERNAL_BACKEND_TESTS` files. `test_gpu.py` runs in the core job.

changes/453.misc.md
fedonman: It was needed by `GPUtil`, no longer needed.

.github/workflows/tests.yml
fedonman: Two things.

There is no duplication. The core job ignores the four files in EXTERNAL_BACKEND_TESTS and the external job runs only those four, so every test runs exactly once per Python version.

And the jobs are parallel, so the extra ones cost us nothing in waiting. In the last run all 8 started within 2 seconds of each other, no queueing, and the whole workflow took 4m21s, set by the 3.11 job that also builds with coverage and runs the C++ suite.

Lumorti (approved):
Thanks for addressing the comments! In that case seems good to me. 💯️

PR #465, opened 2026-09-08
List the alternative requirements when no optional group is satisfied

fedonman:
When an optional feature needs any one of several dependency groups and none of them is installed, the warning only said "no dependency group satisfied", so you had to go read the feature definition to find out what to install. It now lists the requirements each alternative wanted, the same way the all-of case already did: `[Optionals] Optional feature cudaq unavailable, unsatisfied requirements ['cuda-quantum-cu12>=0.14.0 (not installed)'] or ['cuda-quantum-cu13>=0.14.0 (not installed)']`.


NataliaSirera (approved):
Thanks!!

PR #467, opened 2026-09-09
[Bug-466] fixing the plotting in the experiments

### LINKED ISSUE #466: [Bug]: Wrong plotting in the experiments class
### Bug Description

`ExperimentResult._plot_2d` in `src/qilisdk/experiments/experiment_result.py` mirrors the
data whenever a sweep axis descends, and gets the cell geometry wrong on every 2-D plot.

```python
import matplotlib; matplotlib.use("Agg")
import numpy as np
from qilisdk.experiments.experiment_result import Dimension, ExperimentResult

class R(ExperimentResult):
    plot_title = "mirror"

x = np.arange(0.515, 0.48, -0.0035)          # descending, 11 points
y = np.arange(-0.2e9, -0.05e9, 1e6)
data = np.zeros((len(x), len(y), 2))
data[0, :, 0] = 50.0                          # one bright column at x[0] == 0.515

R(qubit=0, averages=1, data=data,
  dims=[Dimension(["flux"], [x]), Dimension(["freq"], [y])]).plot(save_to="mirror.png")
```

### Expected Behavior

it should render the bright colum at 0.515.

### Actual Behavior

```shell
The bright column was measured at 0.515 and renders at 0.480, the far left.
```

### System Information

```shell
------------------------------------------------------
             Start of QiliSDK Debug Info
------------------------------------------------------
QiliSDK Version: 0.2.2rc1
Python Version: 3.13.9 (main, Oct 31 2025, 23:02:44) [Clang 21.1.4 ]
Numpy Version: 2.5.2
SciPy Version: 1.18.0
QuTiP Version: 5.3.1
CUDA-Q Version: CUDA-Q Version 0.15.1 (https://github.com/NVIDIA/cuda-quantum aca5853a76d499ecc3d5f97c2e06163ae99d9c75)
Platform: Linux 7.0.0-30-generic (#30~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Fri Aug  7 13:27:52 UTC 2)
Processor: x86_64
CPU Info: 13th Gen Intel(R) Core(TM) i9-13900H
Number of CPU Cores: 14
Number of Logical Processors: 20
Available Memory: 32 GB
GPU Info: NVIDIA GeForce RTX 4070 Laptop GPU with 8 GB VRAM
CUDA Version: 13.0
NVIDIA Driver Version: 580.178.04
g++ Version: g++ (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
g++ OpenMP Support: Yes
clang++ Version: Not Found
QiliSim Import: Success
QTensor Import: Success
------------------------------------------------------
              End of QiliSDK Debug Info
------------------------------------------------------
```

amirazzam2000:
2D `ExperimentResult.plot()` maps were mirrored whenever a sweep ran downwards, and every 2D map had slightly wrong cell geometry. The plot built its mesh edges with `np.linspace(values.min(), values.max(), len(values) + 1)`, which always ascends, while the data was painted in the order it was measured.

Any experiment sweeping a parameter from high to low (for example a flux bias ramped down) therefore produced a map flipped about the midpoint of that axis, with correct-looking tick labels and no warning. **Anyone who has published or fitted a figure from a descending sweep should re-plot it.** Only the rendering was affected: the stored data is intact, so re-plotting an existing `ExperimentResult` is enough to recover the correct figure.

The same edge calculation also treated the first and last swept points as the outer edges of the mesh rather than as cell centres, so `n` points were drawn as `n` cells of width `(n - 1) * step / n`: the axis was compressed and shifted by half a cell, and non-uniformly spaced sweeps (a log sweep, say) were redrawn as an even grid. The error is negligible for a few thousand points but reaches 9% at 11 points.

Each cell is now centred on the point it was measured at, so descending and non-uniformly spaced sweeps are drawn correctly:

```python
import numpy as np

from qilisdk.experiments import Dimension, ExperimentResult


class TwoToneVsFluxBias(ExperimentResult):
    plot_title = "two_tone_vs_flux_bias"


flux = np.arange(0.515, 0.48, -0.0035)  # a ramp played downwards
frequency = np.arange(-0.2e9, -0.05e9, 1e6)
data = np.zeros((len(flux), len(frequency), 2))
data[0, :, 0] = 50.0  # a feature measured at flux[0] == 0.515

result = TwoToneVsFluxBias(
    qubit=0,
    averages=1,
    data=data,
    dims=[Dimension(["Flux bias (V)"], [flux]), Dimension(["Frequency (Hz)"], [frequency])],
)
result.plot()  # the feature now renders at 0.515, not at 0.480
```

Secondary (twin) axes were fixed alongside: they took their limits from the smallest and largest secondary values, which flipped a secondary axis running opposite to its primary one, and left it misaligned with the primary axis. They are now placed by mapping the primary axis limits onto the secondary values, so both axes label the same positions.

fixes https://github.com/qilimanjaro-tech/qilisdk/issues/466

### DIFF
--- a/changes/467.bugfix.1.md
+++ b/changes/467.bugfix.1.md
@@ -1,0 +1,3 @@
+`ExperimentResult.plot()` now lays out the figure before writing it to disk, so a saved plot looks like the one on screen.
+
+Two separate problems: `_plot_2d` called `tight_layout()` *after* saving, so the file kept the untidied layout while the displayed figure got the tidy one, and `_plot_1d` never called it at all. In both cases a secondary axis label pushed the title off the top of the canvas, so every saved figure with a `twiny`/`twinx` axis came out with its title clipped. Re-saving an affected figure is enough to fix it.
--- a/changes/467.bugfix.2.md
+++ b/changes/467.bugfix.2.md
@@ -1,0 +1,32 @@
+On a 1D `ExperimentResult.plot()` with a secondary x axis, the fit curve was drawn on the secondary axis instead of the primary one, in the secondary axis' coordinates, which placed it far outside the visible range: the fit did not appear at all.
+
+`add_fit` is handed no axes, so an implementation can only draw on pyplot's current axes, and `Axes.twiny()` had already made the secondary axis current by the time it was called. The fit is now performed before the secondary axis is created, so the current axes is the primary one:
+
+```python
+import numpy as np
+
+from qilisdk.experiments import Dimension, ExperimentResult
+
+
+class ResonatorSpectroscopy(ExperimentResult):
+    plot_title = "resonator_spectroscopy"
+
+    @staticmethod
+    def add_fit(x_values, y_values, initial_guess=None):
+        import matplotlib.pyplot as plt
+
+        plt.plot(x_values, y_values, "-", color="red", label="fit")
+
+
+frequency = np.linspace(4.0e9, 5.0e9, 8)
+data = np.stack([np.linspace(0.1, 1.0, 8), np.linspace(-0.5, 0.5, 8)], axis=-1)
+result = ResonatorSpectroscopy(
+    qubit=0,
+    averages=1,
+    data=data,
+    dims=[Dimension(["Frequency (Hz)", "Flux bias (V)"], [frequency, np.linspace(-1.0, 1.0, 8)])],
+)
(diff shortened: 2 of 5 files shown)


Lumorti (approved):
Nice fix, thanks! 🐦‍⬛️

PR #492, opened 2026-09-14
Fix YAML Typo

### LINKED ISSUE #433: [Bug]: A typo disables the autoapi skip hook for to_yaml and from_yaml
### Bug Description

`skip_yaml_class_methods` in `docs/conf.py` tests `what == "methpod"`, but autoapi only ever passes `"method"`, so the branch is dead and the hook always returns `skip` unchanged. It is connected to `autoapi-skip-member`, so someone wrote it and wired it up and it has never once fired. The impact is one page rather than the whole reference: `to_yaml` and `from_yaml` are only defined on `QiliEnum`, and `autoapi_options` does not include `inherited-members`, so the two methods leak onto the `QiliEnum` page and nowhere else. One word to fix. Separately `source_suffix` lists `.pynb`, a typo for `.ipynb`, but that one is harmless dead text and not a second bug: nbsphinx calls `add_source_suffix('.ipynb', ...)` itself when the suffix is missing, so notebooks are picked up either way. Nothing in `tests/` touches `docs/conf.py`, which is why neither was caught.

```python
import ast, pathlib
import autoapi._objects as ao

src = pathlib.Path("docs/conf.py").read_text()
fn = next(n for n in ast.parse(src).body
          if isinstance(n, ast.FunctionDef) and n.name == "skip_yaml_class_methods")
ns = {}
exec(ast.unparse(fn), ns)
hook = ns["skip_yaml_class_methods"]

print("autoapi passes what =", repr(ao.PythonMethod.type))
name = "qilisdk.core.types.QiliEnum.to_yaml"
print("hook(what='method')  ->", hook(None, "method", name, None, False, None))
print("hook(what='methpod') ->", hook(None, "methpod", name, None, False, None))
```

### Expected Behavior

The hook returns `True` for a method named `to_yaml`, so it is skipped in the API reference.

### Actual Behavior

```shell
autoapi passes what = 'method'
hook(what='method')  -> False
hook(what='methpod') -> True
```

### System Information

```shell
Python 3.13.12, sphinx-autoapi 3.8.0
Linux-7.0.0-30-generic-x86_64
```

Lumorti:
Fixed a typo in `conf.py` which would cause the `to_yaml` and `from_yaml` methods to not be skipped in the API docs are they were supposed to be. Added tests for the `conf.py` since previously there weren't any.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/433


fedonman (approved):
Thank you Luke!

PR #495, opened 2026-09-15
Allow Multiple qreg ins OpenQASM2 Parser

### LINKED ISSUE #409: [Bug]: from_qasm2 only works when the quantum register is named "q"
### Bug Description

`from_qasm2` hardcodes the register names `q` and `c` in its operand and measure regexes. A file that names its register anything else fails with `UnsupportedGateError("Only one- and two-qubit gates are supported.")`, because the operand regex matches nothing and the qubit list comes out empty, so the error points at gate arity instead of the register name. A second `qreg` replaces the circuit instead of extending it, and a measurement into a creg not named `c` is dropped with no error at all. Qiskit, Cirq and hand written QASM2 all use names like `qr` or `q0` and often split a circuit over several registers, so the importer really only reads files QiliSDK wrote itself. QASM3 already builds a register table in `OpenQasmParser._evaluate_register` and QASM2 should do the same.

```python
from qilisdk.utils.openqasm import from_qasm2

# register not named "q": raises
from_qasm2("OPENQASM 2.0;\nqreg qr[2];\nh qr[0];\ncx qr[0], qr[1];\n")

# creg not named "c": the measurement is silently dropped
c = from_qasm2("OPENQASM 2.0;\nqreg q[1];\ncreg cr[1];\nh q[0];\nmeasure q[0] -> cr[0];\n")
print([str(g) for g in c.gates])
```

### Expected Behavior

The first import gives a 2 qubit circuit with `H(0)` and `CNOT(0, 1)`. The second prints `['H(0)', 'M(0)']`.

### Actual Behavior

```shell
qilisdk.digital.exceptions.UnsupportedGateError: Only one- and two-qubit gates are supported.

# with the first import removed, the measure case prints:
['H(0)']
```

### System Information

```shell
Python 3.13.12
Linux-7.0.0-30-generic-x86_64
```

Lumorti:
Multiple quantum registers are now supported in the OpenQASM2 parser. 

This fixes the latter half of https://github.com/qilimanjaro-tech/qilisdk/issues/409, the former half was already fixed in main in an earlier PR.


fedonman (approved):
Thank you Luke!

PR #499, opened 2026-09-15
Circuit addition no longer mutates the original

### LINKED ISSUE #337: [Bug]: Circuit composition mutates its operands
### Bug Description

`Circuit.__add__` calls `self.add` and returns `self`, so `c1 + c2` mutates `c1`. `__radd__` mutates too and `__iadd__ = __add__`. Unsupported operands return a `NotImplementedError` instance instead of `NotImplemented`, so `circuit + 5` gives back an exception object instead of raising `TypeError`. `insert()` with the default `index=-1` does `self._insert(gate, i + index)` for iterables, which scrambles the order, and there is no `Circuit.copy()` while `append` / `prepend` share gate objects, so setting parameters on one circuit changes the other. All in `src/qilisdk/digital/circuit.py`.

```python
from qilisdk.digital import Circuit, H, X, RX

c1 = Circuit(2); c1.add(X(0))
c2 = Circuit(2); c2.add(H(1))
print("c3 is c1:", (c1 + c2) is c1, "| c1.gates:", c1.gates)
print("c + 5 ->", repr(Circuit(2) + 5))

c4 = Circuit(2); c4.add([X(0), X(1)]); c4.insert([H(0), H(1)])
print("insert default index:", c4.gates)

a = Circuit(1); a.add(RX(0, theta=0.5))
b = Circuit(1); b.append(a); b.set_parameter_values([1.234])
print("shared:", b.gates[0] is a.gates[0], "| a:", a.get_parameter_values())
```

### Expected Behavior

`c1 + c2` returns a new circuit and leaves `c1` alone, `circuit + 5` raises `TypeError`, `insert` with the default index appends in order, and there is a `Circuit.copy()` giving independent parameters.

### Actual Behavior

```shell
c3 is c1: True | c1.gates: [X(0), H(1)]
c + 5 -> NotImplementedError('Addition is only supported between Circuit objects or a Circuit and a Gate objects')
insert default index: [H(1), X(0), H(0), X(1)]
shared: True | a: [1.234]
```

### System Information

```shell
QiliSDK Version: 0.2.2
Python Version: 3.13.11
Numpy Version: 2.4.4
```

Lumorti:
Doing `new_circuit = circuit_1 + circuit_2` no longer mutates the `circuit_1`, now `new_circuit` is an entirely new object which has the gates of `circuit_1` and `circuit_2` appended to it.

Fixed a bug in which doing `Circuit.insert(gate_list, index=-1)` would change the order of the gate list appended.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/337


src/qilisdk/digital/circuit.py
```diff
@@ -229,19 +230,32 @@ def _insert(self, gate: Gate, index: int = -1) -> None:
...
+            index (int | None, optional): The index at which the first gate is inserted, following the
+                usual list indexing rules, so a negative value counts from the end of the circuit.
+                Defaults to None, which appends the gates at the end of the circuit.
+
+        Raises:
+            QubitOutOfRangeError: If any qubit index used by a gate is not within the circuit's qubit range.
         """
         logger.trace("[Circuit] Inserting gates: {} at index: {}", gates, index)
+        if index is None:
+            start = len(self._gates)
```
amirazzam2000: should this be len(self._gates) -1? otherwise if the user sets the index to -1 which in theory would mean placing the item at the end of the list (aka. append) would not be equivalent to setting the index to None

src/qilisdk/digital/circuit.py
```diff
@@ -229,19 +230,32 @@ def _insert(self, gate: Gate, index: int = -1) -> None:
...
+        elif index < 0:
+            start = max(len(self._gates) + index, 0)
```
amirazzam2000, line 251: to me something is weird here, because in python when you index the list with -1 it means selecting the last item, which i guess is what this would do, but then it's not so intuitive that if you insert in -1 then you don't actually insert in the last position but rather you insert in the position before that. I would assume that inserting at -1 would put it at the end of the circuit. 

What do you think?

amirazzam2000 (approved):
Thanks Luke!! 💯 

I left one comment for discussion but I will approve it so that you can merge once the discussion is resolved.

src/qilisdk/digital/circuit.py
Lumorti: That was my original idea and I think part of the reason this second bug existed haha. I think often insert is "before the element the user selects", like insert at 0 means insert **before** element 0 so by extension insert at -1 would actually be insert one before the end.

The default Python list insert behavior:

```
list.insert(0, "to the start");
list.insert(len(list), "to the end");
```

so to insert at the end of the list you actually have to give an index outside of the bounds of the list (`len(list)`), if you have `len(list)-1` (which is what passing -1 would represent) then it would put it one before

I also agree it feels wrong, but I believe it is consistent with Python behavior. If you want we could also disable support for the negatives.

amirazzam2000: alright yes I checked again and you are right, then we can keep it as is.

PR #504, opened 2026-09-17
Added Logging to Docs

### LINKED ISSUE #502: [Task]: Add a docs section for logging
### What needs doing

Right now there's no documentation about the QiliSDK logging system, we should add something about it to the docs

### Why

We agreed it would be useful for the users to know how to configure the logging

Lumorti:
A section about logging in QiliSDK has been added to the docs.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/502


docs/modules/logging/logging.rst
```diff
@@ -0,0 +1,93 @@
...
+     - Meaning
+   * - ``TRACE``
+     - 🔬
+     - Very detailed information, typically only useful when written to a file.
+   * - ``DEBUG``
+     - 🐞
+     - Information meant for developers, emitted at most once per step of a normal workflow.
+   * - ``INFO``
+     - 💡
+     - Information meant for users, emitted at most once per step of a normal workflow.
```
amirazzam2000: are we sure they are all emitted at most once per step?

amirazzam2000 (approved):
Thanks Luke!! ⛰️ 
I left a small comment just to check before merging 🙌

docs/modules/logging/logging.rst
Lumorti: So this was what I tried to ensure, but "step" is sort of debatable, like for instance is the user constructing a Circuit something that is a "step"? What about creating a backend? In a simple workflow yes, but maybe a more advanced workflow has a loop running different circuits with different seeded backends etc., so then INFO would fill their terminal.

I'll tweak this wording, it's more like "should not flood the terminal during a simple workflow"

PR #505, opened 2026-09-18
Docs Utils Section

### LINKED ISSUE #503: [Task]: Add utils module to docs
### What needs doing

Add the utils module to the documentation.

### Why

Right now the utils module is still undocumented, so stuff like the interop and classical_solvers has no documentation.

Lumorti:
Added a utils section to the docs.

Resolves https://github.com/qilimanjaro-tech/qilisdk/issues/503


docs/modules/utils/utils_classical_solvers.rst
```diff
@@ -0,0 +1,197 @@
...
+   :class:`~qilisdk.utils.classical_solvers.brute_force_solver.BruteForceSolver`                   any ``Model``        ✔ (exponential cost)
+   ---------------------------------------------------------------------------------------------- -------------------- ------------------------------------------------
+   :class:`~qilisdk.utils.classical_solvers.scipy_solver.ScipySolver`                              any ``Model``        ✕ (local minimizer by default)
+   ---------------------------------------------------------------------------------------------- -------------------- ------------------------------------------------
+   :class:`~qilisdk.utils.classical_solvers.simulated_annealing_solver.SimulatedAnnealingSolver`   ``QUBO`` only        ✕ (heuristic)
+   ---------------------------------------------------------------------------------------------- -------------------- ------------------------------------------------
+   :class:`~qilisdk.utils.classical_solvers.scip_solver.ScipSolver`                                any ``Model``        ✔ (requires the ``scip`` extra)
```
amirazzam2000: doesn't this also have exponential cost? we should probably say it. because right now this looks like that the method solves any model (and finds the absolute ground truth) without exponential cost. Which I don't believe is true.

docs/modules/utils/utils_overview.rst
```diff
@@ -0,0 +1,18 @@
+Overview
```
amirazzam2000, line 1: I am missing the trotterization from this section.

amirazzam2000 (changes requested):
Thanks Luke for adding this! I left some comments about some of the wordings and maybe missing sections. But overall very well documented!! 📝

docs/modules/utils/utils_overview.rst
Lumorti: I originally planned to add it, but then I figured that the Trotterization is covered in other places, right? Like TrotterizedSchedule is covered in the digital module already, should the user ever import anything from utils.trotterization? I think there it's just helper functions

docs/modules/utils/utils_classical_solvers.rst
Lumorti: It's doing branch and cut, so it's problem dependent, like if it's an easy problem it will solve it easily but if it's hard then yes it could be worst case exponential.

Lumorti: but yes, added an extra column to the table

